PR #5, opened 2022-03-24
Add babelfish_extensions as submodule for integrated build

wieck:
### Description

Adding the babelfish_extensions repository as a submodule to postgresql_modified_for_babelfish allows building the Babelfish Extension modules within $(top_builddir) and without the need for PG_SRC and PG_CONFIG environment variables to be set.

**In this commit the submodule's URL points to a clone of the babelfish_extensions repository, which has the necessary Makefile fixes for all this to actually work. Once that pull request is merged we need to adjust that to point to the official source repo.**


nasbyj:
@3manuek you've looked at this stuff a bunch; do you think setting up a submodule will help us?

@wieck, is there a way to have the submodule default to the correct extensions branch? IE, in BABEL_1_X_DEV__PG_* it should pull BABEL_1_X_DEV, but in BABEL_2_X_DEV__PG_* it should pull BABEL_2_X_DEV

wieck:
> @wieck, is there a way to have the submodule default to the correct extensions branch? IE, in BABEL_1_X_DEV__PG_* it should pull BABEL_1_X_DEV, but in BABEL_2_X_DEV__PG_* it should pull BABEL_2_X_DEV

@nasbyj The exact behavior of a submodule is that when you check it out, it shows in the parent module as "modified". You now commit in the parent at wherever you are and that parent-commit records the exact commit-hash of the submodule(s). When you later checkout the parent commit/tag again you perform `git submodule update` and it adjust the submodule to that hash.

So it is not just that BABEL_1_X_DEV__PG_13_* will pull in BABEL_1_X_DEV, but the exact submodule hash that was checked out (and git-add'ed) on each and every commit/tag in that parent branch.

nasbyj:
How would that work for releases then? I'm guessing we'd need to stamp the extensions repo, set that to be current in the engine repo, commit that, then tag the engine repo?

Do you think we should utilize any of this for standard development (one obvious challenge there is we commit to the extensions repo far more often, so this submodule setup is "backwards" from what we'd ideally want in that regard).

wieck:
On 3/24/22 18:32, nasbyj wrote:
> How would that work for releases then? I'm guessing we'd need to stamp 
> the extensions repo, set that to be current in the engine repo, commit 
> that, then tag the engine repo?

Correct. Stamp in the extensions repo, git-add + git-commit in the 
engine (parent) repo and then tag the engine.

This will also work across parent repos because you'd be able to fetch 
and pull the submodule separately in another parent (Aurora PG) and then 
reproduce the other steps there.

> Do you think we should utilize any of this for standard development (one 
> obvious challenge there is we commit to the extensions repo far more 
> often, so this submodule setup is "backwards" from what we'd ideally 
> want in that regard).

I don't think this is much of a problem. Sure, everyone who does a fresh 
clone of the engine needs to do a BABEL_X_Y_DEV checkout in the 
extensions submodule, but on an everyday base I don't think people 
constantly pull the engine. They'd only do that if there are changes in 
the engine branch they are working on. And if there are changes, then 
the latest corresponding extensions hash should be committed with them.


Regards, Jan

PR #10, opened 2022-04-21
Add logic to SET enable_domain_typmod guc for domains having typmod enabled

rishabhtanwar29:
During pg_dump we loose the setting of enable_domain_typmod
GUC and hence restored domains show incorrect behaviour.

1. Fix this by dumping enable_domain_typmod guc setting
   for the domains which supports domain typmod.
2. Also, replace unnecessary sql_dialect check in function 
   `coerce_to_domain` with typname check for some predefined
    domains. T-SQL allows typmod for the following domains:
    sys.nvarchar, sys.nchar, sys.varbinary, sys.binary, sys.decimal

Task: BABEL-3107

### Check List

### DIFF
--- a/src/backend/commands/typecmds.c
+++ b/src/backend/commands/typecmds.c
@@ -868,13 +868,7 @@ DefineDomain(CreateDomainStmt *stmt)
 	/* Domains never accept typmods, so no typmodin/typmodout needed */
 
 	/* Only allow typmods if GUC is set, used by babelfishpg_tsql extension */
-	if (enable_domain_typmod ||
-		(strcmp(get_namespace_name(domainNamespace), "sys") == 0 &&
-		 (strcmp(domainName, "nvarchar") == 0 ||
-		  strcmp(domainName, "nchar") == 0 ||
-		  strcmp(domainName, "varbinary") == 0 ||
-		  strcmp(domainName, "binary") == 0 ||
-		  strcmp(domainName, "decimal") == 0)))
+	if (enable_domain_typmod)
 	{
 		typmodinOid = baseType->typmodin;
 		typmodoutOid = baseType->typmodout;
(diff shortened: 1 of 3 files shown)


src/backend/commands/typecmds.c
```diff
@@ -868,7 +868,13 @@ DefineDomain(CreateDomainStmt *stmt)
 	/* Domains never accept typmods, so no typmodin/typmodout needed */
 
 	/* Only allow typmods if GUC is set, used by babelfishpg_tsql extension */
-	if (enable_domain_typmod)
+	if (enable_domain_typmod ||
```
2jungkook: It will work in a short-term, i.e., MVU for sys catalog, however a more desirable solution would be to modify pg_dump when considering user-defined domains, e.g.,
```
SET enable_domain_typmod = TRUE;
CREATE DOMAIN user_defined_domain_with_typmod AS user_defined_type;
RESET enable_domain_typmod;

SET enable_domain_typmod = FALSE;
CREATE DOMAIN user_defined_domain_without_typmod AS user_defined_type;
RESET enable_domain_typmod;
```

If we are able to detect whether a domain has typmod enabled during pg_dump, then we can generate the same SQL as above.

2jungkook (commented):
Not specific to this PR, but I think we need a way to trigger Github actions for the extension package even the change is only for the engine.

src/backend/commands/typecmds.c
rishabhtanwar29: Added above logic in pg_dump.

src/bin/pg_dump/pg_dump.c
```diff
@@ -11371,6 +11382,10 @@ dumpDomain(Archive *fout, const TypeInfo *tyinfo)
 
 	appendPQExpBufferStr(q, ";\n");
 
+	if (typsupporttypmod)
+		appendPQExpBuffer(q,
+						"SET enable_domain_typmod = FALSE;\n");
```
2jungkook: Let's restore the previous value instead of putting a hard-coded value.

rishabhtanwar29: Fixed!

2jungkook (approved):
Let's use `mvu-dev` branch.

PR #13, opened 2022-05-11
BCP Related Bug-fixes

KushaalShroff:
Description
In this commit we fix 3 issues:

1. Bcp does not handle ‘db..table’ syntax. The issue was pertaining to
sp_describe_first_resultset where SPI API was not able to resolve the
DB..TABLE name. To fix this we are making use of ANTLR Parser to resolve
the full name, in addition it also provides syntax checks.

2. Bcp-in does not handle identity column. The issue was with views created
from select statement does not inherit the constraints from the base columns
which was the reason why sp_describe_first_resultset sent wrong values for
identity columns. To Fix this we lookup the catalog at the time of view
creation in TSQL Dialect to initialise the constraints for it.

3. We had restricted sp_describe_first_result_set to only work with SELECT
statements and for non-select statements we threw an error. With this commit
we return 0 rows for any valid non-select query.

Task: BABEL-3193, BABEL-3194, BABEL-3208
Signed-off-by: Kushaal Shroff ([kushaal@amazon.com](mailto:kushaal@amazon.com))
 
### Check List


src/backend/commands/view.c
```diff
@@ -71,6 +72,27 @@ DefineVirtualRelation(RangeVar *relation, List *tlist, bool replace,
 											exprType((Node *) tle->expr),
 											exprTypmod((Node *) tle->expr),
 											exprCollation((Node *) tle->expr));
+			/*
+			 * If the dialect is TSQL then preserve the constraints.
+			 */
+			if (sql_dialect == SQL_DIALECT_TSQL)
```
surendravishnoi: It should be a hook. Second, why should a view needs to inherit properties of columns which are relevant from insert perspective?

KushaalShroff: Created a hook. In SQL Server, we observe from system catalog information that the Views' inherit such constraints from base columns.

kuntalghosh (dismissed):

thephantomthief (dismissed):

hong-yuan (approved):

kuntalghosh (approved):

PR #19, opened 2022-06-09
Fix mismatch of target columns and expressions when table has default NULL column in OUTPUT clause

AvantikaDG:
When target columns are not specified in the OUTPUT INTO target table,
add the column names as target columns to the parse tree during rewrite
to CTE.

This commit also reverts the catalog lookup introduced as a part of the
BABEL-2811 fix to check for default valued columns.

Task: BABEL-2928
Signed-off-by: Avantika Dasgupta <davantik@amazon.com>

### DIFF
--- a/src/backend/parser/analyze.c
+++ b/src/backend/parser/analyze.c
@@ -56,6 +56,9 @@ pre_parse_analyze_hook_type pre_parse_analyze_hook = NULL;
 /* Hook to handle qualifiers in returning list for output clause */
 pre_transform_returning_hook_type pre_transform_returning_hook = NULL;
 
+/* Hook to modify insert statement in output clause */
+pre_transform_insert_hook_type pre_transform_insert_hook = NULL;
+
 /* Hook to read a global variable with info on output clause */
 get_output_clause_status_hook_type get_output_clause_status_hook = NULL;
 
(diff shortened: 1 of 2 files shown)


Sairakan (approved):

sangilsong (approved):

macwoody:
the code change looks good to me, I would like to double check one more fact 

My understanding is that OUPUT is implemented with "WITH clause", that's why we would like to check stmt->with ..
And the reason is that normally, we would like to throw an error for column mismatch. 
Then how do we distinguish user specified insert statement with CTE.  
Why not add NULL when we do the rewriting?

AvantikaDG:
> the code change looks good to me, I would like to double check one more fact
> 
> My understanding is that OUPUT is implemented with "WITH clause", that's why we would like to check stmt->with .. And the reason is that normally, we would like to throw an error for column mismatch. Then how do we distinguish user specified insert statement with CTE. Why not add NULL when we do the rewriting?

There is a check in the hook function of a global variable that is set at the time of parsing. If `output_into_insert_transformation` is false, we distinguish the with statement as user specified insert statement with CTE. The initial idea was to find the NULL-default columns and add NULL when we do rewriting, but it turns out that there is no good way to exhaustively find NULL-default columns. I had added a catalog lookup as a part of BABEL-2811 that would do that, but it was not enough and had to redo the whole logic.

macwoody (approved):
discussed offline

PR #21, opened 2022-06-14
Computed column ISNULL(CONVERT()) crashes PG server with segfault

thephantomthief:
### Description

In PG, the typmod of a CoalesceExpr is -1 if all the arguments of the
expression are unable to agree on type and typmod. This is not the case
for T-SQL ISNULL() (which is implemented using PG COALESCE() in
Babelfish and so uses CoalesceExpr internally). In case of T-SQL
ISNULL(), the type/typmod is determined by the first not null argument.

This commit introduces a hook namely "coalesce_typmod_hook" which is
called if we encounter a T-SQL ISNULL() function. This hook returns the
typmod of the first argument in a CoalesceExpr. If the coalesce type is
numeric, then instead of sending typmod of the first argument (which is
not entirely correct since we may not be able to fit other arguments in
that scale/precision), we use the "resolve_numeric_typmod_from_exp"
method to figure out the scale/precision for us.

Also added a boolean member variable "tsql_is_null" in the
CoalesceExpr struct which is true if we encounter a T-SQL
ISNULL(); false otherwise.

Signed-off-by: Sharu Goel <goelshar@amazon.com>
 
### Issues Resolved

https://github.com/babelfish-for-postgresql/babelfish_extensions/issues/117
 
### Check List

### DIFF
--- a/src/include/nodes/primnodes.h
+++ b/src/include/nodes/primnodes.h
@@ -1121,6 +1121,7 @@ typedef struct CoalesceExpr
 	Oid			coalescecollid; /* OID of collation, or InvalidOid if none */
 	List	   *args;			/* the arguments */
 	int			location;		/* token location, or -1 if unknown */
+	bool			tsql_is_null;	/* whether the calling function is T-SQL ISNULL() */
 } CoalesceExpr;
 
 /*
--- a/src/backend/nodes/copyfuncs.c
+++ b/src/backend/nodes/copyfuncs.c
@@ -2023,6 +2023,7 @@ _copyCoalesceExpr(const CoalesceExpr *from)
 	COPY_SCALAR_FIELD(coalescecollid);
 	COPY_NODE_FIELD(args);
 	COPY_LOCATION_FIELD(location);
+	COPY_SCALAR_FIELD(tsql_is_null);
 
 	return newnode;
 }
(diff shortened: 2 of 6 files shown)


src/include/nodes/primnodes.h
```diff
@@ -1121,6 +1121,7 @@ typedef struct CoalesceExpr
 	Oid			coalescecollid; /* OID of collation, or InvalidOid if none */
 	List	   *args;			/* the arguments */
 	int			location;		/* token location, or -1 if unknown */
+	bool			tsql_is_null;	/* whether the calling function is T-SQL ISNULL() */
```
2jungkook, line 1124: This might cause a problem in MVU because both dump and restore are done in postgres dialect, i.e., `tsql_is_null` cannot be set during MVU.

To clarify this, would you check whether a dumped SQL contains correct information? e.g., `bin/pg_dump -d jdbc_testdb -t master_dbo.t_3147_1`

2jungkook, line 1124: This might cause a problem in MVU because both dump and restore are done in postgres dialect, i.e., `tsql_is_null` cannot be set during MVU.

To clarify this, would you check whether a dumped SQL contains correct information? e.g., `bin/pg_dump -d jdbc_testdb -t master_dbo.t_3147_1`

2jungkook (approved):

surendravishnoi (approved):

PR #22, opened 2022-06-15
Add reverse dependency between TSQL trigger and trigger function during restore

surendravishnoi:
TSQL triggers needs reverse dependency from trigger function to
trigger. It helps in dropping function when trigger itself is dropped
as trigger functions are internal to TSQL triggers. This dependency
is added specifically for TSQL triggers only.

During restore, TSQL triggers are restored using two parts, first function
restore followed by trigger restore using PG semantics. This code path
does not execute TSQL logic to add reverse dependency. As a fix, we have
moved reverse dependency addition to a place which executes during both
normal trigger creation and restore.

Task: BABEL-3249

Signed-off-by: Surendra Vishnoi <vishnosu@amazon.com>

### DIFF
--- a/src/backend/commands/trigger.c
+++ b/src/backend/commands/trigger.c
@@ -1164,6 +1164,15 @@ CreateTriggerFiringOn(CreateTrigStmt *stmt, const char *queryString,
 	referenced.objectSubId = 0;
 	recordDependencyOn(&myself, &referenced, DEPENDENCY_NORMAL);
 
+	/*
+	 * For composite triggers, add dependency from trigger function
+	 * to trigger so that a drop trigger will result in cascade drop
+	 * for function as well. Trigger functions are created as part of
+	 * create trigger for composite triggers.
+	 */
+	if (is_composite_trigger)
+		recordDependencyOn(&referenced, &myself, DEPENDENCY_NORMAL);
+
 	if (isInternal && OidIsValid(constraintOid))
 	{
 		/*


src/backend/commands/trigger.c
```diff
@@ -1164,6 +1164,15 @@ CreateTriggerFiringOn(CreateTrigStmt *stmt, const char *queryString,
...
 	recordDependencyOn(&myself, &referenced, DEPENDENCY_NORMAL);
 
+	/*
+	 * For composite triggers, add dependency from trigger function
+	 * to trigger so that a drop trigger will result in cascade drop
+	 * for function as well. Trigger functions are created as part of
+	 * create trigger for composite triggers.
+	 */
+	if (is_composite_trigger)
+		recordDependencyOn(&referenced, &myself, DEPENDENCY_NORMAL);
```
2jungkook, line 1174: 1. Is there a test case for this?
2. Why we need `DEPENDENCY_NORMAL` instead of `DEPENDENCY_INTERNAL`?

surendravishnoi: yes, I will raise that as part of extension changes.
I am just rearranging this code from extension to here and it was NORMAL dependency always.

rishabhtanwar29 (approved):

2jungkook (approved):

sangilsong (approved):

forestkeeper (approved):

kuntalghosh (approved):

PR #31, opened 2022-06-24
Instead of trigger behaves incorrectly during recursive execution

forestkeeper:
Recursively calling an instead of trigger will result in data loss for last DML on underlying trigger table

### Description

Row inserted by an INSTEAD-OF INSERT trigger is missing in the example below: 

```
-- repro in Babelfish:
1>
2> CREATE TABLE t
3> (
4> c1 int IDENTITY(1,1) PRIMARY KEY,
5> c2 varchar(20) UNIQUE,
6> );
7>
8> INSERT INTO t(c2)
9> VALUES ('Joe'), ('Steve');
10>
11>
12> SELECT * FROM t
13> GO
c1          c2
----------- --------------------
          1 Joe
          2 Steve
1>
2> CREATE TRIGGER TR_t ON t
3> INSTEAD OF INSERT
4> AS
5>  DECLARE @c1 int, @c2 varchar(20)
6> SELECT @c1 = c1, @c2 = c2 FROM inserted
7> INSERT INTO t (c2) VALUES(@c2 + '2') --BUG: This INSERT disappears
8> GO
1>
2> INSERT INTO t(c2) VALUES ('Joe')
3> GO
1>
2> --There should be 3 rows, but there are only 2 rows. The INSERT from the trigger is lost.
3> SELECT * FROM t
4> GO
c1          c2
----------- --------------------
          1 Joe
          2 Steve
```
Explanation: 
- Insert inside instead of trigger causes a recursive trigger execution
- A recursive trigger execution is capped at nesting level of one and should be a no-op.
- We skip both DML and trigger execution but we should execute the original DML as trigger is a no-op.

The same problem will happen for Delete/Update as well.

### Issues Resolved

The fix identifies a recursive instead of trigger and does not skip the DML execution
 
### Check List

### DIFF
--- a/src/backend/commands/trigger.c
+++ b/src/backend/commands/trigger.c
@@ -4861,7 +4861,7 @@ isTsqlInsteadofTriggerExecution(EState *estate, ResultRelInfo *relinfo, TriggerE
 	{
 		Trigger *trigger = &trigdesc->triggers[i];
 		if (TriggerEnabled(estate, relinfo, trigger, event, NULL, NULL, NULL)){
-			return true;
+			return !TsqlRecuresiveCheck(relinfo);
 		}
 	}
 	return false;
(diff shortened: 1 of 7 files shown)


macwoody (dismissed):

surendravishnoi (changes requested):
Please improve commit message

surendravishnoi (approved):

kuntalghosh (commented):
Please mention the extension PR that adds tests for the same PR. Also, please mention the JIRA in the commit message.

sangilsong (approved):

PR #36, opened 2022-07-04
[DO NOT MERGE] Remove restriction, input parameters after one with a default value must also have defaults, for all PL/tsql functions

rishabhtanwar29:
### Description
* PG has a restriction that input parameters after one with
a default value must also have defaults. Previously we
provided a workaround for this by letting user provide
default parameters anywhere but filling NULL as default
value for all parameters after first default parameter.  
* The commit removes this restriction for PL/tsql functions
by also storing the position of default parameters along
with the actual expression so that we can use the position
while looking for the parameter default expression. Introduced
a new structure `FuncDefault` for this purpose.

Task: BABEL-2877
signed-off-by: Rishabh Tanwar <ritanwar@amazon.com>
 
### Check List

### DIFF
--- a/src/backend/catalog/namespace.c
+++ b/src/backend/catalog/namespace.c
@@ -1247,13 +1247,13 @@ FuncnameGetCandidates(List *names, int nargs, List *argnames,
 			for (i = 0; i < pronargs; i++)
 				newResult->args[i] = proargtypes[argnumbers[i]];
 
-			newResult->argdefaults = defaults;
+			newResult->tsql_argdefaults = defaults;
 		}
 		else
 		{
 			/* Simple positional case, just copy proargtypes as-is */
 			memcpy(newResult->args, proargtypes, pronargs * sizeof(Oid));
-			newResult->argdefaults = NIL;
+			newResult->tsql_argdefaults = NIL;
 		}
 		if (variadic)
 		{
@@ -1531,7 +1531,7 @@ MatchNamedCall(HeapTuple proctup, int nargs, List *argnames,
 		char		*str;
 		List		*argdefaults = NIL;
 		bool		isnull;
-		bool		special = false;
+		bool		tsql_funcdefault_node = false;
 		ListCell	*def_item = NULL;
 
 		proargdefaults = SysCacheGetAttr(PROCOID, proctup,
@@ -1543,7 +1543,7 @@ MatchNamedCall(HeapTuple proctup, int nargs, List *argnames,
 			str = TextDatumGetCString(proargdefaults);
 			argdefaults = castNode(List, stringToNode(str));
 			def_item = list_head(argdefaults);
-			special = IsA(lfirst(def_item), FuncDefault) ? true : false;
+			tsql_funcdefault_node = IsA(lfirst(def_item), FuncDefault) ? true : false;
 			pfree(str);
 		}
 		for (pp = numposargs; pp < pronargs; pp++)
@@ -1556,7 +1556,7 @@ MatchNamedCall(HeapTuple proctup, int nargs, List *argnames,
 			 * special handling. Look into FuncDefault into node to find out
 			 * the default expression for pp'th argument.
 			 */
-			if (special)
+			if (tsql_funcdefault_node)
 			{
 				bool found = false;
 
--- a/src/backend/catalog/namespace.c
+++ b/src/backend/catalog/namespace.c
@@ -214,7 +214,7 @@ static void RemoveTempRelationsCallback(int code, Datum arg);
 static void NamespaceCallback(Datum arg, int cacheid, uint32 hashvalue);
 static bool MatchNamedCall(HeapTuple proctup, int nargs, List *argnames,
 						   bool include_out_arguments, int pronargs,
-						   int **argnumbers);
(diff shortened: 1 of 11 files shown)


src/include/nodes/primnodes.h
```diff
@@ -504,6 +504,17 @@ typedef struct FuncExpr
 	int			location;		/* token location, or -1 if unknown */
 } FuncExpr;
 
+
+/*
+ * FuncDefault - expression node for a function's  default argument
+ */
+typedef struct FuncDefault
+{
+	Expr		xpr;
```
2jungkook, line 513: Where is it used?

src/include/nodes/primnodes.h
```diff
@@ -504,6 +504,17 @@ typedef struct FuncExpr
 	int			location;		/* token location, or -1 if unknown */
 } FuncExpr;
 
+
+/*
+ * FuncDefault - expression node for a function's  default argument
+ */
+typedef struct FuncDefault
```
2jungkook, line 511: If it is only used by T-SQL, what about `TsqlFuncDefault`?

src/include/catalog/namespace.h
```diff
@@ -35,6 +35,7 @@ typedef struct _FuncCandidateList
 	int			nvargs;			/* number of args to become variadic array */
 	int			ndargs;			/* number of defaulted args */
 	int		   *argnumbers;		/* args' positional indexes, if named call */
+	List	   *argdefaults;    /* list of default args, only set for PL/tsql function */
```
2jungkook: Using tsql-specific names, e.g., tsql_argdefaults, may help us to resolve conflicts with the community PG in the future.

src/backend/utils/adt/ruleutils.c
```diff
@@ -3273,7 +3273,18 @@ print_function_arguments(StringInfo buf, HeapTuple proctup,
 		if (argname && argname[0])
 			appendStringInfo(buf, "%s ", quote_identifier(argname));
 		appendStringInfoString(buf, format_type_be(argtype));
-		if (print_defaults && isinput && inputargno > nlackdefaults)
+
+		if (nextargdefault != NULL && IsA(lfirst(nextargdefault), FuncDefault))
```
2jungkook: Just wondering which test case is failing without this?

src/backend/catalog/namespace.c
```diff
@@ -1180,6 +1181,47 @@ FuncnameGetCandidates(List *names, int nargs, List *argnames,
 			/* Ignore if it doesn't match requested argument count */
 			if (nargs >= 0 && pronargs != nargs && !variadic && !use_defaults)
 				continue;
+
+			/*
+			 * If call uses all positional arguments, then validate if all
+			 * the remaining arguments have defaults.
+			 */
+			if (use_defaults)
```
2jungkook, line 1189: Can we make a hook for this?

src/backend/catalog/namespace.c
```diff
@@ -1478,14 +1523,70 @@ MatchNamedCall(HeapTuple proctup, int nargs, List *argnames,
...
+		proargdefaults = SysCacheGetAttr(PROCOID, proctup,
+										 Anum_pg_proc_proargdefaults,
+										 &isnull);
+
+		if (!isnull)
+		{
+			str = TextDatumGetCString(proargdefaults);
+			argdefaults = castNode(List, stringToNode(str));
+			def_item = list_head(argdefaults);
+			special = IsA(lfirst(def_item), FuncDefault) ? true : false;
```
2jungkook: nit: is there any better name than `special`?

src/include/nodes/primnodes.h
rishabhtanwar29: This is a common field for all nodes which is needed to get NodeTag.

src/include/catalog/namespace.h
rishabhtanwar29: Yeah makes sense.

src/include/nodes/primnodes.h
rishabhtanwar29: It is right that currently we are only using this node for PL/tsql functions but I am thinking to keep this node and logic very generic so that it can be used for any kind of function.

src/backend/catalog/namespace.c
rishabhtanwar29: Yup, changed!

src/backend/catalog/namespace.c
rishabhtanwar29: I have created `FuncDefault` node to be generic enough so that it can be used for any kind of function (currently only being used for PL/tsql functions though) so that's why I have kept underlying logic in the engine. Moreover, the below logic does not seem tightly coupled to extension as the logic depends on nodeType (FuncDefault) and not on kind of function.

src/backend/utils/adt/ruleutils.c
rishabhtanwar29: Yeah looks like this is already getting handled in `get_rule_expr` and the above logic can be removed.

macwoody (changes requested):
Change metadata format of PG is sensitive. Let's discuss more on the impact and look for alternative solution if needed

PR #38, opened 2022-07-25
Fix MVU failure when a view contains sys.suser_sid()

2jungkook:
If a view from v13.6 contains sys.suser_sid() and its
parameter is not explicitly casted to sys.sysname, then Major Version
Upgrade (MVU) to v14.latest fails. This is because we had assigned typmod for
all domains in T-SQL dialect in v13.x, which was fixed in v14.latest. This is only
problematic when a base type of a sys domain has a fixed typmod, e.g., sys.sysname.

To prevent the MVU failure, this commit adds a new hook removing typmods
for sys.sysname and sys._ci_sysname in gram.y.

Task: BABEL-3402
Signed-off-by: Jungkook Lee <jungkook@amazon.com>

### Check List

### DIFF
--- a/src/backend/parser/gram.y
+++ b/src/backend/parser/gram.y
@@ -12940,6 +12940,11 @@ GenericType:
 					$$ = makeTypeNameFromNameList(lcons(makeString($1), $2));
 					$$->typmods = $3;
 					$$->location = @1;
+
+					const char *dump_restore = GetConfigOption("babelfishpg_tsql.dump_restore", true, false);
+					if (dump_restore && strcmp(dump_restore, "on") == 0 &&
+						strcmp(TypeNameToString($$), "sys.sysname") == 0)
+						$$->typmods = NIL;
 				}
 			| type_function_name attrs opt_type_modifiers WITHOUT TIME ZONE
 				{
--- a/src/backend/parser/gram.y
+++ b/src/backend/parser/gram.y
@@ -12944,7 +12944,7 @@ GenericType:
 					const char *dump_restore = GetConfigOption("babelfishpg_tsql.dump_restore", true, false);
 					if (dump_restore && strcmp(dump_restore, "on") == 0 &&
 						strcmp(TypeNameToString($$), "sys.sysname") == 0)
-						$$->typmods = NIL;
+						$$->typmods = $3;
 				}
 			| type_function_name attrs opt_type_modifiers WITHOUT TIME ZONE
 				{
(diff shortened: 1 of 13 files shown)


src/backend/parser/parse_coerce.c
```diff
@@ -787,6 +787,7 @@ coerce_to_domain(Node *arg, Oid baseTypeId, int32 baseTypeMod, Oid typeId,
 		 strcmp(type_name, "nchar") == 0 ||
 		 strcmp(type_name, "varbinary") == 0 ||
 		 strcmp(type_name, "binary") == 0 ||
+		 strcmp(type_name, "sysname") == 0 ||
```
rishabhtanwar29: This looks incorrect since sysname is not supposed to have typemod.

2jungkook: That was one of my trials....removing this line also didn't help.

src/backend/parser/gram.y
```diff
@@ -12940,6 +12941,8 @@ GenericType:
 					$$ = makeTypeNameFromNameList(lcons(makeString($1), $2));
 					$$->typmods = $3;
 					$$->location = @1;
+
+					fix_tsql_domain_typmods($$);
```
surendravishnoi: will it make sense to have a hook here?

src/backend/parser/gram.y
```diff
@@ -16959,6 +16962,23 @@ check_generic_type_with_or_without_time_zone(const TypeName *typname, bool with_
...
+ * problematic cases and fixes it.
+ */
+static void
+fix_tsql_domain_typmods(TypeName *typname)
+{
+	const char *dump_restore = GetConfigOption("babelfishpg_tsql.dump_restore", true, false);
+	if (!dump_restore || strcmp(dump_restore, "on") != 0)
+		return;
+
+	if (strcmp(strVal(linitial(typname->names)), "sys") == 0 &&
```
surendravishnoi: have we verified that sys.sysname is only domain with this problem?

2jungkook: The problem only happens when a domain’s base type has a fixed typmod, e.g., `varchar(128)`, and a function is created under psql, e.g., `sys.suser_sid(sys.sysname)`, and a customer-created view refers that function. So, as Rishabh mentioned, the only exception would be `sys._ci_sysname` though it is very unlikely...

src/backend/parser/gram.y
2jungkook: will do that

surendravishnoi (approved):

priyansx (approved):
LGTM

PR #43, opened 2022-08-19
Remove restriction, input parameters after one with a default value must also have defaults, for all PL/tsql procedures/functions

rishabhtanwar29:
#### Engine
* PG has a restriction that input parameters after one with a default
value must also have defaults. Previously we provided a workaround
for this by letting user provide default parameters anywhere but
filling NULL as default value for all parameters after first default
parameter.  
* This commit removes this restriction for all PL/tsql functions. We will
be storing the positions of default parameters in a separate babelfish
catalog table (babelfish_function_ext) so that we can use these
positions while looking for the parameter default expression. Introduced
following hooks to achieve the functionality:
  1. `match_pltsql_func_call_hook`: Will help in parsing a PL/tsql function call.
  2. `insert_pltsql_function_defaults_hook`: Will be used to insert
      actual default values in a PL/tsql function call during execution phase.
  3. `print_pltsql_function_arguments_hook`: Will be used to print
      arguments for PL/tsql function depending upon default positions in
      catalog table.

  Additionally, introduced a syscache id `PROCNSPSIGNATURE`
  for new babelfish_function_ext catalog which takes sysdatabase oid,
  logical schema name and function signature as input keys.

#### Extensions
* Implemented all the hooks introduced on engine side.
* Added a sql definition of babelfish_function_ext catalog table in which
all three columns dbid, schema_name and function_signature combined
will be used as a primary key and default_positions column will hold
the list of argument's default positions.

Task: BABEL-2877
signed-off-by: Rishabh Tanwar <ritanwar@amazon.com>
 
### Check List


src/backend/commands/functioncmds.c
```diff
@@ -1347,6 +1350,9 @@ CreateFunction(ParseState *pstate, CreateFunctionStmt *stmt)
 						   procost,
 						   prorows);
 
+		if (store_func_default_positions_hook)
```
macwoody: Could you please try avoid this new hook by using the existing utility hook? 
All other extended catalogs were managed through utility hook

src/include/catalog/namespace.h
```diff
@@ -81,6 +83,12 @@ typedef void (*RangeVarGetRelidCallback) (const RangeVar *relation, Oid relId,
  */
 typedef Oid (*relname_lookup_hook_type) (const char *relname, Oid relnamespace);
 extern PGDLLIMPORT relname_lookup_hook_type relname_lookup_hook;
+typedef bool (*MatchNamedCallHookType) (HeapTuple proctup, int nargs, List *argnames,
+										bool include_out_arguments, int pronargs,
+										int **argnumbers, List **defaults);
+extern PGDLLIMPORT MatchNamedCallHookType MatchNamedCallHook;
+typedef bool (*MatchUnNamedCallHookType) (HeapTuple proctup, int nargs, int pronargs);
```
macwoody: I wonder if it would be better to just create one hook to cover both cases

src/backend/utils/adt/ruleutils.c
```diff
@@ -3175,6 +3177,11 @@ print_function_arguments(StringInfo buf, HeapTuple proctup,
 	{
 		Datum		proargdefaults;
 		bool		isnull;
+		char		*langname = get_language_name(proc->prolang, true);
+
```
macwoody: logically, if we are duplicating most of the lost, it might be better if we inject our logic from the beginning (e.g line 3169)  and only check dialect.

src/include/catalog/namespace.h
```diff
@@ -35,6 +36,7 @@ typedef struct _FuncCandidateList
 	int			nvargs;			/* number of args to become variadic array */
 	int			ndargs;			/* number of defaulted args */
 	int		   *argnumbers;		/* args' positional indexes, if named call */
+	List	   *tsql_argdefaults;    /* list of default args, only set for PL/tsql function */
```
macwoody, line 39: I am hesitate to add new member here, however, I think I understand the purpose and merit.  Let's accept it for the moment.

src/backend/commands/functioncmds.c
rishabhtanwar29: Got it, removed the hook!

src/include/catalog/namespace.h
rishabhtanwar29: It requires a lot of parameters to pass to that single hook, but anyway replaced both the hooks with only one.

src/backend/utils/adt/ruleutils.c
rishabhtanwar29: Dialect check won't work here as this particular function gets called to dump a function in PG dialect, so that's why I am only checking for function's language but I moved this hook to the beginning.

PR #52, opened 2022-10-12
Support UDTS for sequences and ISC-sequences view

aadityavermaa:
- Creation of view information_schema_tsql.sequences to support for T-SQL INFORMATION_SCHEMA.SEQUENCES. 
- Provide support for creation of a sequence using user-defined data types which is currently not supported in Babelfish.
- Added support for cross-schema UDTs

Pull request for changes in extensions repository:
https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/663

### Testing 
No testing required here as the changes are getting tested in the PR on the extensions repository. 

Task: BABEL-3504, BABEL-3452
Signed-off-by: Aditya Verma <adiityav@amazon.com>

### Check List

### DIFF
--- a/src/backend/commands/sequence.c
+++ b/src/backend/commands/sequence.c
@@ -1405,7 +1405,7 @@ init_params(ParseState *pstate, List *options, bool for_identity,
 											   &max_value,
 											   &min_value);
 
-		else newtypid = typenameTypeId(pstate,defGetTypeName(as_type));
+		if(!newtypid) newtypid = typenameTypeId(pstate,defGetTypeName(as_type));
 
 		if (newtypid != INT2OID &&
 			newtypid != INT4OID &&
--- a/src/backend/commands/sequence.c
+++ b/src/backend/commands/sequence.c
@@ -46,7 +46,6 @@
 #include "utils/resowner.h"
 #include "utils/syscache.h"
 #include "utils/varlena.h"
-#include "parser/parser.h"
 
 
 /*
--- a/src/backend/commands/sequence.c
+++ b/src/backend/commands/sequence.c
@@ -1395,6 +1395,12 @@ init_params(ParseState *pstate, List *options, bool for_identity,
 	/* AS type */
 	if (as_type != NULL)
 	{
+		/*
+		 * typenameTypeID is called after the hook to modify the schema name
+		 * internally used within the function when sequence is defined using a
+		 * user-defined data type
+		 */
+
 		Oid			newtypid = 0;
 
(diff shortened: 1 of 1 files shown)


src/backend/commands/sequence.c
```diff
@@ -1401,6 +1401,8 @@ init_params(ParseState *pstate, List *options, bool for_identity,
 											   &max_value,
 											   &min_value);
 
+		if(!newtypid) newtypid = typenameTypeId(pstate,defGetTypeName(as_type));
+
```
HarshLunagariya, line 1415: I think we can avoid engine changes here. Can't we do the same inside the hook?

HarshLunagariya (changes requested):

src/backend/commands/sequence.c
aadityavermaa: Earlier we were getting a "schema does not exist" error when the UDT was defined under a schema, which was originating from the typenameTypeId function. So I have made the related changes inside the hook; but we need to call the function again after the hook else newtypid would be uninitialised when the query is called from PG endpoint.

aadityavermaa:
The reason we need to make these changes on the engine side is because the typenameTypeId function which internally fetches the schema name to return the type oid. But it does not fetch the correct physical schema name in multi-db mode, and hence throws a "schema does not exist" error. Hence, newtypid is initialised via the hook with the correct physical schema name. The condition after the hook is for when a query is executed from PG endpoint.

HarshLunagariya (dismissed):

src/backend/commands/sequence.c
surendravishnoi: this can be handled inside hook, only override newtypid at the end of function, so for every other condition it will retain the original value

KushaalShroff (dismissed):

HarshLunagariya (dismissed):

HarshLunagariya (dismissed):

KushaalShroff (approved):

src/backend/commands/sequence.c
```diff
@@ -1395,7 +1395,13 @@ init_params(ParseState *pstate, List *options, bool for_identity,
 	/* AS type */
 	if (as_type != NULL)
 	{
-		Oid			newtypid = typenameTypeId(pstate, defGetTypeName(as_type));
+		/*
```
surendravishnoi, line 1398: Put comment just before hook call and explain that hook might fail to set newtypid in error cases and we will fall back to typenameTypeId. Also use PR template in commit message

PR #53, opened 2022-10-12
Support TSQL ownership chaining within same DB in procedure

HarshLunagariya:
### Description

Added pg_*_aclcheck hooks for supporting TSQL ownership chaining. These hooks will:
- check whether current statement execution is inside procedure or not by looking at the PLtsql_exec_state stack.
- If it is inside procedure then it compares the referenced object and procedure's owner.
- If both are same then this hook bypasses the permission check and allows execution of that statement.

Task: BABEL-3014
Signed-off-by: Harsh Lunagariya <lunharsh@amazon.com>
 
### Issues Resolved
BABEL-3014
 
### Check List

### DIFF
--- a/src/backend/executor/execExpr.c
+++ b/src/backend/executor/execExpr.c
@@ -50,6 +50,7 @@
 #include "utils/lsyscache.h"
 #include "utils/typcache.h"
 
+is_execstate_inside_tsql_proc_hook_type is_execstate_inside_tsql_proc_hook = NULL;
 
 typedef struct LastAttnumInfo
 {
@@ -2464,7 +2465,7 @@ ExecInitFunc(ExprEvalStep *scratch, Expr *node, List *args, Oid funcid,
 
 	/* Check permission to call function */
 	aclresult = pg_proc_aclcheck(funcid, GetUserId(), ACL_EXECUTE);
-	if (aclresult != ACLCHECK_OK)
+	if ((is_execstate_inside_tsql_proc_hook && !(*is_execstate_inside_tsql_proc_hook)(OBJECT_FUNCTION, funcid)) && aclresult != ACLCHECK_OK)
 		aclcheck_error(aclresult, OBJECT_FUNCTION, get_func_name(funcid));
 	InvokeFunctionExecuteHook(funcid);
 
--- a/src/backend/executor/execExpr.c
+++ b/src/backend/executor/execExpr.c
@@ -50,7 +50,6 @@
 #include "utils/lsyscache.h"
 #include "utils/typcache.h"
 
-is_execstate_inside_tsql_proc_hook_type is_execstate_inside_tsql_proc_hook = NULL;
 
 typedef struct LastAttnumInfo
 {
@@ -2465,7 +2464,7 @@ ExecInitFunc(ExprEvalStep *scratch, Expr *node, List *args, Oid funcid,
 
 	/* Check permission to call function */
 	aclresult = pg_proc_aclcheck(funcid, GetUserId(), ACL_EXECUTE);
-	if ((is_execstate_inside_tsql_proc_hook && !(*is_execstate_inside_tsql_proc_hook)(OBJECT_FUNCTION, funcid)) && aclresult != ACLCHECK_OK)
+	if ((check_ownership_chaining_for_tsql_proc_hook && !(*check_ownership_chaining_for_tsql_proc_hook)(OBJECT_FUNCTION, funcid)) && aclresult != ACLCHECK_OK)
 		aclcheck_error(aclresult, OBJECT_FUNCTION, get_func_name(funcid));
 	InvokeFunctionExecuteHook(funcid);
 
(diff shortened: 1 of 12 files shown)


xhfanhe (dismissed):

src/backend/executor/execMain.c
```diff
@@ -576,6 +576,12 @@ ExecCheckRTPerms(List *rangeTable, bool ereport_on_violation)
 		RangeTblEntry *rte = (RangeTblEntry *) lfirst(l);
 
 		result = ExecCheckRTEPerms(rte);
```
macwoody: I wonder if it makes more sense if add the hook in this function? 
Since it is one returning true and false. 
And for our case, we simplify want to return true in certain cases.

src/backend/executor/execExpr.c
```diff
@@ -2464,7 +2464,7 @@ ExecInitFunc(ExprEvalStep *scratch, Expr *node, List *args, Oid funcid,
 
 	/* Check permission to call function */
 	aclresult = pg_proc_aclcheck(funcid, GetUserId(), ACL_EXECUTE);
-	if (aclresult != ACLCHECK_OK)
+	if (!(check_ownership_chaining_for_tsql_proc_hook && (*check_ownership_chaining_for_tsql_proc_hook)(OBJECT_FUNCTION, funcid)) && aclresult != ACLCHECK_OK)
```
macwoody: pg_proc_aclcheck is used a lot , I wonder why only these two cases are considered. 
One of the benefit I could see if we make it in a lower place is that there is existing hook for us to do additional aclcheck

HarshLunagariya: Actually, we are considering only the entry point of procedure/function in normal cases.
1. But yes, we might need to add this same at the other places as well. But for now, Based on what we've tested this was sufficient.
2. We can't add it in pg_proc_aclcheck() directly - Because system function which returns permission like `has_function_privilege(username, functionoid, privilege)` info will behave incorrectly - while executing these function they'll call this ownership chaining hook and if `has_function_privilege(username, functionoid, privilege)`'s owner and function(which has oid = functionoid)'s owner are same they'll return ACLCHECK_OK/TRUE which is not correct

src/backend/executor/execMain.c
xhfanhe: I think it makes no difference because `ExecCheckRTEPerms` is only called at this place.

src/backend/executor/execExpr.c
macwoody: 1. do we know what would happen for those missing places? 
2. has_function_privilege is a PG built-in function, how can it trigger your logic? Could you please share an example when it could go wrong? 

Adding a feature specific hook and maintaining it would be make us pay the cost for every future major versions. I think we need to think thoroughly before adding a new one

src/backend/executor/execMain.c
macwoody: It is not about how many times a function is called, it is about violation of original intention of each individual function. 

my understanding is that 
ExecCheckRTEPerms determine privilege check result 
ExecCheckRTPerms acts on the result by throwing an error or calling additional hook.

HarshLunagariya: Yes, it makes sense to move this check in ExecCheckRTEPerms. Will address this change in next revision.

src/backend/executor/execExpr.c
HarshLunagariya: 1. For those missing places, if somehow we encounter it then it will exhibit the normal behaviour( w/o any ownership chaining). This will be inconsistent as compared to normal TSQL behaviour.
2. Let's say user A called a system view VW which uses has_function_privilege(let's consider one of the variation described as follows):
```
Datum
has_function_privilege_name(PG_FUNCTION_ARGS)
{
    text       *functionname = PG_GETARG_TEXT_PP(0);
    text       *priv_type_text = PG_GETARG_TEXT_PP(1);
    Oid         roleid;
    Oid         functionoid;
    AclMode     mode;
    AclResult   aclresult;

    roleid = GetUserId();
    functionoid = convert_function_name(functionname);
    mode = convert_function_priv_string(priv_type_text);

    aclresult = pg_proc_aclcheck(functionoid, roleid, mode);

    PG_RETURN_BOOL(aclresult == ACLCHECK_OK);
}
```
- Now when we'll execute `select * from VW` then it will try to execute `has_function_privilege(some_pg_internal_object)`(as user A).
- While executing it we'll hit the check_ownership_chaining hook in pg_proc_aclcheck() - if we add this check directly to *_aclcheck function's according to your suggestion.
- Hook(inside the pg_proc_aclcheck function call) will see that function `has_function_privilege_name`'s owner is the same as that internal object's owner so it will form the ownership chaining and returns true as the result of has_function_privilege(some_pg_internal_object)'s result even if user A doesn't have rights for it.

xhfanhe (dismissed):

macwoody (dismissed):

xhfanhe (approved):

PR #55, opened 2022-10-25
Improve identity value insert performance with IDENTITY_INSERT=on

kuntalghosh:
### Description
With IDENTITY_INSERT=on in a table, users are allowed to insert values in identity columns. We also need to update the last used identity for the table(seed) so that subsequent insertion with default identity can start from the same seed. As per the current approach, after inserting one row, we're doing a full scan of the table to find out the min/max identity value and updating these variables. This creates a huge performance bottleneck for data insertion and the performance degrades significantly as the number of rows increases in the table.

In T-SQL, the last used identity of a table is non-transactional. That means, it always increases (or decreases) irrespective of the underlying transaction commits/rollbacks. To implement this, we don't have to scan the entire table. We can just calculate the max(current seed, value to be inserted) and update the seed under the same lock.

Task: BABEL-3506
Signed-off-by: Kuntal Ghosh <kuntalgh@amazon.com>

### Functional Testing
Please check the extension PR for the same - https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/728

### Performance Testing
Please check the extension PR for the same - https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/728

### DIFF
--- a/src/backend/commands/sequence.c
+++ b/src/backend/commands/sequence.c
@@ -47,7 +47,6 @@
 #include "utils/syscache.h"
 #include "utils/varlena.h"
 
-#include "parser/parser.h"		/* only needed for GUC variables */
 
 /*
  * We don't want to log each fetching of a value from a sequence,
@@ -92,6 +91,7 @@ pltsql_sequence_validate_increment_hook_type pltsql_sequence_validate_increment_
 pltsql_sequence_datatype_hook_type pltsql_sequence_datatype_hook = NULL;
 pltsql_nextval_hook_type pltsql_nextval_hook = NULL;
 pltsql_resetcache_hook_type pltsql_resetcache_hook = NULL;
+pltsql_setval_hook_type pltsql_setval_hook = NULL;
 
 static HTAB *seqhashtab = NULL; /* hash table for SeqTable items */
 
@@ -934,8 +934,7 @@ do_setval(Oid relid, int64 next, bool iscalled)
 	HeapTuple	pgstuple;
 	Form_pg_sequence pgsform;
 	int64		maxv,
-				minv,
-				seqincrement;
+				minv;
 
 	/* open and lock sequence */
 	init_sequence(relid, &elm, &seqrel);
@@ -952,7 +951,6 @@ do_setval(Oid relid, int64 next, bool iscalled)
 	pgsform = (Form_pg_sequence) GETSTRUCT(pgstuple);
 	maxv = pgsform->seqmax;
 	minv = pgsform->seqmin;
-	seqincrement = pgsform->seqincrement;
 	ReleaseSysCache(pgstuple);
 
 	/* read-only transactions may only modify temp sequences */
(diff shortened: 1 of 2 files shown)


src/backend/commands/sequence.c
```diff
@@ -966,6 +969,23 @@ do_setval(Oid relid, int64 next, bool iscalled)
...
 	seq = read_seq_tuple(seqrel, &buf, &seqdatatuple);
 
+	/*
+	 * BABELFISH: In T-SQL, with identity_insert=on, sequence value is set as
+	 * max(value to be inserted, last used sequence value) when increment is
+	 * positive. Min value is set with a negative increment.
+	 * We need to calculate the max/min and set the value without releasing
+	 * the lock so that other backends can't overwrite the value concurrently.
+	 */
+	if (sql_dialect == SQL_DIALECT_TSQL)
```
2jungkook: First of all, could you introduce a hook to minimize the engine changes? At least, we don't need to add `parser.h` and `seqincrement`.

Secondly, we should think more carefully on the dialect check.
1. What would be our expectation when a user tries to insert in PG connection against a T-SQL table? You may need to sync up with Di regarding those kind of inter-ops behavior.
2. In the future, we would want to support data dump and restore. Suppose we have two tables in the BBF database, t1 created from PG connection and t2 created from TDS connection. In that case, how can we distinguish those two tables?

kuntalghosh: > First of all, could you introduce a hook to minimize the engine changes? At least, we don't need to add parser.h and seqincrement.

You're right. I should add a function hook instead.

> Secondly, we should think more carefully on the dialect check.
> 
> 1. What would be our expectation when a user tries to insert in PG connection against a T-SQL table? You may need to sync up with Di regarding those kind of inter-ops behavior.

These changes only affect whether the sequence seed update should be transactional or not in TSQL dialect. As per the existing implementation,
1. From TDS connection, if you insert an identity column, the seed update is transactional. That means if the insert rollbacks, the seed will not be updated with the new value. This is not the expected T-SQL behaviour.
2. Internally, this is implemented as 'GENERATED ALWAYS AS IDENTITY'. From PG connection, you're not allowed to insert directly in an identity column. It'll throw an error. 
But, there are ways to insert identity value from PG connection as well - a) using OVERRIDING SYSTEM VALUE with INSERT b) using COPY. But, both these options don't update the seed by default. User has to manually update the seed using setval function. In fact, pg_dump uses the same approach for dumping identity columns. An example:

```
-- Table created from TDS connection
CREATE TABLE t1(a int IDENTITY, b int);
INSERT INTO t1 VALUES(1);

-- the pg_dump script for the table (pg_dump jdbc_testdb --table=master_dbo.t1 --column-inserts)
CREATE TABLE master_dbo.t1 (
    a integer NOT NULL,
    b integer
);
ALTER TABLE master_dbo.t1 ALTER COLUMN a ADD GENERATED ALWAYS AS IDENTITY (
    SEQUENCE NAME master_dbo.t1_a_seq
    START WITH 1
    INCREMENT BY 1
    NO MINVALUE
    NO MAXVALUE
    CACHE 1
);
INSERT INTO master_dbo.t1 (a, b) OVERRIDING SYSTEM VALUE VALUES (1, 1);
SELECT pg_catalog.setval('master_dbo.t1_a_seq', 1, false);
```

pg_dump calculates the seed as following:
```
select max(a) from master_dbo.t1
```
and generates the setval statement in the dump.

In this patch, there will be no behavioural changes in PG connection (as a result pg_dump as well). Only from TDS connection, the seed update will become non-transactional similar to T-SQL and PG.

> 2. In the future, we would want to support data dump and restore. Suppose we have two tables in the BBF database, t1 created from PG connection and t2 created from TDS connection. In that case, how can we distinguish those two tables?

As I explained above, with these changes, the behaviour will be now same for TDS and PG connections. So, I don't think we need to distinguish the connection type at least from this feature's perspective.

2jungkook (dismissed):

2jungkook (approved):

PR #69, opened 2022-12-08
Create table or proc should throw error for declaring non-permissible datalength in variable length datatypes; Fix tests related to Numeric datatype in REL3_X_DEV branch

bsatarupa:
### Description
PG15 has introduced a community change for numeric datatype as below:
Formerly, when specifying NUMERIC(precision, scale), the scale had to be in the range [0, precision], which was per SQL spec. This commit extends the range of allowed scales to [-1000, 1000], independent of
the precision (whose valid range remains [1, 1000]). A negative scale implies rounding before the decimal point. For
example, a column might be declared with a scale of -3 to round values to the nearest thousand. Note that the display scale remains non-negative, so in this case the display scale will be zero, and all digits before the decimal point will be displayed.
A scale greater than the precision supports fractional values with zeros immediately after the decimal point.

From SQL spec aspect which means checking explicitly whether 0 <= scale <= precision and 1 <= precision <= 38, which was earlier done by underlying PG code for any numeric/decimal input data.

Along with the above change, it is found that all variable length datatypes like numeric, decimal, time, datetime2, datetimeoffset need explicit checking on their permissible datalength at the time of table or stored procedure creation and generate error if required.
Before fix, Babelfish was not complaining for the above behaviour.

Task: BABEL-3720, BABEL-3793
Signed-off-by: Satarupa Biswas [satarupb@amazon.com](mailto:satarupb@amazon.com)
 
### Check List

### DIFF
--- a/src/backend/parser/parse_type.c
+++ b/src/backend/parser/parse_type.c
@@ -31,6 +31,7 @@ static int32 typenameTypeMod(ParseState *pstate, const TypeName *typeName,
 							 Type typ);
 
 check_or_set_default_typmod_hook_type check_or_set_default_typmod_hook = NULL;
+validate_var_datatype_scale_hook_type validate_var_datatype_scale_hook = NULL;
 
 /*
  * LookupTypeName
@@ -412,6 +413,13 @@ typenameTypeMod(ParseState *pstate, const TypeName *typeName, Type typ)
 					 parser_errposition(pstate, typeName->location)));
 		datums[n++] = CStringGetDatum(cstr);
 	}
+
+	/*
+	 * Checks whether variable length datatypes like numeric, decimal, time, datetime2, datetimeoffset
+	 * are declared with permissible datalength at the time of table or stored procedure creation
+	 */
+	if (validate_var_datatype_scale_hook)
+			(*validate_var_datatype_scale_hook)(typeName, typ);
 
 	/* hardwired knowledge about cstring's representation details here */
 	arrtypmod = construct_array(datums, n, CSTRINGOID,
--- a/src/include/parser/parse_type.h
+++ b/src/include/parser/parse_type.h
@@ -61,4 +61,11 @@ extern void parseTypeString(const char *str, Oid *typeid_p, int32 *typmod_p, boo
 typedef void (*check_or_set_default_typmod_hook_type)(TypeName * typeName, int32 *typmod, bool is_cast);
 extern PGDLLIMPORT check_or_set_default_typmod_hook_type check_or_set_default_typmod_hook;
 
+/*
+ * Hook to check whether variable length datatypes like numeric, decimal, time, datetime2, datetimeoffset
+ * are declared with permissible datalength at the time of table or stored procedure creation
+ */
+typedef void (*validate_var_datatype_scale_hook_type)(const TypeName *typeName, Type typ);
+extern PGDLLIMPORT validate_var_datatype_scale_hook_type validate_var_datatype_scale_hook;
+
 #endif							/* PARSE_TYPE_H */


src/backend/parser/parse_type.c
```diff
@@ -384,7 +388,9 @@ typenameTypeMod(ParseState *pstate, const TypeName *typeName, Type typ)
 
 			if (IsA(&ac->val, Integer))
 			{
-				cstr = psprintf("%ld", (long) intVal(&ac->val));
```
rishabhtanwar29: I would suggest not to change this line and just add our logic directly (this will reduce future merge conflicts):
```
scale[count] = intVal(&ac->val);
count++;
```

src/backend/parser/parse_type.c
```diff
@@ -413,6 +419,26 @@ typenameTypeMod(ParseState *pstate, const TypeName *typeName, Type typ)
 		datums[n++] = CStringGetDatum(cstr);
 	}
 
+	dataTypeName = TypeNameToString(typeName);
+	/*
```
rishabhtanwar29: **Just a suggestion:** Instead of this swapping logic, can we directly call the hook depending upon the condition? Something like:
```
if (verify_valid_scale_in_variable_length_datatypes_hook)
{
    char *dataTypeName = TypeNameToString(typeName);
    if (strstr(dataTypeName, "numeric") || strstr(dataTypeName, "decimal"))
         (*verify_valid_scale_in_variable_length_datatypes_hook)(dataTypeName, precision, scale);
    else
         (*verify_valid_scale_in_variable_length_datatypes_hook)(dataTypeName, scale, precision);
}
```

src/backend/parser/parse_type.c
bsatarupa: Alright

src/backend/parser/parse_type.c
bsatarupa: Alright!

src/backend/parser/parse_type.c
```diff
@@ -413,6 +420,24 @@ typenameTypeMod(ParseState *pstate, const TypeName *typeName, Type typ)
...
+	 */
+	if (validate_var_datatype_scale_hook)
+	{
+		dataTypeName = TypeNameToString(typeName);
+		
+		/*
+		 * Since numeric/decimal datatype stores precision in scale[0] and scale in scale[1];
+		 * for all other datatypes, scale is stored in scale[0]
+		 */
+		if (strstr(dataTypeName, "numeric") || strstr(dataTypeName, "decimal"))
```
rishabhtanwar29: I think we should not use `strstr` here as it can easily match any user defined domain name which consists `numeric` or `decimal` as substring. Better to use `typ` to validate the schema as well as the whole typename. In that case we can directly pass `typ` to the hook which will take care of verifying the typenames as well.

src/backend/parser/parse_type.c
```diff
@@ -413,6 +420,24 @@ typenameTypeMod(ParseState *pstate, const TypeName *typeName, Type typ)
 		datums[n++] = CStringGetDatum(cstr);
 	}
 
+	/*
+	 * Checks whether variable length datatypes like numeric, decimal, time, datetime2, datetimeoffset
+	 * are declared with permissible datalength at the time of table or stored procedure creation
+	 */
+	if (validate_var_datatype_scale_hook)
```
surendravishnoi, line 421: We should put all the code inside hook, even getting scale value from typmods. Also, please see how PG uses type name for similar logic. Using str* functions can be risky. We should follow and reuse PG logic about how they interpret typeName->names

bsatarupa: done

rishabhtanwar29 (approved):

surendravishnoi (approved):

PR #74, opened 2022-12-29
Improve security definer for procedures/functions by avoid Lookup for GUC values

shah-nirmit:
### Description

Inside function `fmgr_security_definer` function there were frequent calls to `set_config_option/GetConfigOption` API's for `sql_dialect` GUC which would need Cache Lookup, which is unnecessary for the dialect. Changed the implementation to directly set/get the GUC value using the global variable.

Draft PR on Extension Repo: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/977
### Issues Resolved

BABEL-2994
 
### Test Scenarios Covered ###
* **Use case based -**
Already Existing Test Cases

* **Boundary conditions -**

* **Arbitrary inputs -**

* **Negative test cases -**

* **Minor version upgrade tests -**

* **Major version upgrade tests -**

* **Performance tests -**
There would be a general performance for procedures,functions which use the FMGR framework
Here is one specific scenario
 A test database populated with 15k dummy procedures and executed the following query
`exec sp_sproc_columns  @procedure_name='sp_statistics' , @ODBCVer=3, @fUsePattern=0`
We had the following results
```
Without the change
Network packet size (bytes): 4096
50 xact[s]:
Clock Time (ms.): total     16516  avg   330.3 (3.0 xacts per sec.)

vs 

With the change
Network packet size (bytes): 4096
50 xact[s]:
Clock Time (ms.): total      9335  avg   186.7 (5.4 xacts per sec.)
```
Explanation:-For sp_sproc_columns, we he have created an internal view sp_sproc_columns_view for all procedures. The view itself has a lot of cast functions; mostly for each column  and each row for example if we cast to string types then varchar function is called using FMGR framework.
Later we select only one row for particular input based on procedure_name, but even there there’s a mismatch in datatype length so varchar function is called for one column, each row.

Attached the perf reports 
[perf-without](https://user-images.githubusercontent.com/44708230/210231957-03ceaee9-f0f5-413d-af37-27786f46b8f0.svg)
[perf-with-2994](https://user-images.githubusercontent.com/44708230/210231963-4c6bd6cf-5dd0-4726-a0d1-6516e285a086.svg)

* **Tooling impact -**

* **Client tests -**

### DIFF
--- a/src/backend/utils/fmgr/fmgr.c
+++ b/src/backend/utils/fmgr/fmgr.c
@@ -870,8 +870,10 @@ fmgr_security_definer(PG_FUNCTION_ARGS)
 		 * design for PG vs TSQL procedures.
 		 */
 		if (set_sql_dialect)
+		{
 			sql_dialect = sql_dialect_value_old;
 			assign_sql_dialect(sql_dialect_value_old, newextra);
+		}
 
 		PG_RE_THROW();
 	}
--- a/src/backend/utils/fmgr/fmgr.c
+++ b/src/backend/utils/fmgr/fmgr.c
@@ -710,10 +710,11 @@ fmgr_security_definer(PG_FUNCTION_ARGS)
 	bool		set_sql_dialect;
 	int			sql_dialect_value;
 	int			sql_dialect_value_old;
-	int			pg_dialect = 0;
-	int			tsql_dialect = 1;
+	int			pg_dialect = SQL_DIALECT_PG;
+	int			tsql_dialect = SQL_DIALECT_TSQL;
 	int			sys_func_count = 0;
 	int			non_tsql_proc_count = 0;
+	void	   *newextra = NULL;
 
 	if (!fcinfo->flinfo->fn_extra)
 	{
@@ -824,6 +825,7 @@ fmgr_security_definer(PG_FUNCTION_ARGS)
 		ReleaseSysCache(tuple);
 		sql_dialect_value_old = sql_dialect;
 		sql_dialect = sql_dialect_value;
+		assign_sql_dialect(sql_dialect_value, newextra);
 
 	}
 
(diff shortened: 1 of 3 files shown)


src/backend/utils/misc/guc.c
```diff
@@ -4990,6 +4990,9 @@ static struct config_enum ConfigureNamesEnum[] =
 		NULL, NULL, NULL
 	},
 
+/* When changing this guc value directly without set_config_option,The function
```
surendravishnoi: please use
/*
   comment
*/

src/backend/utils/fmgr/fmgr.c
```diff
@@ -821,14 +823,10 @@ fmgr_security_definer(PG_FUNCTION_ARGS)
...
 		ReleaseSysCache(tuple);
-		sql_dialect_value_old = GetConfigOption("babelfishpg_tsql.sql_dialect", true, true);
-		set_config_option("babelfishpg_tsql.sql_dialect", sql_dialect_value,
-						  (superuser() ? PGC_SUSET : PGC_USERSET),
-						  PGC_S_SESSION,
-						  GUC_ACTION_SAVE,
-						  true,
-						  0,
-						  false);
+		sql_dialect_value_old = sql_dialect;
```
surendravishnoi, line 826: do we have some simple test case which shows benefit of this change? Ideally, it should be a function/proc where cross dialect calls are done in each statement and the statements are executed in a loop

surendravishnoi (commented):
Use PR template

src/backend/utils/misc/guc.c
shah-nirmit: Fixed in the latest commit

KushaalShroff:
> Use PR template

@shah-nirmit Please add one or more of the scenarios where this was a bottleneck and the perfomance impact % numbers for them

surendravishnoi (dismissed):

src/backend/utils/fmgr/fmgr.c
shah-nirmit: Addressed in PR description

KushaalShroff (approved):

PR #82, opened 2023-01-12
Refactor utility function to be used by both SUM and AVG Aggregates

shah-nirmit:
### Description
Refactor the previously introduced helper function into a common utility (`bigint_utility`) function which can be used by both SUM and AVG  aggregates for BIGINT datatype
 
### Issues Resolved
BABEL-3507

#### Extension PR
https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1045

 Signed-off-by: Nirmit Shah <nirmisha@amazon.com>
### Check List

### DIFF
--- a/src/backend/utils/adt/numeric.c
+++ b/src/backend/utils/adt/numeric.c
@@ -5906,6 +5906,9 @@ numeric_poly_sum(PG_FUNCTION_ARGS)
 #endif
 }
 
+/*
+ *	Final function for BIGINT datatype for TSQL Aggregates (SUM,AVG)
+ */
 Datum
 bigint_utility(FunctionCallInfo fcinfo, bool isSum)
 {
--- a/src/backend/utils/adt/numeric.c
+++ b/src/backend/utils/adt/numeric.c
@@ -5910,7 +5910,7 @@ numeric_poly_sum(PG_FUNCTION_ARGS)
  *	Final function for BIGINT datatype for TSQL Aggregates (SUM,AVG)
  */
 Datum
-bigint_utility(FunctionCallInfo fcinfo, bool isSum)
+bigint_utility(FunctionCallInfo fcinfo, int tsqlAggType)
 {
 
 	PolyNumAggState		*state;
@@ -5938,8 +5938,9 @@ bigint_utility(FunctionCallInfo fcinfo, bool isSum)
 						errmsg("Arithmetic overflow error converting expression to data type bigint.")));
 		}
 		else {
-			if (isSum)
+			if (tsqlAggType == TSQL_SUM)
 				PG_RETURN_INT64((int64) result);
+			/* If the aggregate type is TSQL_AVG */
 			else
 			{
 				result /= state->N;
@@ -5947,8 +5948,9 @@ bigint_utility(FunctionCallInfo fcinfo, bool isSum)
 			}
 		}
 	#else
-		if (isSum)
+		if (tsqlAggType == TSQL_SUM)
 			temp = numeric_sum(fcinfo);
+		/* If the aggregate type is TSQL_AVG */
 		else
 			temp = numeric_avg(fcinfo);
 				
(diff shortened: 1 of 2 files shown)


src/backend/utils/adt/numeric.c
```diff
@@ -5934,10 +5937,21 @@ bigint_poly_sum(PG_FUNCTION_ARGS)
 					(errcode(ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE),
 						errmsg("Arithmetic overflow error converting expression to data type bigint.")));
 		}
-		else
-			PG_RETURN_INT64((int64) result);
+		else {
+			if (isSum)
+				PG_RETURN_INT64((int64) result);
+			else
```
surendravishnoi, line 5947: add a comment to make clear that else here means avg only

src/backend/utils/adt/numeric.c
```diff
@@ -5906,8 +5906,11 @@ numeric_poly_sum(PG_FUNCTION_ARGS)
 #endif
 }
 
+/*
+ *	Final function for BIGINT datatype for TSQL Aggregates (SUM,AVG)
+ */
```
Deepesh125: Add more explanation about function here.

Deepesh125:
> Refactor the previously introduced helper function into a common utility function which can be used by both SUM and AVG aggregates for BIGINT datatype

Add details about which function are you referring to in the description.

Deepesh125 (changes requested):

src/backend/utils/adt/numeric.c
shah-nirmit: Fixed in the latest commit

src/backend/utils/adt/numeric.c
```diff
@@ -5906,8 +5906,11 @@ numeric_poly_sum(PG_FUNCTION_ARGS)
...
 Datum
-bigint_poly_sum(PG_FUNCTION_ARGS)
+bigint_utility(FunctionCallInfo fcinfo, int tsqlAggType)
```
2jungkook: 1. The current function name is a little bit general. Can we use a more specific name, e.g., `bigint_poly_aggr` or `bigint_aggr_final`?
2. Instead of implementing our own logic, can we leverage `numeric_poly_sum` and `numeric_poly_avg` in the function body? We may call the existing function and then check overflow.

shah-nirmit: Regarding point 2:-
1. In T-Sql, The average aggregates  doesn't calculate the average of values first and  then checks for bounds but rather sums up the values and checks if its out of bound , if not then it proceeds in calculating the average, So to match this behaviour we have introduced our own logic
2.  If we use numeric_poly_sum / numeric_poly_average we would be get the value/result back in numeric which will again need additional processing, So I didn't use them in the common utility function

src/backend/utils/adt/numeric.c
shah-nirmit: Fixed in the latest commit

src/backend/utils/adt/numeric.c
shah-nirmit: Fixed 1. in latest commit

src/backend/utils/adt/numeric.c
```diff
@@ -5906,8 +5906,14 @@ numeric_poly_sum(PG_FUNCTION_ARGS)
...
 
+/*
+ *	Final function for BIGINT datatype for TSQL Aggregates (SUM,AVG).
+ *  The tsql aggregate return non-default return types when compared with 
+ *  Default aggregate for integer datatypes , this function takes the accumlated
+ *  State and tranform the result into return type expected by tsql.
+ */
 Datum
-bigint_poly_sum(PG_FUNCTION_ARGS)
+bigint_poly_aggr_final(FunctionCallInfo fcinfo, int tsqlAggType)
```
surendravishnoi: use enum type instead of int

shah-nirmit: Fixed in the latest commit

Deepesh125 (approved):

surendravishnoi (approved):

PR #99, opened 2023-02-07
Create table or proc should throw error for declaring non-permissible…

shalinilohia50:
### Description

It is found that all variable length datatypes like numeric, decimal, time, datetime2, datetimeoffset need explicit checking on their permissible datalength at the time of table or stored procedure creation and generate error if required. Before fix, Babelfish was not complaining for the above behaviour.

Task: BABEL-3793
Co-authored-by: Satarupa Biswas satarupb@amazon.com
Signed-off-by: Shalini Lohia <lshalini@amazon.com>

 
### Issues Resolved
BABEL-3793


jsudrik (approved):
Identical to 3_X changes. Approved.

PR #108, opened 2023-03-01
Support database-level dump/restore

rishabhtanwar29:
### Description
Added support for dumping a single Babelfish logical database. Changes are
summarized in following bullet points:
1. Added a new option `--bbf-database-name` in both pg_dumpall and pg_dump to
let user provide the name of logical database to be dumped.
2. Fetch all the physical schema names corresponding to the given logical database
and add all of them into `schema_include_patterns` list. With this, existing dump
logic will make sure to only dump all these schemas as well as all the objects
contained by these schemas.
3. Function `bbf_selectDumpableTableData` takes care of explicitly marking all
Babelfish catalog table data to be dumped.
4. We need to selectively dump the Babelfish catalog table data corresponding to
given logical database which is being handled by function `getCursorForBbfCatalogTableData`
for `COLUMN-INSERT` and `fixCopyCommand` for `COPY`. These function prepares
query using joins and conditions to filter the data.
5. We skip dumping the `dbid` column as it might conflict with the existing database's `dbid`
so we generate the new `dbid` during restore. It has been handled for both `COLUMN-INSERT`
and `COPY` options:
 a. `COPY`: Implemented a hook `fill_missing_values_in_copyfrom_hook` which will use the
 sequence to generate the new `dbid`.  
 b. `COLUMN-INSERT`: Used existing `pre_parse_analyze_hook` hook to handle this case.
5. Added a new file `dumpall_babel_utils.c` to implement Babelfish specific logic to
selectively dump users. New file is needed since existing `dump_babel_utils.c` can't
be used with pg_dumpall utility.
7. Logins do not belong to a single database so we will not be dumping the logins so it and
we will only dump default users of the database (dbo, db_owner and db_guest).
Due to this `CREATE` and `INSERT` might fail for certain logins but it should be fine.
8. Current implementation does not dump foreign servers and foreign data wrappers as
these are global objects and does not belong to a single logical database.
9. **Note**: A logical database with certain migration mode (single-db/multi-db) can only be 
restored on the Babelfish server with the same migration mode. Which means a single-db
database can be restored only on single-db Babelfish server, same thing applies to multi-db.
Although the builtin databases like `master` can be restored on any kind of server.

Task: BABEL-3870
 
### Check List

### DIFF
--- a/src/bin/pg_dump/dump_babel_utils.c
+++ b/src/bin/pg_dump/dump_babel_utils.c
@@ -34,8 +34,11 @@
 static const CatalogId nilCatalogId = {0, 0};
 static char *escaped_bbf_db_name = NULL;
 static int bbf_db_id = 0;
+static SimpleOidList catalog_table_include_oids = {NULL, NULL};
 
 static char *getMinOid(Archive *fout);
+static bool isBabelfishConfigTable(TableInfo *tbinfo);
+static void addFromClauseForLogicalDatabaseDump(PQExpBuffer buf, TableInfo *tbinfo, bool is_builtin_db);
 
 static char *
 getMinOid(Archive *fout)
@@ -109,6 +112,27 @@ isBabelfishDatabase(Archive *fout)
 }
 
 /*
+ * isBabelfishConfigTable:
+ * Returns true if given table is a configuration table (for which catalog
+ * table data needs to be dumped), false otherwise.
+ * It is mainly used during Babelfish logical database dump.
+ */
+static bool
+isBabelfishConfigTable(TableInfo *tbinfo)
+{
+	/* We don't want to dump logins during logical database dump */
+	if (tbinfo && tbinfo->dobj.namespace &&
+		strcmp(tbinfo->dobj.name, "babelfish_authid_login_ext") == 0)
+		return false;
+
+	if (catalog_table_include_oids.head != NULL &&
+		simple_oid_list_member(&catalog_table_include_oids, tbinfo->dobj.catId.oid))
+		return true;
+
+	return false;
+}
+
+/*
  * dumpBabelGUCs:
  * Dumps Babelfish specific GUC settings if current
  * database is a Babelfish database.
@@ -123,10 +147,13 @@ dumpBabelGUCs(Archive *fout)
 		return;
 
 	qry = createPQExpBuffer();
-	oid = getMinOid(fout);
 	appendPQExpBufferStr(qry, "SET babelfishpg_tsql.dump_restore = TRUE;\n");
-	appendPQExpBuffer(qry, "SET babelfishpg_tsql.dump_restore_min_oid = %s;\n", oid);
-	free(oid);
+	if (fout->dopt->binary_upgrade)
+	{
+		oid = getMinOid(fout);
+		appendPQExpBuffer(qry, "SET babelfishpg_tsql.dump_restore_min_oid = %s;\n", oid);
+		free(oid);
+	}
 
 	ArchiveEntry(fout, nilCatalogId, createDumpId(),
 				 ARCHIVE_OPTS(.tag = "BABELFISHGUCS",
@@ -173,7 +200,7 @@ bbf_selectDumpableCast(CastInfo *cast)
 /*
  * bbf_selectDumpableTableData:
  * Marks Babelfish catalog table data to be dumped if not in binary-upgrade mode.
- * It is mostly used during Babelfish logical database dump as none of the extensions
+ * It is mainly used during Babelfish logical database dump as none of the extensions
  * are marked to be dumped so catalog table data explicitly need to be marked as dumpable.
  */
 void
@@ -182,17 +209,8 @@ bbf_selectDumpableTableData(TableInfo *tbinfo, Archive *fout)
 	if (!isBabelfishDatabase(fout) || fout->dopt->binary_upgrade)
 		return;
 
-	if (tbinfo && tbinfo->dobj.namespace &&
-		strcmp(tbinfo->dobj.namespace->dobj.name, "sys") == 0 &&
-		(strcmp(tbinfo->dobj.name, "babelfish_sysdatabases") == 0 ||
-		 strcmp(tbinfo->dobj.name, "babelfish_namespace_ext") == 0 ||
-		 strcmp(tbinfo->dobj.name, "babelfish_function_ext") == 0 ||
-		 strcmp(tbinfo->dobj.name, "babelfish_authid_user_ext") == 0 ||
-		 strcmp(tbinfo->dobj.name, "babelfish_view_def") == 0 ||
-		 strcmp(tbinfo->dobj.name, "babelfish_domain_mapping") == 0))
-	{
+	if (isBabelfishConfigTable(tbinfo))
 		tbinfo->dobj.dump |= DUMP_COMPONENT_DATA;
-	}
 }
 
 /*
@@ -656,9 +674,21 @@ prepareForLogicalDatabaseDump(Archive *fout, SimpleStringList *schema_include_pa
 
 	bbf_db_id = atooid(PQgetvalue(res, 0, PQfnumber(res, "dbid")));
 	PQclear(res);
-	destroyPQExpBuffer(query);
+	resetPQExpBuffer(query);
 
-	query = createPQExpBuffer();
+	/* Get oids of all the Babelfish catalog configuration tables */
+	appendPQExpBufferStr(query,
+						 "SELECT unnest(extconfig)::oid "
+						 "FROM pg_catalog.pg_extension "
+						 "WHERE extname = 'babelfishpg_tsql';");
+	res = ExecuteSqlQuery(fout, query->data, PGRES_TUPLES_OK);
+	ntups = PQntuples(res);
+	for (i = 0; i < ntups; i++)
+		simple_oid_list_append(&catalog_table_include_oids, atooid(PQgetvalue(res, i, 0)));
+
+	PQclear(res);
+	resetPQExpBuffer(query);
+
 	/* Get all the physical schema names from sys.babelfish_namespace_ext with given dbid */
 	appendPQExpBuffer(query,
 					  "SELECT pg_catalog.quote_ident(nspname) AS nspname "
@@ -676,9 +706,8 @@ prepareForLogicalDatabaseDump(Archive *fout, SimpleStringList *schema_include_pa
 	{
 		char *schema_name;
 
-		schema_name = pg_strdup(PQgetvalue(res, i, PQfnumber(res, "nspname")));
+		schema_name = PQgetvalue(res, i, PQfnumber(res, "nspname"));
 		simple_string_list_append(schema_include_patterns, schema_name);
-		pfree(schema_name);
 	}
 
 	PQclear(res);
@@ -686,11 +715,11 @@ prepareForLogicalDatabaseDump(Archive *fout, SimpleStringList *schema_include_pa
 }
 
 /*
- * getBabelfishDependencies:
+ * setBabelfishDependenciesForLogicalDatabaseDump:
  * Sets required dependencies for babelfish objects.
  */
 void
-getBabelfishDependencies(Archive *fout)
+setBabelfishDependenciesForLogicalDatabaseDump(Archive *fout)
 {
 	PQExpBuffer		query;
 	PGresult		*res;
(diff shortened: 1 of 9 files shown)


src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -541,3 +612,326 @@ updateExtConfigArray(Archive *fout, char ***extconfigarray, int nconfigitems)
...
+
+	/*
+	 * Add all physical schemas corresponding to the logical database into
+	 * schema_include_patterns so that we dump only those schemas.
+	 */
+	for (i = 0; i < ntups; i++)
+	{
+		char *schema_name;
+
+		schema_name = pg_strdup(PQgetvalue(res, i, PQfnumber(res, "nspname")));
```
surendravishnoi: why to create a dup string if we immediately free it?

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -541,3 +612,326 @@ updateExtConfigArray(Archive *fout, char ***extconfigarray, int nconfigitems)
...
+
+	resetPQExpBuffer(buf);
+	appendPQExpBufferStr(buf, "DECLARE _pg_dump_cursor CURSOR FOR SELECT ");
+	*nfields = 0;
+	for (i = 0; i < tbinfo->numatts; i++)
+	{
+		if (tbinfo->attisdropped[i])
+			continue;
+		/* Skip dbid column, we will generate new database id during restore. */
+		if (!is_builtin_db && strcmp(tbinfo->attnames[i], "dbid") == 0)
```
surendravishnoi: why buildin_db is exception here?

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -124,6 +170,31 @@ bbf_selectDumpableCast(CastInfo *cast)
...
+ */
+void
+bbf_selectDumpableTableData(TableInfo *tbinfo, Archive *fout)
+{
+	if (!isBabelfishDatabase(fout) || fout->dopt->binary_upgrade)
+		return;
+
+	if (tbinfo && tbinfo->dobj.namespace &&
+		strcmp(tbinfo->dobj.namespace->dobj.name, "sys") == 0 &&
+		(strcmp(tbinfo->dobj.name, "babelfish_sysdatabases") == 0 ||
```
surendravishnoi: how do we make sure that a new catalog table or schema change is also taken care of in utility as well?

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -91,6 +108,35 @@ isBabelfishDatabase(Archive *fout)
...
+	char		*oid;
+	PQExpBuffer	qry;
+
+	if (!isBabelfishDatabase(fout))
+		return;
+
+	qry = createPQExpBuffer();
+	oid = getMinOid(fout);
+	appendPQExpBufferStr(qry, "SET babelfishpg_tsql.dump_restore = TRUE;\n");
+	appendPQExpBuffer(qry, "SET babelfishpg_tsql.dump_restore_min_oid = %s;\n", oid);
```
surendravishnoi: why are we dumping min oid?

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -541,3 +612,326 @@ updateExtConfigArray(Archive *fout, char ***extconfigarray, int nconfigitems)
...
+	PQExpBuffer	q;
+	int			i;
+	bool		is_builtin_db = false;
+	bool		needComma = false;
+
+	if (!isBabelfishDatabase(fout) || bbf_db_name == NULL)
+		return;
+
+	if (tbinfo->dobj.namespace == NULL ||
+		strcmp(tbinfo->dobj.namespace->dobj.name, "sys") != 0 ||
```
surendravishnoi: catalog table list based decision are done at multiple places. Put them in a separate function

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -541,3 +612,326 @@ updateExtConfigArray(Archive *fout, char ***extconfigarray, int nconfigitems)
...
+		appendPQExpBuffer(copyBuf, "COPY %s (%s) FROM stdin;\n",
+						  fmtQualifiedDumpable(tbinfo),
+						  q->data);
+	else
+	{
+		appendPQExpBuffer(copyBuf, "COPY (SELECT %s ",
+						  q->data);
+		destroyPQExpBuffer(q);
+
+		if (strcmp(tbinfo->dobj.name, "babelfish_sysdatabases") == 0)
```
surendravishnoi: see if this logic can be shared with above function

src/bin/pg_dump/dumpall_babel_utils.c
```diff
@@ -0,0 +1,129 @@
...
+	char	*escaped_bbf_db_name;
+	bool	is_builtin_db = false;
+
+	if (!bbf_db_name)
+		return;
+
+	is_builtin_db = (pg_strcasecmp(bbf_db_name, "master") == 0 ||
+			pg_strcasecmp(bbf_db_name, "tempdb") == 0 ||
+			pg_strcasecmp(bbf_db_name, "msdb") == 0)
+			? true : false;
```
surendravishnoi, line 46: does anything at all is required for built in db?

src/bin/pg_dump/pg_dump.c
```diff
@@ -413,6 +416,7 @@ main(int argc, char **argv)
 		{"on-conflict-do-nothing", no_argument, &dopt.do_nothing, 1},
 		{"rows-per-insert", required_argument, NULL, 10},
 		{"include-foreign-data", required_argument, NULL, 11},
+		{"bbf-database-name", required_argument, NULL, 12},
```
surendravishnoi: this is problematic as community can use 12 in future and we have to resolve it every time. May be use a large number to leave sufficient buffer for community

src/bin/pg_dump/dump_babel_utils.c
2jungkook: Dumping min oid would be only needed for binary_upgrade.

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -541,3 +612,326 @@ updateExtConfigArray(Archive *fout, char ***extconfigarray, int nconfigitems)
...
+	PQclear(res);
+	destroyPQExpBuffer(query);
+}
+
+/*
+ * getBabelfishDependencies:
+ * Sets required dependencies for babelfish objects.
+ */
+void
+getBabelfishDependencies(Archive *fout)
```
2jungkook: Usually, `get` means returning something. In this case, I think `set` or `prepare` could be a better naming. Also, I'd suggest to add a suffix like `ForLogicalDatabaseDump` to make its use case more clear.

src/bin/pg_dump/pg_dump.c
```diff
@@ -2166,6 +2181,7 @@ dumpTableData_insert(Archive *fout, const void *dcontext)
 	if (tdinfo->filtercond)
 		appendPQExpBuffer(q, " %s", tdinfo->filtercond);
 
+	getCursorForBbfCatalogTableData(fout, tbinfo, q, &nfields);
```
2jungkook: Rather than `get`, I think this is more like `fix` or `reset` because it resets `PQExpBuffer` and `nfields`. Same arguments can be applied for other function names.

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -541,3 +612,326 @@ updateExtConfigArray(Archive *fout, char ***extconfigarray, int nconfigitems)
...
+						  fmtQualifiedDumpable(tbinfo), escaped_bbf_db_name, escaped_bbf_db_name,
+						  escaped_bbf_db_name);
+		/*
+		 * Builtin db users will already be present in the target
+		 * server so no need to dump their catalog data.
+		 */
+		if (is_builtin_db)
+			appendPQExpBufferStr(buf, "LIMIT 0 ");
+	}
+	else
```
2jungkook: Seems like it is only for `babelfish_domain_mapping`. Let's make it explicit and error out for the `else` block.

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -541,3 +612,326 @@ updateExtConfigArray(Archive *fout, char ***extconfigarray, int nconfigitems)
...
+			pg_strcasecmp(bbf_db_name, "tempdb") == 0 ||
+			pg_strcasecmp(bbf_db_name, "msdb") == 0)
+			? true : false;
+
+	resetPQExpBuffer(buf);
+	appendPQExpBufferStr(buf, "DECLARE _pg_dump_cursor CURSOR FOR SELECT ");
+	*nfields = 0;
+	for (i = 0; i < tbinfo->numatts; i++)
+	{
+		if (tbinfo->attisdropped[i])
```
2jungkook: nit; for future safety, let's add the corresponding logic or throw errors for generated columns, e.g.,
```
		if (tbinfo->attgenerated[i] && dopt->column_inserts)
			continue;
		if (tbinfo->attgenerated[i])
			appendPQExpBufferStr(q, "NULL");
		else
			appendPQExpBufferStr(q, fmtId(tbinfo->attnames[i]));
		attgenerated[nfields] = tbinfo->attgenerated[i];
```

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -541,3 +612,326 @@ updateExtConfigArray(Archive *fout, char ***extconfigarray, int nconfigitems)
...
+			appendPQExpBufferStr(q, "a.");
+		appendPQExpBufferStr(q, fmtId(tbinfo->attnames[i]));
+		needComma = true;
+	}
+
+	resetPQExpBuffer(copyBuf);
+	if (isFrom)
+		appendPQExpBuffer(copyBuf, "COPY %s (%s) FROM stdin;\n",
+						  fmtQualifiedDumpable(tbinfo),
+						  q->data);
```
2jungkook, line 934: may need to destroy `q`?

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: Yeah right, I didn't realize this while moving this piece of code from pg_dump.c to here.

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: Missed this while refactoring the code.

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: Actually the idea for the function name came from original pg_dump function [getDependencies](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/BABEL_3_X_DEV__PG_15_X/src/bin/pg_dump/pg_dump.c#L17759) but I too agree with your suggestion of using `set`.

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: Yup

src/bin/pg_dump/pg_dump.c
rishabhtanwar29: Makes sense.

src/bin/pg_dump/pg_dump.c
rishabhtanwar29: Changed it 30 which I think should be good since it's very unlikely for dump utility to get 19 more options.

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: In case of builtin db, we will dump the dbid column since we don't need to regenerate it during restore as DBIDs are fixed for builtin dbs. updated the comment as well.

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: Done.

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: Added a helper function for it.

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: Done!

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: Done!

src/bin/pg_dump/dumpall_babel_utils.c
rishabhtanwar29: Actually we don't want to dump the built users/roles as they we would already be present in the target server.

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: I have added a new helper function `isBabelfishConfigTable` which actually checks if the current table is extension configuration table or not. So now we are not hard-coding the table names. But for filtering logic, we would have to add new logic in pg_dump utility for new catalog tables.

src/bin/pg_dump/dumpall_babel_utils.c
surendravishnoi: I meant it is necessary to return query with LIMIT 0 or can't we return an empty string or equivalent of no-op query

src/bin/pg_dump/dumpall_babel_utils.c
```diff
@@ -0,0 +1,129 @@
...
+						  escaped_bbf_db_name, escaped_bbf_db_name);
+
+		/* builtin db users will already be present in the target server so no need to dump them */
+		if (is_builtin_db)
+			appendPQExpBufferStr(buf, "LIMIT 0");
+	}
+}
+
+/*
+ * Returns query to fetch al the roles, members and grantors related
```
surendravishnoi: al -> all

surendravishnoi (dismissed):

src/bin/pg_dump/dumpall_babel_utils.c
rishabhtanwar29: Yeah I think there can be other ways to do this since LIMIT 0 is actually not necessary here but since LIMIT 0 has constant execution time (no performance impact as such) so it looked simpler to use and is more readable. We can't send empty string as we execute that string/query later.

src/bin/pg_dump/dumpall_babel_utils.c
rishabhtanwar29: Fixed!

2jungkook (dismissed):

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -121,8 +121,9 @@ static bool
 isBabelfishConfigTable(TableInfo *tbinfo)
 {
 	/* We don't want to dump logins during logical database dump */
-	if (tbinfo && tbinfo->dobj.namespace &&
-		strcmp(tbinfo->dobj.name, "babelfish_authid_login_ext") == 0)
+	if (tbinfo->relkind != RELKIND_RELATION ||
```
surendravishnoi: if we are checking for tbinfo being not null, it should come first

rishabhtanwar29: Fixed!

surendravishnoi (approved):

PR #119, opened 2023-03-31
[BABEL-3865] Add pg_depend, pg_shdepend to ENR for temp tables

time-change:
### Description

Add pg_depend and pg_shdepend to ENR for temp tables in TSQL. Additionally, add logic to avoid ENR usage for temp tables that may have dependencies. 

### Issues Resolved

Fix is part of TV rework, and a requirement for transactional semantics fixes. 
 
### Check List


src/backend/catalog/heap.c
```diff
@@ -1142,8 +1183,11 @@ heap_create_with_catalog(const char *relname,
 		/*
 		 * in TSQL, temporary table name should start with '#'.
 		 * If temporary table name does not start with '#', assume it is a PG temporary table.
-		 * This can happen in the case of internal query to create PG temporary table for Babelfish */
-		if (relname != 0 && strlen(relname) >= 1 && relname[0] == '#')
+		 * This can happen in the case of internal query to create PG temporary table for Babelfish
+		 *
+		 * Also skip ENR for cases where dependent objects might be involved.
+		 */
+		if (relname != 0 && strlen(relname) >= 1 && relname[0] == '#' && !CheckTempTableHasDependencies(tupdesc))
```
surendravishnoi, line 1190: How are we covering table variables? for them table name starts with "@"? Also, comment is not accurate as we are only skipping when dependent objects are user defined.

src/backend/catalog/heap.c
```diff
@@ -1067,6 +1067,47 @@ AddNewRelationType(const char *typeName,
...
+			if (typtup->typtype != TYPTYPE_DOMAIN)
+			{
+				continue;
+			}
+
+			SysCacheGetAttr(TYPEOID, tup, Anum_pg_type_typacl, &isnull);
+			if (!isnull)
+			{
+				ReleaseSysCache(tup);
+				return 1;
```
surendravishnoi, line 1103: why not true/false?

src/backend/utils/misc/queryenvironment.c
```diff
@@ -317,6 +321,48 @@ bool ENRgetSystableScan(Relation rel, Oid indexId, int nkeys, ScanKey key, List
...
+			{
+				ListCell   *lc;
+				foreach(lc, enr->md.cattups[ENR_CATTUP_DEPEND]) {
+					Form_pg_depend tup = (Form_pg_depend) GETSTRUCT((HeapTuple) lfirst(lc));
+					if (indexId == DependDependerIndexId &&
+						tup->classid == (Oid)v1 &&
+						tup->objid == (Oid)v2) {
+						*tuplist = enr->md.cattups[ENR_CATTUP_DEPEND];
+						*tuplist_i = foreach_current_index(lc);
+						return true;
```
lejaokri, line 334: Let's say we have scan key, v1 = 1, v2 = 2 and the list has the following data [(v1=3, v2=1), (v1=1, v2=2), (v1=4, v2=5)] then this logic will find the match at index 1: (v1=1, v2=2) and it will assume anything after that also match the scan key which is wrong because the item at index 2 doesnt match the scan key. 

Try this scenario:

```
dev-dsk-klejao-2b-ae2f4aac % sqlcmd -S localhost -U jdbc_user -P 1 -e -i ~/test.sql
create table #temp(c1 int)
drop table #temp

Msg 3723, Level 16, State 1, Server BABELFISH, Line 2
cannot drop table "#temp" because type "#temp" requires it
```

src/backend/catalog/heap.c
```diff
@@ -1067,6 +1067,47 @@ AddNewRelationType(const char *typeName,
...
+ */
+static int CheckTempTableHasDependencies(TupleDesc tupdesc)
+{
+	int i;
+	int	natts = tupdesc->natts;
+
+	for (i = 0; i < natts; i++)
+	{
+		Oid atttypid = TupleDescAttr(tupdesc, i)->atttypid;
+		HeapTuple tup = SearchSysCache1(TYPEOID, ObjectIdGetDatum(atttypid));
```
lejaokri, line 1088: Please verify if you need to call `ReleaseSysCache(tup);`

src/backend/catalog/heap.c
```diff
@@ -1067,6 +1067,47 @@ AddNewRelationType(const char *typeName,
...
+
+		if (HeapTupleIsValid(tup))
+		{
+			bool isnull;
+			Form_pg_type typtup = (Form_pg_type) GETSTRUCT(tup);
+			if (typtup->typtype != TYPTYPE_DOMAIN)
+			{
+				continue;
```
lejaokri, line 1096: missing `ReleaseSysCache(tup);` before `continue`?

src/backend/catalog/heap.c
lejaokri, line 1190: What about if we declare temporary tables with unique constraint / primary key. I tested some scenarios and keep hitting a Trap in the logfile. I was able to isolate it to the following scenario:

Scenario
```
1> CREATE TABLE #temp(a INT UNIQUE NOT NULL, b INT, c CHAR(15) DEFAULT 'Whoops')
2> go
1> quit
```

In the logfile:
```
TRAP: FailedAssertion("!TTS_EMPTY(slot)", File: "execTuples.c", Line: 1650, PID: 121854)
postgres: jdbc_user jdbc_testdb 172.22.162.176(48261) idle(ExceptionalCondition+0xb9)[0xb36533]
postgres: jdbc_user jdbc_testdb 172.22.162.176(48261) idle(ExecFetchSlotHeapTuple+0x60)[0x763d5a]
postgres: jdbc_user jdbc_testdb 172.22.162.176(48261) idle(systable_recheck_tuple+0x2a)[0x525c17]
postgres: jdbc_user jdbc_testdb 172.22.162.176(48261) idle[0x5c6fa7]
postgres: jdbc_user jdbc_testdb 172.22.162.176(48261) idle(performDeletion+0x6e)[0x5c65fa]
postgres: jdbc_user jdbc_testdb 172.22.162.176(48261) idle[0x5df365]
postgres: jdbc_user jdbc_testdb 172.22.162.176(48261) idle[0x5df3a5]
postgres: jdbc_user jdbc_testdb 172.22.162.176(48261) idle(shmem_exit+0x8b)[0x95c6e1]
postgres: jdbc_user jdbc_testdb 172.22.162.176(48261) idle[0x95c5b7]
postgres: jdbc_user jdbc_testdb 172.22.162.176(48261) idle(proc_exit+0x15)[0x95c50e]
postgres: jdbc_user jdbc_testdb 172.22.162.176(48261) idle(PostgresMain+0xcdc)[0x99528f]
/local/home/klejao/pg14.latest/postgres/lib/babelfishpg_tds.so(+0x2dea1)[0x7f48e540bea1]
postgres: jdbc_user jdbc_testdb 172.22.162.176(48261) idle[0x8d1bc9]
postgres: jdbc_user jdbc_testdb 172.22.162.176(48261) idle[0x8d1554]
postgres: jdbc_user jdbc_testdb 172.22.162.176(48261) idle[0x8cdba8]
postgres: jdbc_user jdbc_testdb 172.22.162.176(48261) idle(PostmasterMain+0x123f)[0x8cd15a]
postgres: jdbc_user jdbc_testdb 172.22.162.176(48261) idle[0x7d0e82]
```

src/backend/utils/misc/queryenvironment.c
```diff
@@ -263,7 +265,9 @@ bool ENRgetSystableScan(Relation rel, Oid indexId, int nkeys, ScanKey key, List
 				reloid != AttributeRelationId && 
 				reloid != ConstraintRelationId && 
 				reloid != StatisticRelationId &&
-				reloid != StatisticExtRelationId)
+				reloid != StatisticExtRelationId &&
+				reloid != DependRelationId &&
+				reloid != SharedDependRelationId)
 		return NULL;
```
surendravishnoi, line 271: right now there is fallback for not supported reloid to use pg catalog. We should disable the fallback and detect cases which are not handled. This way we make sure that ENR is all or nothing for a given table.

src/backend/utils/misc/queryenvironment.c
```diff
@@ -465,6 +511,44 @@ bool ENRgetSystableScan(Relation rel, Oid indexId, int nkeys, ScanKey key, List
 	return false;
```
surendravishnoi, line 511: we should remove this fallback

src/backend/utils/misc/queryenvironment.c
```diff
@@ -500,7 +584,51 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
...
+							if (tf1->classid == tf2->classid &&
+								tf1->objid == tf2->objid &&
+								tf1->objsubid == tf2->objsubid) { 
+								lc = curlc;
+								break;
+							}
+						}
+						ret = true;
+					}
+					break;
```
surendravishnoi, line 606: remove fallback here and similar places in the function

src/backend/utils/misc/queryenvironment.c
```diff
@@ -465,6 +511,44 @@ bool ENRgetSystableScan(Relation rel, Oid indexId, int nkeys, ScanKey key, List
...
+	QueryEnvironment		*queryEnv = currentQueryEnv;
+	ListCell   *curlc;
+
+	while (queryEnv)
+	{
+		switch (catalog_oid) {
+			case RelationRelationId:
+				return get_ENR_withoid(queryEnv, oid);
+			case TypeRelationId:
+				foreach(curlc, queryEnv->namedRelList) {
```
surendravishnoi, line 525: we need to see if this is performant when there are large number of ENR objects

src/backend/catalog/dependency.c
```diff
@@ -1351,7 +1351,8 @@ deleteOneObject(const ObjectAddress *object, Relation *depRel, int flags)
 
 	while (HeapTupleIsValid(tup = systable_getnext(scan)))
 	{
-		CatalogTupleDelete(*depRel, &tup->t_self);
+		if (!ENRdropTuple(*depRel, tup))
```
surendravishnoi, line 1354: we can change code style to not use fallbacks rather know upfront which catalog to use

lejaokri: There is a problem with the way systable_getnext handles ENR while deleting something from the list.

```
520     if (sysscan->enr)
521     {
...
525         if (sysscan->enr_tuplist && sysscan->enr_tuplist_i < sysscan->enr_tuplist->length)
526         {
527             htup = lfirst(list_nth_cell(sysscan->enr_tuplist, sysscan->enr_tuplist_i++));
528         }
529     }
```

Consider the case when enr->tuplist = [1, 2, 3, 4, 5, 6, 7, 8], ie length = 8 and enr_tup_list_i = 0
Now we call line 527, so it returns item at index 0 and it does enr_tup_list_i++.

Next we call ENRdropTuple to remove the item from the list.
Now list becomes [2, 3, 4, 5,  6, 7,  8] and enr_tup_list_i = 1 so next iteration of the while loop will return item at index 1 which is 3.

I think we should do this instead

```
if (!scan->enr) {
   while (HeapTupleIsValid(tup = systable_getnext(scan)))
       CatalogTupleDelete(*depRel, &tup->t_self);
} else {
   // drop everything from scan->enr_tuplist
}
```

lejaokri: @surendravishnoi is correct. I hit one crash with this. `deleteOneObject` caller is recursive so the object we're trying to delete here may already been drop so `ENRdropTuple` would return false here so we would call `CatalogTupleDelete` which would crash because tuple is not in catalog.

We can do this instead

```
if (scan->enr)
            ENRdropTuple(*depRel, tup);
else
            CatalogTupleDelete(*depRel, &tup->t_self);
```

src/backend/utils/misc/queryenvironment.c
lejaokri: I've fixed this in https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/130

src/backend/catalog/heap.c
lejaokri: I've fixed this in https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/130

src/backend/catalog/dependency.c
lejaokri: I've fixed this in https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/130

time-change:
This PR is being combined with Kristian's change in https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/130

PR #120, opened 2023-03-31
In CTE Null first order behavior is not correct

forestkeeper:
Previously we add the Null first check after analyze, and in that case it'll miss to add the Null first tag to the sort clause for a CTE query

change the place adding Null first tag to the place sortClause is initialized

Task: BABEL-3991
Signed-off-by: Zhibai Song <szh@amazon.com>

### Description

the following sql not produce the right order for NULL : 
```
drop table if exists t1;
create table t1(a int);
insert t1 values (1);
insert t1 values (3);
insert t1 values (null);
with t1cte AS (
select top(3) a from t1 order by 1
)
select * from t1cte; -- Returns NULL at bottom not expected
select top(3) a from t1 order by 1; -- Returns NULL at the TOP as expected
```

the root cause is we deal with sort clause inproperly

### Issues Resolved

Will return the right order for null include 
- asc order by return NULL first
- no order by , natural asc, return NULL last
- desc order by return NULL last

checked this behavior with 
- simple select
- select order by
- subquery order by 
- cte with inner query order by
- babel_functions
- etc other remainning explain query

### Check List
- [ x ] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/src/backend/parser/parse_clause.c
+++ b/src/backend/parser/parse_clause.c
@@ -54,6 +54,8 @@
 
 tle_name_comparison_hook_type  tle_name_comparison_hook = NULL;
 
+sort_nulls_first_hook_type  sort_nulls_first_hook = NULL;
+
 static int	extractRemainingColumns(ParseNamespaceColumn *src_nscolumns,
 									List *src_colnames,
 									List **src_colnos,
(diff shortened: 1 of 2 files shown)


src/backend/parser/parse_clause.c
```diff
@@ -3383,6 +3385,9 @@ addTargetToSortList(ParseState *pstate, TargetEntry *tle,
 			case SORTBY_NULLS_DEFAULT:
 				/* NULLS FIRST is default for DESC; other way for ASC */
 				sortcl->nulls_first = reverse;
+				if (sort_nulls_first_hook){
+					sort_nulls_first_hook(sortcl, reverse);
```
2jungkook: 1. Please follow the coding convention about brackets.
2. nulls_first is a BBF-specific behavior so a better hook name could be `sort_nulls_default_hook`?

forestkeeper: How about sort_nullsfirst_default_hook ?

since it's a behavior controls whether nullsfirst is default or not ?

2jungkook (approved):

PR #131, opened 2023-04-18
Incorrect metadata for sql_variant datatype on a restored server

aadityavermaa:
### Description
When pg_dump/restore is performed on database with table containing sql_variant data type columns, it was observed that the metadata seemed to get lost during the dump process. Therefore, the values were not correctly restored, which reflected in the failure of some test cases in JDBC framework. This change aims to separately handle table containing sql_variant data type columns so that the values can be correctly dumped and restored. 
 
Tests added in extensions PR: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1385
### Issues Resolved

BABEL-4036 
Signed-off-by: Aditya Verma <adiityav@amazon.com>
### Check List

### DIFF
--- a/src/bin/pg_dump/dump_babel_utils.c
+++ b/src/bin/pg_dump/dump_babel_utils.c
@@ -39,6 +39,8 @@ static SimpleOidList catalog_table_include_oids = {NULL, NULL};
 static char *getMinOid(Archive *fout);
 static bool isBabelfishConfigTable(TableInfo *tbinfo);
 static void addFromClauseForLogicalDatabaseDump(PQExpBuffer buf, TableInfo *tbinfo, bool is_builtin_db);
+static void castSqlvariantToBasetype(PGresult* res, Archive* fout, PQExpBuffer q, int tuple, int field);
+static int getMbstrlen(const char* mbstr,Archive* fout);
 
 static char *
 getMinOid(Archive *fout)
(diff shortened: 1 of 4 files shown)


src/bin/pg_dump/pg_dump.c
```diff
@@ -320,7 +320,7 @@ static void appendReloptionsArrayAH(PQExpBuffer buffer, const char *reloptions,
 									const char *prefix, Archive *fout);
 static char *get_synchronized_snapshot(Archive *fout);
 static void setupDumpWorker(Archive *AHX);
-static TableInfo *getRootTableInfo(const TableInfo *tbinfo);
+// static TableInfo *getRootTableInfo(const TableInfo *tbinfo);
```
surendravishnoi: please remove it

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,435 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+	DumpOptions *dopt = fout->dopt;
+	PQExpBuffer q = createPQExpBuffer();
+	PQExpBuffer insertStmt = NULL;
+	char	   *attgenerated;
+	bool	   *atttypnames;
+	PGresult   *res;
+	int			nfields,
+				i;
+	int			rows_per_statement = dopt->dump_inserts;
+	int			rows_this_statement = 0;
```
rishabhtanwar29: Nit: indentation seems a bit off.

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,435 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+	/*
+	 * If we're going to emit INSERTs with column names, the most efficient
+	 * way to deal with generated columns is to exclude them entirely.  For
+	 * INSERTs without column names, we have to emit DEFAULT rather than the
+	 * actual column value --- but we can save a few cycles by fetching nulls
+	 * rather than the uninteresting-to-us value.
+	 */
+	attgenerated = (char *) pg_malloc(tbinfo->numatts * sizeof(char));
+	atttypnames = (bool *) pg_malloc(tbinfo->numatts * sizeof(bool));
+	atttypnames = memset(atttypnames, false, tbinfo->numatts * sizeof(bool));
```
rishabhtanwar29: Should use pg_malloc0 to allocate and initialize with all false. Also, name seems to be confusing, let's name it something like `is_sqlvaraint`.

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,435 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+				archputs(" (", fout);
+			else if (rows_this_statement > 0)
+				archputs(",\n\t(", fout);
+			else
+				archputs("\n\t(", fout);
+
+			orig_field = 0;
+			for (int field = 0; field < nfields_new; field++)
+			{
+				if(is_sqlvariant){
```
rishabhtanwar29: Can't understand the use of `is_sqlvariant` here, we already have built an array `atttypnames` to know if it is a sql_variant column right?

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,435 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+		mbstr+= PQmblen(mbstr, fout->encoding);
+		len++;
+	}
+	return len;
+}
+
+/*
+ * castSqlvariantToBasetype:
+ * Cast sql_variant column values to their basetype with corresponding metadata
+ * so that the data can be restored correctly
```
rishabhtanwar29: This doesn't seem to be explaining the actual functionality. Let's clearify how exactly this function casts a sql_variant value.

src/bin/pg_dump/pg_dump.c
```diff
@@ -2439,7 +2439,12 @@ dumpTableData(Archive *fout, const TableDataInfo *tdinfo)
 	/* We had better have loaded per-column details about this table */
 	Assert(tbinfo->interesting);
 
-	if (dopt->dump_inserts == 0)
+	if(hasSqlvariantColumn(tbinfo)){
```
rishabhtanwar29: Let's also add `isBabelfishDatabase` check here or in `hasSqlvariantColumn` function.

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,430 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+					return true;
+	return false;
+}
+
+/*
+ * getMbstrlen:
+ * returns the length of a multibyte string
+ */
+static int
+getMbstrlen(const char* mbstr,Archive* fout)
```
surendravishnoi: space after ,

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,430 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+{
+	int len = 0;
+
+	while (*mbstr){
+		mbstr+= PQmblen(mbstr, fout->encoding);
```
surendravishnoi: space after mbstr

src/bin/pg_dump/dump_babel_utils.c
surendravishnoi: we should check for mbstr being null for safety

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,430 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+					is_sqlvariant_column[nfields_new] = true;
+					nfields_new = nfields_new + 2;
+			}
+		}
+		attgenerated[nfields] = tbinfo->attgenerated[i];
+		nfields++; nfields_new++;
+	}
+
+	/* Servers before 9.4 will complain about zero-column SELECT */
+	if (nfields == 0)
```
surendravishnoi: babel will not have use case for servers before 9.4

src/bin/pg_dump/dump_babel_utils.c
aadityavermaa: Fixed in latest commit. Thanks.

src/bin/pg_dump/dump_babel_utils.c
aadityavermaa: Replaced `pg_malloc` and `memset` to `pg_malloc0`. I've changed the name of the array from atttypnames to is_sqlvariant_column.

src/bin/pg_dump/dump_babel_utils.c
aadityavermaa: The bool variable was to check in each iteration if the previous column is a sql_variant column. But since we have a boolean array to determine that already, it is redundant to have this. I've updated in the latest commit. Thanks.

src/bin/pg_dump/dump_babel_utils.c
aadityavermaa: Updated the Description.

src/bin/pg_dump/pg_dump.c
aadityavermaa: Added another check in the if condition.

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,430 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+/*
+ * dumpSqlvariantTableData:
+ * Dump the contents of a table with atleast one column of sql_variant data
+ * type. We use insert commands to dump the table data irrespective of value of
+ * dopt->dump_inserts. This function is derived from dumpTableData_insert in
+ * pg_dump.c modified to handle tables with sql_variant data type columns
+ * separately.
+ */
+int
+dumpSqlvariantTableData(Archive *fout, const void *dcontext)
```
surendravishnoi: Copying whole function for babel might be problematic in future as same code maintained at two places. Have we looked into doing in place changes (with putting SQL variant specific logic in smaller functions wherever applicable)?

surendravishnoi (commented):
Please add SQL variant SME to reviewers as well

src/bin/pg_dump/pg_dump.c
```diff
@@ -2240,12 +2261,18 @@ dumpTableData_insert(Archive *fout, const void *dcontext)
...
 							appendPQExpBufferStr(insertStmt, ", ");
 						appendPQExpBufferStr(insertStmt,
-											 fmtId(PQfname(res, field)));
+							fmtId(PQfname(res, field)));
+						/*
+						 * skip over columns defined additionally for
+						 * sql_variant data type columns
+						 */
+						if (is_sqlvariant_column[field])
+								field=field+2;
```
surendravishnoi: use space on both side of =

src/bin/pg_dump/dump_babel_utils.c
aadityavermaa: Added changes. Please check the latest commit.

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -39,6 +39,8 @@ static SimpleOidList catalog_table_include_oids = {NULL, NULL};
 static char *getMinOid(Archive *fout);
 static bool isBabelfishConfigTable(TableInfo *tbinfo);
 static void addFromClauseForLogicalDatabaseDump(PQExpBuffer buf, TableInfo *tbinfo, bool is_builtin_db);
+static void castSqlvariantToBasetype(PGresult* res, Archive* fout, PQExpBuffer q, int tuple, int field);
+static int getMbstrlen(const char* mbstr,Archive* fout);
```
rishabhtanwar29: Nit: * should be with variable name and not with the type.

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,213 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
 	destroyPQExpBuffer(q);
 }
+
+/*
+ * fixCursorForBbfSqlvariantTableData:
+ * Prepare custom cursor for all Babelfish tables with atleast one sql_variant
+ * datatype column to correctly dump sql_variant data.
+ */
+void
+fixCursorForBbfSqlvariantTableData(Archive *fout, TableInfo *tbinfo, PQExpBuffer buf, int *nfields, bool *is_sqlvariant_column, bool *has_sqlvariant_column)
```
rishabhtanwar29: Nit: We can wrap the long lines:
```
fixCursorForBbfSqlvariantTableData(Archive *fout,
                                   TableInfo *tbinfo,
                                   PQExpBuffer buf,
                                   int *nfields,
                                   bool *is_sqlvariant_column,
                                   bool *has_sqlvariant_column)
```
Additionally, we should explain these input parameters (except the known ones like `fout`, `tbinfo` etc) a little in the description above describing about what they are meant for.

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,213 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+
+/*
+ * fixBbfSqlvariantData:
+ * Modify the values in INSERT query that dump sql_variant column table data so
+ * that metadata is preserved during restore. If default value is not used or 
+ * the value is not NULL, we add a cast expression to typecast the value to its
+ * appropriate basetype.
+ */
+void
+fixBbfSqlvariantData(Archive *fout, PGresult* res, PQExpBuffer q, char* attgenerated, int tuple, int *orig_field, int *field)
```
rishabhtanwar29: * Nit: we can wrap the line here and some description about what `orig_field` and `field` refer to would be helpful.
* Nit++: * should be with variable name and not with the type.
* I think parameter `PQExpBuffer q` is not needed since we are resetting it below so it doesn't carry any useful information. If we really need a buffer then better to create it here rather than passing from the original function.
* Parameter name `tuple` looks confusing, I think it's row number so better name it like that.

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,213 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+ * Prepare custom cursor for all Babelfish tables with atleast one sql_variant
+ * datatype column to correctly dump sql_variant data.
+ */
+void
+fixCursorForBbfSqlvariantTableData(Archive *fout, TableInfo *tbinfo, PQExpBuffer buf, int *nfields, bool *is_sqlvariant_column, bool *has_sqlvariant_column)
+{
+	int nfields_new = 0;
+	*(has_sqlvariant_column) = hasSqlvariantColumn(tbinfo);
+	
+	if(!has_sqlvariant_column || !isBabelfishDatabase(fout))
```
surendravishnoi: we should put babelfish check first, even before calling hasSqlvariantColumn

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,213 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+	}
+	*(nfields) = nfields_new;
+	return;
+}
+
+/*
+ * fixBbfSqlvariantData:
+ * Modify the values in INSERT query that dump sql_variant column table data so
+ * that metadata is preserved during restore. If default value is not used or 
+ * the value is not NULL, we add a cast expression to typecast the value to its
```
rishabhtanwar29: Some corrections are needed:
1. Looks like this function just works on a single value at a time:
`s/Modify the values in INSERT query that dump sql_variant column table data so that metadata is preserved during restore/The function Modifies the value in INSERT command for sql_variant column so that metadata is preserved during restore/g`
`s/unless the column is generated or the value is NULL/If default value is not used or the value is not NULL/g`

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,213 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+	}
+	resetPQExpBuffer(q);
+	appendStringLiteralAH(q,
+				PQgetvalue(res, tuple, *field),
+								fout);
+	castSqlvariantToBasetype(res, fout, q, tuple, *field);
+
+	*orig_field = *(orig_field) + 1;
+	*field = *(field) + 2;
+	return;
```
rishabhtanwar29: Redundant

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,213 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+
+/*
+ * hasSqlvariantColumn:
+ * Returns true if any of the columns in table is a sqlvariant data type column
+ */
+
+bool
+hasSqlvariantColumn(TableInfo *tbinfo)
+{
+	for(int i=0; i<tbinfo->numatts; i++)
```
rishabhtanwar29: Nit: missing spaces before/after operators and variables.

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,213 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+}
+
+/*
+ * castSqlvariantToBasetype:
+ * Modify INSERT query in dump file by adding a CAST expression for a sql_variant
+ * column data entry in order to preserve the metadata of data type otherwise
+ * lost during restore.
+ */
+static void
+castSqlvariantToBasetype(PGresult* res, Archive* fout, PQExpBuffer q, int tuple, int field)
```
rishabhtanwar29: Same as above:
1. Better name from `tuple`
2. Remove redundant `PQExpBuffer q`.

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,213 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+	char* type = PQgetvalue(res, tuple, field+1);
+	int datalength = atoi(PQgetvalue(res, tuple, field+2));
+	int precision;
+	int scale;
+	int i;
+
+	archprintf(fout, "CAST(%s AS ", q->data);
+
+	/* data types defined in sys schema should be handled separately */
+	if(!pg_strcasecmp(type, "datetime") || !pg_strcasecmp(type, "datetimeoffset")
```
rishabhtanwar29: Nit: commenting here but applies to everywhere, please add a space between `if/else if` and `(`.

src/bin/pg_dump/dump_babel_utils.h
```diff
@@ -40,5 +40,8 @@ extern void setBabelfishDependenciesForLogicalDatabaseDump(Archive *fout);
 extern void dumpBabelGUCs(Archive *fout);
 extern void fixCursorForBbfCatalogTableData(Archive *fout, TableInfo *tbinfo, PQExpBuffer buf, int *nfields, char *attgenerated);
 extern void fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFrom);
+extern bool hasSqlvariantColumn(TableInfo *tbinfo);
+extern void fixBbfSqlvariantData(Archive *fout, PGresult* res, PQExpBuffer q, char* attgenerated, int tuple, int *orig_field, int *field);
+extern void fixCursorForBbfSqlvariantTableData(Archive *fout, TableInfo *tbinfo, PQExpBuffer buf, int *nfields, bool *is_sqlvariant_column, bool *has_sqlvariant_column);
```
rishabhtanwar29: Nit: wrap the lines.

src/bin/pg_dump/pg_dump.c
```diff
@@ -2156,6 +2159,7 @@ dumpTableData_insert(Archive *fout, const void *dcontext)
 	 * rather than the uninteresting-to-us value.
 	 */
 	attgenerated = (char *) pg_malloc(tbinfo->numatts * sizeof(char));
+	is_sqlvariant_column = (bool *) pg_malloc0(3 * tbinfo->numatts * sizeof(bool));
```
surendravishnoi: do this allocation inside fixCursorForBbfSqlvariantTableData, declare it as NULL here. That way the same variable can be used for both purpose (is_sqlvariant_column && is_sqlvariant_column[field]) and has_sqlvariant_column is no longer needed

src/bin/pg_dump/pg_dump.c
```diff
@@ -2181,6 +2185,9 @@ dumpTableData_insert(Archive *fout, const void *dcontext)
 		attgenerated[nfields] = tbinfo->attgenerated[i];
 		nfields++;
 	}
+
+	fixCursorForBbfSqlvariantTableData(fout, tbinfo, q, &nfields,
```
surendravishnoi: nfields could be return value to avoid passing by reference

src/bin/pg_dump/pg_dump.c
```diff
@@ -2283,8 +2296,15 @@ dumpTableData_insert(Archive *fout, const void *dcontext)
...
 			{
+				if(has_sqlvariant_column && is_sqlvariant_column[field])
+				{
+					fixBbfSqlvariantData(fout, res, q, attgenerated, tuple, &orig_field, &field);
+					continue;
+				}
+
 				if (field > 0)
 					archputs(", ", fout);
 				if (attgenerated[field])
```
surendravishnoi: how will this work? shouldn't it use orig_field?

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -39,6 +39,8 @@ static SimpleOidList catalog_table_include_oids = {NULL, NULL};
 static char *getMinOid(Archive *fout);
 static bool isBabelfishConfigTable(TableInfo *tbinfo);
 static void addFromClauseForLogicalDatabaseDump(PQExpBuffer buf, TableInfo *tbinfo, bool is_builtin_db);
+// static void castSqlvariantToBasetype(PGresult *res, Archive *fout, int row, int field);
```
surendravishnoi: Please remove if not required

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,179 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+
+	if (!isBabelfishDatabase(fout) || !hasSqlvariantColumn(tbinfo))
+		return 0;
+
+	*sqlvar_metdata_pos = (int *) pg_malloc0(3 * tbinfo->numatts * sizeof(int));
+	for (int i = 0; i < tbinfo->numatts; i++)
+	{
+		if (tbinfo->attisdropped[i])
+			continue;
+		if (tbinfo->attgenerated[i] && fout->dopt->column_inserts)
```
rishabhtanwar29: We shouldn't be checking for `fout->dopt->column_inserts` since for COPY it won't be set but for tables with sql_variant columns we will assume this to be true.

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,179 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+			* To find the basetype and bytelength of string data types we
+			* invoke sys.sql_variant_property and sys.datalength function on
+			* the sqlvariant column. These extra columns are added at the end of
+			* the select cursor query so that they do not interfere with
+			* expected dump behaviour.
+		*/
+		if (pg_strcasecmp(tbinfo->atttypnames[i],
+			quote_all_identifiers ? "\"sys\".\"sql_variant\"" : "sys.sql_variant") == 0)
+		{
+			appendPQExpBuffer(buf, ", sys.SQL_VARIANT_PROPERTY(%s, 'BaseType')", tbinfo->attnames[i]);
```
rishabhtanwar29: missing fmtid for tbinfo->attnames[i]

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,179 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+									TableInfo *tbinfo,
+									PQExpBuffer query,
+									int nfields,
+									int **sqlvar_metdata_pos)
+{
+	int orig_nfields = 0;
+	PQExpBuffer buf = createPQExpBuffer();
+
+	if (!isBabelfishDatabase(fout) || !hasSqlvariantColumn(tbinfo))
+		return 0;
```
rishabhtanwar29: Should return nfields.

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,179 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+ * fixCursorForBbfSqlvariantTableData:
+ * Prepare custom cursor for all Babelfish tables with atleast one sql_variant
+ * datatype column to correctly dump sql_variant data.
+ */
+int
+fixCursorForBbfSqlvariantTableData( Archive *fout,
+									TableInfo *tbinfo,
+									PQExpBuffer query,
+									int nfields,
+									int **sqlvar_metdata_pos)
```
surendravishnoi: metdata -> metadata

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,179 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+	*sqlvar_metdata_pos = (int *) pg_malloc0(3 * tbinfo->numatts * sizeof(int));
+	for (int i = 0; i < tbinfo->numatts; i++)
+	{
+		if (tbinfo->attisdropped[i])
+			continue;
+		if (tbinfo->attgenerated[i] && fout->dopt->column_inserts)
+			continue;
+
+		/* Skip TSQL ROWVERSION/TIMESTAMP column, it should be re-generated during restore. */
+		if (pg_strcasecmp(tbinfo->atttypnames[i],
```
surendravishnoi, line 1005: this is not needed, SQL variant check is sufficient

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,179 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+			quote_all_identifiers ? "\"sys\".\"sql_variant\"" : "sys.sql_variant") == 0)
+		{
+			appendPQExpBuffer(buf, ", sys.SQL_VARIANT_PROPERTY(%s, 'BaseType')", tbinfo->attnames[i]);
+			appendPQExpBuffer(buf, ", sys.datalength(%s)", fmtId(tbinfo->attnames[i]));
+			(*sqlvar_metdata_pos)[orig_nfields] = nfields;
+			nfields = nfields + 2;
+		}
+		orig_nfields++;
+	}
+	appendPQExpBuffer(query, buf->data);
```
rishabhtanwar29: Wrong, should be : appendPQExpBuffer(query, "%s", buf->data).
Need to destroy buf as well: destroyPQExpBuffer(buf);

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,179 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+	return nfields;
```
surendravishnoi, line 1030: cleanup local buffer

src/bin/pg_dump/pg_dump.c
```diff
@@ -2298,6 +2311,12 @@ dumpTableData_insert(Archive *fout, const void *dcontext)
 					continue;
 				}
 
+				if(sqlvar_metdata_pos && sqlvar_metdata_pos[field])
```
surendravishnoi: sqlvar_metdata_pos[field] -> sqlvar_metdata_pos[field] > 0

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -942,3 +944,179 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
+									int nfields,
+									int **sqlvar_metdata_pos)
+{
+	int orig_nfields = 0;
+	PQExpBuffer buf = createPQExpBuffer();
+
+	if (!isBabelfishDatabase(fout) || !hasSqlvariantColumn(tbinfo))
+		return 0;
+
+	*sqlvar_metdata_pos = (int *) pg_malloc0(3 * tbinfo->numatts * sizeof(int));
```
surendravishnoi: why are we allocating 3x, we just need one entry per attribute

src/bin/pg_dump/dump_babel_utils.c
aadityavermaa: If we remove this check then we will end up incrementing orig_nfields for rowversion/timestamp columns also, then we will end up storing the position at a different index.

surendravishnoi: makes sense

src/bin/pg_dump/dump_babel_utils.c
aadityavermaa: Used appendPQExpBufferStr instead.

src/bin/pg_dump/dump_babel_utils.c
aadityavermaa: Updated in latest commit. Thanks.

rishabhtanwar29 (approved):
LGTM

surendravishnoi (approved):

PR #132, opened 2023-04-21
support '[' and ']' wildcharacter for Like operater

forestkeeper:
Previously Babelfish didn't support '[' and ']' for like operator. This commit support this feature by changing the like execution. Add support to deal with '[' and ']' when it appears in pattern.

### Description

We have followed the document of : https://github.com/MicrosoftDocs/sql-docs/blob/live/docs/t-sql/language-elements/like-transact-sql.md to add the implement of '[' and ']' into like expression, like expression only works when it's tsql_dialect.
 
### Issues Resolved

Solved the issue customer complain about when they use '[' and ']' in like expression, they'll got un supported error msg

 
### Check List

- [ x ] Commits are signed per the DCO using --signoff 


src/backend/utils/adt/like_match.c
```diff
@@ -198,6 +204,66 @@ MatchText(const char *t, int tlen, const char *p, int plen,
 			NextByte(p, plen);
 			continue;
 		}
+		else if (*p == '[' && sql_dialect == SQL_DIALECT_TSQL)
```
macwoody, line 207: from high level, how can we reduce the foot print of this feature ? 
Can we convert TSQL pattern to PG compatible one? 
If yes, we may only need one hook from the entry point of like predicate.

src/backend/utils/adt/like_match.c
```diff
@@ -198,6 +204,66 @@ MatchText(const char *t, int tlen, const char *p, int plen,
...
+		{
+			/* Tsql deal with [ and ] wild character */
+			bool find_match, reverse_mode;
+			const char * p1, * prev;
+			int p1len;
+			p1 = p;
+			prev = NULL;
+			p1len = plen;
```
time-change: nit: can these be set at the declarations right above?

src/backend/utils/adt/like_match.c
```diff
@@ -198,6 +204,66 @@ MatchText(const char *t, int tlen, const char *p, int plen,
...
+			NextByte(p1, p1len);
+			reverse_mode = false;
```
time-change: same as above

src/backend/utils/adt/like_match.c
```diff
@@ -180,6 +180,12 @@ MatchText(const char *t, int tlen, const char *p, int plen,
 
 					if (matched != LIKE_FALSE)
 						return matched; /* TRUE or ABORT */
+				} else if (firstpat == '[' && sql_dialect == SQL_DIALECT_TSQL)
```
time-change, line 183: nit: brackets on newline

src/backend/utils/adt/like_match.c
```diff
@@ -198,6 +204,66 @@ MatchText(const char *t, int tlen, const char *p, int plen,
...
+			NextByte(p1, p1len);
+			reverse_mode = false;
+			while (p1len > 0 && *p1 != ']'){
+				NextByte(p1, p1len);
+			}
+			if (*p1 != ']')
+			{
+				if (GETCHAR(*p) != GETCHAR(*t))
+					return LIKE_FALSE;
+			}else{
```
JimFinnerty: preferred style is to use:
   }
   else
   {
       ...
   }

src/backend/utils/adt/like_support.c
```diff
@@ -1052,37 +1052,17 @@ like_fixed_prefix(Const *patt_const, bool case_insensitive, Oid collation,
...
-		}
-	}
 	
 	match = palloc(pattlen + 1);
 	match_pos = 0;
 	for (pos = 0; pos < pattlen; pos++)
 	{
+		/* [ and ] are special characters in LIKE for TSQL */
+		if ((patt[pos] == '[' ||
+			patt[pos] == ']') && sql_dialect == SQL_DIALECT_TSQL)
```
JimFinnerty, line 1062: does this do the correct thing when a ']' is encountered first?  Is the correct thing to treat ']' as a non-special character?

src/backend/utils/adt/like_match.c
JimFinnerty, line 207: Where are the test cases?

src/backend/utils/adt/like_support.c
forestkeeper: This is the optimize part, not the execution part, so any ] or [ will lead to full execution of like expression is the right thing in analyze / optimize part

src/backend/utils/adt/like_match.c
forestkeeper: the corresponding test cases is 
b like '%[abc]'

macwoody (approved):

src/backend/utils/adt/like_match.c
forestkeeper: We actually not, similar to operator is way too far to support '[' and ']' 
- it supports lots of additional regex match
- similar to itself has some problem in case insensitive / sensitive operator node transform to do before actually execution and it could not cover like expression to a variable or a real column value
- similar to is way slower than like ( since it's full regex matching ) 
- based on those fact, I decided to take this approach

So after I have a poc of similar to implement, I decided to manually write the rule into like execution

PR #130, opened 2023-04-17
Table Variable rework

lejaokri:
### Description
Change 1

    Add pg_depend, pg_shdepend to ENR for temp tables

Change 2

    (This depends on Change 1)

    Table Variables inside Transactions (engine code changes)

    This commit refactors how Table Variables work inside a Transaction
    especially when there is a rollback. The Table Variable metadata,
    on-disk files, and tuples should be able to survive rollback
    because Table Variables are non-transactional.

    Also, Table Variables primary key, constraint and toast tables should
    be non-transactional as well.

    On high-level there are three main components in this commit:

      1. Metadata:
         Table Variable metadata are added into ENR in order to bypass
         catalog MVCC.

      2. On-disk files:
         Table Variable relfilenode are not added to pending_deletes
         during table creation so it would not be deleted in case of rollback.

      3. Tuples:
         Introduced a new set of MVCC, Vacuum, Toast visibility functions.

 
### Issues Resolved

[List any issues this PR will resolve]
 
### Check List

- [ ] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is under the terms of the PostgreSQL license, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #134, opened 2023-04-26
Skip shared memory cache invalidation for ENR

time-change:
### Description

With #130 checkin - as ENR access uses system catalog cache, DDLs related to ENR must call cache invalidation APIs. However, as table variables have local scope, we do not need to invalidate shared memory for other backends which creates lock based contention. 

This change gives an improvement in performance in situations with many concurrent connections using temp tables or table variables. 

Local test results on a sample of 500 connections simultaneously creating and manipulating table variables showed a modest (~4%) performance increase. 
 
### Check List

### DIFF
--- a/src/backend/access/heap/heapam.c
+++ b/src/backend/access/heap/heapam.c
@@ -2195,7 +2195,7 @@ heap_insert(Relation relation, HeapTuple tup, CommandId cid,
 	 * the heaptup data structure is all in local memory, not in the shared
 	 * buffer.
 	 */
-	CacheInvalidateHeapTuple(relation, heaptup, NULL);
+	CacheInvalidateHeapTuple(relation, heaptup, NULL, false);
 
 	/* Note: speculative insertions are counted too, even if aborted later */
 	pgstat_count_heap_insert(relation, 1);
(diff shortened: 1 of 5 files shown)


src/backend/access/heap/heapam.c
```diff
@@ -2195,7 +2195,7 @@ heap_insert(Relation relation, HeapTuple tup, CommandId cid,
 	 * the heaptup data structure is all in local memory, not in the shared
 	 * buffer.
 	 */
-	CacheInvalidateHeapTuple(relation, heaptup, NULL);
+	CacheInvalidateHeapTuple(relation, heaptup, NULL, false);
```
surendravishnoi, line 2198: this change is specifically for perf improvement. We should have data in commit description clearly stating test case and improvement achieved via this change. We can have a multi connection test where all the connections run a procedure with table variables.

time-change: Updated original comment with some test results. I will format and publish all numbers from this experiment in a separate document, thought it was best to concisely summarize results in the commit description. 

As an additional note, JDBC tests pass after running.

time-change:
Decided to not merge after discussion, there was not a significant enough performance increase to justify

PR #135, opened 2023-04-28
Block cross version and cross migration mode dump/restore

aadityavermaa:
### Description

Babelfish currently only supports Database level dump and restore for databases with same migration mode ( i.e. either singledb->singledb or multidb->multidb). With this change the restore process with stop execution if the migration mode in the target cluster is different than the one from which the dump file was created. 
 
### Issues Resolved

BABEL-4055
 
### Check List

### DIFF
--- a/src/bin/pg_dump/dump_babel_utils.c
+++ b/src/bin/pg_dump/dump_babel_utils.c
@@ -144,9 +144,7 @@ void
 dumpBabelGUCs(Archive *fout)
 {
 	char		*oid;
-	char		*migration_mode;
 	PQExpBuffer	qry;
-	PGresult	*res;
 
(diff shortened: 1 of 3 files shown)


src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -166,6 +166,43 @@ dumpBabelGUCs(Archive *fout)
 	destroyPQExpBuffer(qry);
 }
 
+void
+dumpBabelMigrationModeCheck(Archive *fout)
```
surendravishnoi: will it be applicable to MVU as well? We are already blocking it CP level but still will this check be dumped during APG MVU

PR #146, opened 2023-06-02
like with bracket work improper for some corner cases

forestkeeper:
previously, like expression for bracket produce wrong result for :
1. [%, when there's no right bracket, % should not work as a wild character
2. the order range of character should not follow the ascii order but should follow the database natural order, for example : '='(ascii:62) should not between 0(ascii 48) to a(ascii 97)

Task: BABEL-4193'
Signed-off-by: Zhibai Song <szh@amazon.com>

### Check List

### DIFF
--- a/src/backend/utils/adt/like.c
+++ b/src/backend/utils/adt/like.c
@@ -25,50 +25,30 @@
 #include "parser/parser.h"
 #include "utils/builtins.h"
 #include "utils/pg_locale.h"
+#include "utils/varlena.h"
 
 
 #define LIKE_TRUE						1
 #define LIKE_FALSE						0
 #define LIKE_ABORT						(-1)
 
-static int varstr_cmp(const char *arg1, int len1, const char *arg2, int len2, pg_locale_t locale);
 
 static int	SB_MatchText(const char *t, int tlen, const char *p, int plen,
-						 pg_locale_t locale, bool locale_is_c);
+						 pg_locale_t locale, bool locale_is_c, Oid collation);
 static text *SB_do_like_escape(text *, text *);
 
 static int	MB_MatchText(const char *t, int tlen, const char *p, int plen,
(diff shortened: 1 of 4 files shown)


src/backend/utils/adt/like.c
```diff
@@ -161,6 +164,11 @@ GenericMatchText(const char *s, int slen, const char *p, int plen, Oid collation
 					 errmsg("nondeterministic collations are not supported for LIKE")));
 	}
 
+	if (sql_dialect == SQL_DIALECT_TSQL && set_like_collation_hook)
```
2jungkook, line 167: From my understanding, TSQL LIKE corresponds to PG ILIKE. In which case, do we need to call this (`GenericMatchText`) function?

src/backend/utils/adt/like.c
```diff
@@ -338,6 +351,11 @@ bytealike(PG_FUNCTION_ARGS)
 	p = VARDATA_ANY(pat);
 	plen = VARSIZE_ANY_EXHDR(pat);
 
+	if (sql_dialect == SQL_DIALECT_TSQL && set_like_collation_hook)
```
2jungkook, line 354: Do we have a test case for `bytealike`?

src/backend/utils/adt/like.c
```diff
@@ -25,6 +25,9 @@
 #include "parser/parser.h"
 #include "utils/builtins.h"
 #include "utils/pg_locale.h"
+#include "utils/varlena.h"
+#include "postgres_ext.h"
```
2jungkook, line 29: Could you explain why we need to include those two files?

src/backend/utils/adt/like.c
forestkeeper: Normally it won't 

Just in case

2jungkook (approved):

PR #151, opened 2023-06-19
Add Index to ENR when used on temp tables, prevent temp table index recreation issue 

time-change:
### Description

With this change, we add newly created indexes to ENR when used on temp tables, to match Table Variable behavior in these cases.

In addition, we add an extra scan in findDependentObjects to properly clean up index objects left in catalog when used with temp tables. Previously, entries would be left in pg_catalog, which would prevent recreating the same table with the same index name. This change also prevents rare cases of dangling catalog entries in cases where we aren't able to clean up properly, such as in a crash.
 
### Check List

### DIFF
--- a/src/backend/catalog/dependency.c
+++ b/src/backend/catalog/dependency.c
@@ -951,6 +951,7 @@ findDependentObjects(const ObjectAddress *object,
 		dependentObjects[numDependentObjects].subflags = subflags;
 		numDependentObjects++;
 	}
+
 	systable_endscan(scan);
 
 	/*
--- a/src/backend/catalog/dependency.c
+++ b/src/backend/catalog/dependency.c
@@ -76,6 +76,7 @@
 #include "commands/typecmds.h"
 #include "funcapi.h"
 #include "nodes/nodeFuncs.h"
+#include "parser/parser.h"
 #include "parser/parsetree.h"
 #include "rewrite/rewriteRemove.h"
 #include "storage/lmgr.h"
@@ -496,6 +497,7 @@ findDependentObjects(const ObjectAddress *object,
 	int			maxDependentObjects;
 	ObjectAddressStack mystack;
 	ObjectAddressExtra extra;
+	bool is_enr;
 
 	/*
 	 * If the target object is already being visited in an outer recursion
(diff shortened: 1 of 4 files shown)


src/backend/catalog/dependency.c
```diff
@@ -950,9 +952,85 @@ findDependentObjects(const ObjectAddress *object,
 		dependentObjects[numDependentObjects].subflags = subflags;
 		numDependentObjects++;
 	}
-
+	is_enr = scan->enr;
 	systable_endscan(scan);
 
+	/* Indices for temp tables will leave metadata in pg_catalog, so we should clean them up here. */
+	if (is_enr)
```
lejaokri: I am wondering why is this needed if code is supposed to add temp table index to ENR? If index is in ENR then we wouldnt need this, no?

src/backend/catalog/index.c
```diff
@@ -750,8 +750,10 @@ index_create(Relation heapRelation,
  		 * Index of Table Variables should also be in ENR and the index
 		 * can only be created as part of DECLARE @[name] TABLE;
 		 * when specifying PRIMARY KEYs and/or CONSTRAINTs.
+		 *
+		 * Indexes of temp tables should be in enr as well, to prevent cache issues.
 		 */
-		is_enr = (indexRelationName && strlen(indexRelationName) > 0 && indexRelationName[0] == '@');
+		is_enr = (indexRelationName && strlen(indexRelationName) > 0 && (indexRelationName[0] == '@' || indexRelationName[0] == '#'));
```
lejaokri: In `heap_create_with_catalog()` you have a check there so not all temp tables are added in ENR but here you are doing all "index" of temp tables should be in enr. What happens in the case where the temp table is NOT in enr but index is in enr?

src/backend/catalog/dependency.c
time-change: This is needed because the below code does add the pg_index entries to ENR, but doesn't prevent them from being created in catalogs. It's those entries in pg_class, pg_depend etc that will not be cleaned up in findDependentObjects.

src/backend/catalog/index.c
time-change: In that case, this line might be unnecessary.

surendravishnoi: we need to follow all in ENR or nothing. We should put indexes in ENR if corresponding tables are in ENR. We should have same check as that for tables.

src/backend/catalog/dependency.c
surendravishnoi: Can we add indexes to ENR only? If we are keeping relations in ENR only, why not indexes?

src/backend/catalog/index.c
time-change: agreed, reworked to handle indexes in ENR using same check from heap_create_with_catalog.

src/backend/catalog/index.c
```diff
@@ -750,8 +751,16 @@ index_create(Relation heapRelation,
...
+		 *
+		 * Indexes of temp tables should be in enr as well, to prevent cache issues.
+		 * Index names can actually contain '#' naturally, but because of construct_unique_index_name
+		 * we have no way of checking that without the original index name provided to parser.
 		 */
-		is_enr = (indexRelationName && strlen(indexRelationName) > 0 && indexRelationName[0] == '@');
+		is_enr = (indexRelationName && strlen(indexRelationName) > 0 && (indexRelationName[0] == '@' || strchr(indexRelationName, '#') != NULL));
+
+		/* One last check, if the table has any dependencies */
+		if(CheckTempTableHasDependencies(heapRelation->rd_att))
```
lejaokri: I can think of two issues:

1. This isnt completely consistent with the check in heap_create_with_catalog(). This check here doesnt ignore Table Variables but the one in heap_create_with_catalog() ignores Table Variables. The reason is that we want everything associated to TVs in ENR.

2. Consider the following test
-> create table where CheckTempTableHasDependencies returns false this means table is in ENR
-> Alter table to include a column where CheckTempTableHasDependencies would return True
-> Create an index, now CheckTempTableHasDependencies would return True so you will end up with table in ENR, and index not in ENR.

time-change: for case 2), i don't know if there's a simple solution for that situation. i think the problem then is more with the alter table, and not the index creation, right? in theory, we would have to do something like move ENR metadata out in that scenario.

I think that fix would be out of scope for this issue.

lejaokri: Perhaps `get_ENR_withoid()` can help

src/backend/catalog/index.c
```diff
@@ -747,11 +748,14 @@ index_create(Relation heapRelation,
...
-		 * when specifying PRIMARY KEYs and/or CONSTRAINTs.
+		 * Index of temp objects in ENR should also be in ENR.
+		 * For table variables, index can only be created as
+		 * part of DECLARE @[name] TABLE; when specifying
+		 * PRIMARY KEYs and/or CONSTRAINTs.
 		 */
-		is_enr = (indexRelationName && strlen(indexRelationName) > 0 && indexRelationName[0] == '@');
+		is_enr = (indexRelationName && strlen(indexRelationName) > 0 &&
+			(indexRelationName[0] == '@' || (strchr(indexRelationName, '#') != NULL
+			&& get_ENR_withoid(currentQueryEnv, heapRelationId, ENR_TSQL_TEMP))));
```
lejaokri: nitpick for performance purposes we do not need to call `get_ENR_withoid()` if it is a table variable.

time-change: I don't think it's being called if it's a TV, there's parens around strchr+get_ENR_withoid

lejaokri: okay got it thanks thats a bit painful to read :)

lejaokri: Are we also allowed to create index for temp tables where the index name does not start with '#'?

time-change: The strchr call is unnecessary, yeah

lejaokri (approved):

priyansx (approved):
LGTM

src/backend/catalog/dependency.c
```diff
@@ -76,6 +76,7 @@
 #include "commands/typecmds.h"
 #include "funcapi.h"
 #include "nodes/nodeFuncs.h"
+#include "parser/parser.h"
```
surendravishnoi, line 79: why this change?

PR #160, opened 2023-07-19
Casting a DATE literal without time part to TIME datatype fails

Suthapalli-Ramya-satya-vasavi-srija:
### Description

If different kind of inputs are given to `time` datatype like `<DATE>T<TIME>`, a date literal string to a TIME datatype when the time component in the string is not present and inputs are of types `YYYYMMDD`, `YYMMDD` and if month is mentioned in text format `YYYY MON DD` then the PG doesn't allows casting and throw errors.

This commit fixes the following issues

-  Considers format `<DATE>T<TIME>[Z]` as valid and return expected result.
-  Considers formats `YYYY`, `YYYYMMDD` and `YYMMDD` as valid and return expected result.
-  If month is mentioned in text format considered date formats `YYYY MON DD`, `YYYY DD MON`, `MON DD YYYY`, `MON YYYY DD`, `DD MON YYYY`, `DD YYYY MON`.
- Accepts Empty String('') to time datatype.
- Accepts only DATE literal without time part to time datatype. Accepted formats are `DD-MON-YYYY`, `YYYY-MON-DD`, `MM{-/.}DD{-/.}YYYY`, `YYYY{-/.}MM{-/.}DD`.
- If timezone is specified then validates the timezone whether correctly mentioned or not i.e., timezone should be of format `[+-]\\d{1,2}:\\d{1,2}`.
- If `[AP]M` specified then made sure that it is present after the time in the input.

 
### Issues Resolved

BABEL-1541, BABEL-1528, BABEL-1539
 
### Check List


src/backend/utils/adt/datetime.c
```diff
@@ -26,6 +26,7 @@
 #include "funcapi.h"
 #include "miscadmin.h"
 #include "nodes/nodeFuncs.h"
+#include "parser/parser.h"
```
surendravishnoi: this is not the ideal approach. We cannot modify PG behaviour for any PG data types. Let us start with a document to describe the data type -> behavior differences and relevant functions which handle the logic. We have to provide  hooks at the right places to override PG behaviour with Babel behavior conditionally. Depending on the number of changes required, we can decide the granularity of hooks.

src/backend/utils/adt/datetime.c
```diff
@@ -799,7 +800,7 @@ ParseDateTime(const char *timestr, char *workbuf, size_t buflen,
 				ftype[nf] = DTK_TIME;
 				APPEND_CHAR(bufp, bufend, *cp++);
 				while (isdigit((unsigned char) *cp) ||
-					   (*cp == ':') || (*cp == '.'))
```
surendravishnoi: let us not change unnecessary lines

Suthapalli-Ramya-satya-vasavi-srija: Will fix in the next revision.

src/backend/utils/adt/datetime.c
```diff
@@ -1972,6 +1973,18 @@ DecodeTimeOnly(char **field, int *ftype, int nf,
 					if (dterr)
 						return dterr;
 				}
+				/*
```
surendravishnoi: let us add a hook at line 1966, no other changes required in this file (except for adding a wrapper)

if (tsql_date_decoder && tsql_date_decoder(field[i], fmaks, tmask, is2digits, tm, &dterr)
{
     return dterr;
}

tsql_date_decoder(params)
{
   if (sql_dialect == SQL_DIALECT_TSQL && nf == 1 && ftype[nf - 1] == DTK_DATE)
       decodeDate(); // Add a wrapper around decodeDate as it is static
        validateDate(); // Call directly, not static
}

src/backend/utils/adt/date.c
```diff
@@ -1387,9 +1390,17 @@ time_in(PG_FUNCTION_ARGS)
...
 						  field, ftype, MAXDATEFIELDS, &nf);
 	if (dterr == 0)
+	{
+		if(modify_field_hook && nf >= 3)
+			(*modify_field_hook)(&nf, field, ftype);
 		dterr = DecodeTimeOnly(field, ftype, nf, &dtype, tm, &fsec, &tz);
+	}
 	if (dterr != 0)
+	{
+		if (time_hook)
```
surendravishnoi: why are we adding a hook to give different error? What is the problem if we give PG error?

Suthapalli-Ramya-satya-vasavi-srija: Its just we are making the behaviour similar to SQL Server, for different kind of errors the SQL Server throws same error.
Note: Another benefit from this is the error message will not show the modified input(modifications which were given to the input) so that the customer will not have problems with understanding why the error message and input has different strings.

surendravishnoi: ok, we an keep this hook

src/backend/utils/adt/datetime.c
```diff
@@ -83,6 +84,8 @@ const char *const months[] = {"Jan", "Feb", "Mar", "Apr", "May", "Jun",
 const char *const days[] = {"Sunday", "Monday", "Tuesday", "Wednesday",
 "Thursday", "Friday", "Saturday", NULL};
 
+handle_textMonth_case_hook_type handle_textMonth_case_hook = NULL;
```
surendravishnoi: handle_text_month_hook_type

src/backend/utils/adt/datetime.c
```diff
@@ -2524,6 +2558,16 @@ DecodeTimeOnly(char **field, int *ftype, int nf,
 	return 0;
 }
 
+/*
```
surendravishnoi: Please add at the end of file

src/backend/utils/adt/datetime.c
```diff
@@ -2263,11 +2277,26 @@ DecodeTimeOnly(char **field, int *ftype, int nf,
 					/* otherwise it is a single date/time field... */
 					else
 					{
-						dterr = DecodeNumber(flen, field[i],
-											 false,
-											 (fmask | DTK_DATE_M),
-											 &tmask, tm,
-											 fsec, &is2digits);
+						if(sql_dialect == SQL_DIALECT_TSQL)
```
surendravishnoi: if(sql_dialect == SQL_DIALECT_TSQL && haveTextMonth)

src/backend/utils/adt/datetime.c
```diff
@@ -1963,6 +1967,16 @@ DecodeTimeOnly(char **field, int *ftype, int nf,
 				if (tzp == NULL)
 					return DTERR_BAD_FORMAT;
 
+				/*
+				 * For date format like 'yyyy-mm-dd', we should return time 00:00:00.0000000
+				 * instead of marking it as invalid time format
+				 */
```
RIC06X: I think you might need to add another format check here since if the cast string comes from a variable, it will pass the antlr check and only reveal its value during execution.

src/backend/utils/adt/date.c
```diff
@@ -1384,6 +1386,9 @@ time_in(PG_FUNCTION_ARGS)
 	int			dtype;
 	int			ftype[MAXDATEFIELDS];
 
+	if (time_in_hook && sql_dialect == SQL_DIALECT_TSQL)
```
surendravishnoi: let us call it tsql_time_in_hook

src/backend/utils/adt/date.c
```diff
@@ -1384,6 +1386,9 @@ time_in(PG_FUNCTION_ARGS)
 	int			dtype;
 	int			ftype[MAXDATEFIELDS];
 
+	if (tsql_time_in_hook && sql_dialect == SQL_DIALECT_TSQL)
```
surendravishnoi: we can move dialect check inside the hook itself and pass result as out parameter

src/backend/utils/adt/date.c
```diff
@@ -28,6 +28,7 @@
 #include "miscadmin.h"
 #include "nodes/supportnodes.h"
 #include "parser/scansup.h"
+#include "parser/parser.h"
```
surendravishnoi: this should not be required once we move dialect check into hook

surendravishnoi (approved):

rishabhtanwar29 (approved):

PR #168, opened 2023-08-02
Sort ENR before deletion to avoid dependency issues

time-change:
### Description

Sort ENR, similar to findDependentObjects, before deletion. This avoids situations where ENR objects are cleaned up in the wrong order, which will throw an error in some cases. For example, in this case: 

```
1> create procedure p
2> as
3> begin
4>      create table #t (id int identity(1,1))
5>      --drop table #t  -- when uncommenting this line, no error is raised
6> end
7> go
1>
2> exec p
3> go
Msg 3723, Level 16, State 1, Server BABELFISH, Line 1
cannot drop sequence "#t_id_seq" because column id of table "#t" requires it
```

An error is thrown due to ENRDropTempTables attempting to drop the sequence #t_id_seq before dropping table #t. In findDependentObjects, dependent objects are sorted into a stable visitation order, so we should do the same in ENRDropTempTables. 
 
### Issues Resolved

BABEL-3168 - Implicit drop in procedure of #tmp table with IDENTITY column raises error
 
### Check List

### DIFF
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -955,40 +955,37 @@ ENRDropTempTables(QueryEnvironment *queryEnv)
 {
 	ListCell   *lc = NULL;
 	ObjectAddress object;
+	ObjectAddresses *objects;
 
 	if (!queryEnv)
 		return;
 
-	object.classId = RelationRelationId;
-	object.objectSubId = 0;
+	objects = new_object_addresses();
 
 	/*
 	 * Loop through the registered ENRs to drop temp tables.
 	 */
-	while (lc || (lc = list_head(queryEnv->namedRelList)))
+	foreach(lc, queryEnv->namedRelList)
 	{
-		EphemeralNamedRelation enr;
+		EphemeralNamedRelation enr = (EphemeralNamedRelation) lfirst(lc);
 
-		enr = (EphemeralNamedRelation) lfirst(lc);
+		if (enr->md.enrtype != ENR_TSQL_TEMP)
+			continue;
 
-		if (enr->md.enrtype != ENR_TSQL_TEMP) {
-			lc = lnext(queryEnv->namedRelList, lc);
-			if (!lc)
-				break;
-			else
-				continue;
-		}
-
-		/*
-		 * performDeletion() will remove the table AND the ENR entry, so no
-		 * need to remove the entry afterwards.
-		 */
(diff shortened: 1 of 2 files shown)


src/backend/utils/misc/queryenvironment.c
```diff
@@ -906,40 +906,37 @@ ENRDropTempTables(QueryEnvironment *queryEnv)
...
-		lc = NULL;
+		add_exact_object_address(&object, objects);
 	}
+
+	/*
+	 * performMultipleDeletions() will remove the table AND the ENR entry,
+	 * so no need to remove the entry afterwards. Sort addresses beforehand
+	 * so that we don't run into dependency issues when dropping objects.
+	 */
+	sort_object_addresses(objects);
```
lejaokri: Did we try this with temp table with identity and primary key?
I am curious how this will work if index is also in enr.

when we do something like `create table #t (id int identity primary key)` then the flow is

create sequence
create table
create index

so after reverse sort, we try to drop index first. Can we quickly check this works?

Thanks

time-change: Thanks for catching that, it actually does run into issues trying to drop the index first. Without the change though we are still running into the original issue with this example as well.

```
1> create procedure pz as
2> begin
3> create table #t (id int identity primary key)
4> create index i on #t(id)
5> end
6> go
1> exec pz
2> go
Msg 3723, Level 16, State 1, Server BABELFISH, Line 1
cannot drop index "#t_pkey" because constraint #t_pkey on table "#t" requires it
```

surendravishnoi: what is the final conclusion here?

time-change: I would say that, the current change is preferable to the old behavior because using just sequence in TV is far more prevalent than this edge case of creating index on temp table with primary key.

lejaokri: Is it possible to come up with a solution that covers both?

time-change: Interestingly enough, removing the `sort_object_addresses` call seems to have fixed both issues. Seems like performMultipleDeletions correctly gets the order to delete objects in. See tests in https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1708

lejaokri (approved):

xhfanhe (approved):

PR #170, opened 2023-08-03
Add pg_sequence into ENR for BBF table variables and Temp tables

lejaokri:
Table Variables and Temp Tables can have identity columns which adds entries to pg_sequence. If the table is in ENR, then pg_sequence entry should also be in ENR otherwise we would have catalog inconsistency.

Task: BABEL-4267

### Description

[Describe what this change achieves]
 
### Issues Resolved

### Check List

### DIFF
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -35,6 +35,7 @@
 #include "catalog/pg_statistic_ext.h"
 #include "catalog/pg_type.h"
 #include "catalog/pg_depend.h"
+#include "catalog/pg_sequence.h"
 #include "catalog/pg_shdepend.h"
 #include "catalog/pg_index_d.h"
 #include "parser/parser.h"      /* only needed for GUC variables */
@@ -169,6 +170,21 @@ List *get_namedRelList()
 	return currentQueryEnv->namedRelList;
 }
 
+bool has_existing_enr_relations()
+{
+	QueryEnvironment *queryEnv = currentQueryEnv;
+
+	while (queryEnv)
+	{
+		if (queryEnv->namedRelList != NIL)
+			return true;
+
+		queryEnv = queryEnv->parentEnv;
+	}
+
+	return false;
+}
+
 /*
  * This returns an ENR if there is a name match in the given collection.  It
  * must quietly return NULL if no match is found.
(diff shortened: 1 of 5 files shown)


src/backend/catalog/dependency.c
```diff
@@ -1427,7 +1427,7 @@ doDeletion(const ObjectAddress *object, int flags)
 
 				/*
 				 * for a sequence, in addition to dropping the heap, also
-				 * delete pg_sequence tuple
+				 * delete pg_sequence tuple.
```
surendravishnoi: not required, let us avoid unnecessary engine changes

src/backend/utils/misc/queryenvironment.c
```diff
@@ -261,7 +264,14 @@ bool ENRgetSystableScan(Relation rel, Oid indexId, int nkeys, ScanKey key, List
 	Datum v1 = 0, v2 = 0, v3 = 0, v4 = 0;
 
 	if (sql_dialect != SQL_DIALECT_TSQL)
-		return false;
+	{
+		/*
+		 * It is possible we are inside a non tsql proc/function
+		 * where sql_dialect is temporarily set to PG.
+		 */
+		if (!in_non_tsql_proc_hook || !in_non_tsql_proc_hook())
```
surendravishnoi: what do we need this change? If dialect is not TSQL, this hook indicates that there is no PG function active in call stack, how does it influence ENR scan?

lejaokri: Hi Surendra, good question.

There are cases when we overwrite the tsql_dialect. In the case for identity, we call nextval_internal from `fmgr_security_definer` where tsql_dialect gets overriden temporarily. When this happens, we execute `nextval_internal` in non tsql dialect and we wont be able to find the sequence entry in ENR.

surendravishnoi: This is not ideal. We are adding objects in ENR which relates to temp table/variables only. Even we skip adding temp table with UDT to ENR to avoid UDT dependency. Given sequence is internal to temp object only. Is it possible to avoid this dependency on PG function call and rather call the function (nextval_internal) directly?

xhfanhe (approved):

src/backend/utils/misc/queryenvironment.c
```diff
@@ -255,15 +271,31 @@ ENRMetadataGetTupDesc(EphemeralNamedRelationMetadata enrmd)
...
-	Oid reloid;
 	bool found = false;
 	int index = 0;
 	Datum v1 = 0, v2 = 0, v3 = 0, v4 = 0;
+	Oid pltsql_lang_oid = InvalidOid;
+	Oid pltsql_validator_oid = InvalidOid;
 
-	if (sql_dialect != SQL_DIALECT_TSQL)
-		return false;
+	Oid reloid = RelationGetRelid(rel);
```
surendravishnoi, line 280: we should have pltsql_lang_oid check first, followed by reloid check. That way we avoid RelationGetRelid call for APG without BBF.

lejaokri: actually `RelationGetRelid` is cheaper and we avoid calling `get_func_language_oids_hook()` for relations other than `SequenceRelationId`

```
#define RelationGetRelid(relation) ((relation)->rd_id)
```

surendravishnoi (approved):

PR #183, opened 2023-08-11
Handle different behaviour with like escape null and like escape ''

vikashprajapati-cell:
### Description

When query `select 1 where 'ABCD' LIKE 'AB[C]D' ESCAPE'' ` executes on babelfish, it should return an error as expected behaviour. But it was not throwing error, this is actually there was no condition added to return error in case of empty ESCAPE char. So i added a condition to return error for this case.
Another query is `select 1 where 'ABCD' LIKE 'AB[C]D' ESCAPE null`. This should return 1 row as expected behaviour and before this fix, this query returned 0 row on execution on babelfish. So i added one condition to capture this case and revert the behaviour as expected. (Behaviour same as no ESCAPE char used - as per Rob's comment on Jira).

Task: BABEL-4271
Authored-by: Vikash Prajapati [vikasprj@amazon.com](mailto:vikasprj@amazon.com)
Signed-off-by: Vikash Prajapati [vikasprj@amazon.com](mailto:vikasprj@amazon.com)

### Issues Resolved

BABEL-4271
 
### Check List

### DIFF
--- a/src/backend/parser/parse_expr.c
+++ b/src/backend/parser/parse_expr.c
@@ -919,7 +919,7 @@ exprIsNullConstant(Node *arg)
  * underlying function that returns an int type so the index is used.
  */
 static void
-rewrite_scope_identity_call(ParseState *pstate, Node **lexpr, Node **rexpr)
+bbf_rewrite_function_call(ParseState *pstate, Node **lexpr, Node **rexpr)
 {
 	Var       *col_expr;
 	FuncExpr  *func_expr;
@@ -931,22 +931,26 @@ rewrite_scope_identity_call(ParseState *pstate, Node **lexpr, Node **rexpr)
 	if (!(*lexpr) || !(*rexpr))
 		return;
 
-	if (IsA(*lexpr, Var) &&
-		IsA(*rexpr, FuncExpr) &&
-		strstr(get_func_name(((FuncExpr*) (*rexpr))->funcid), "like_escape") != NULL)
-    {
-        func_expr = (FuncExpr *) (*rexpr);
-        if((func_expr->args)->length==2){
-            Node *node = lsecond(func_expr->args);
-            if (IsA(node,Const) && ((Const *)(node))->constisnull)
-            {
-                *rexpr = (Node *)((func_expr)->args->elements[0]).ptr_value;
-                return;
-            }
-        }
-		return;
-    }
-	else if ((IsA(*lexpr, Var) && IsA(*rexpr, FuncExpr)))
+	func_expr = (FuncExpr *) (*rexpr);
+	if (IsA(*rexpr, FuncExpr) &&
+		strstr(get_func_name(func_expr->funcid), "like_escape") != NULL)
+	{
+		if((func_expr->args)->length==2)
+		{
+			Node *node = lsecond(func_expr->args);
+			if (IsA(node,Const) && ((Const *)(node))->constisnull)
+			{
+				/*
+				 * This condition handles like escape null query,
+				 * only when using TSQL.
+				 */
+				*rexpr = (Node *)((func_expr)->args->elements[0]).ptr_value;
+				return;
+			}
+		}
+	}
+
+	if ((IsA(*lexpr, Var) && IsA(*rexpr, FuncExpr)))
 	{
 		col_expr = (Var*) *lexpr;
 		func_expr = (FuncExpr*) *rexpr;
@@ -956,23 +960,6 @@ rewrite_scope_identity_call(ParseState *pstate, Node **lexpr, Node **rexpr)
 		col_expr = (Var*) *rexpr;
 		func_expr = (FuncExpr*) *lexpr;
 	}
-	else if (IsA(*rexpr, FuncExpr) &&
-            strstr(get_func_name(((FuncExpr*) (*rexpr))->funcid), "like_escape") != NULL)
-    {
-        func_expr = (FuncExpr *) (*rexpr);
-        if((func_expr->args)->length==2){
-            Node *node = lsecond(func_expr->args);
-            if (IsA(node,Const) && ((Const *)(node))->constisnull)
-            {
-                /*
-                * This condition deals with ESCAPE null, means no ESCAPE char used
-                */
-                *rexpr = (Node *)((func_expr)->args->elements[0]).ptr_value;
-                return;
-            }
-        }
-		return;
-    }
 	else
 		return;
 
@@ -1069,7 +1056,7 @@ transformAExprOp(ParseState *pstate, A_Expr *a)
 		rexpr = transformExprRecurse(pstate, rexpr);
 
 		/* So that where clauses with scope_identity use an index */
-		rewrite_scope_identity_call(pstate, &lexpr, &rexpr);
+		bbf_rewrite_function_call(pstate, &lexpr, &rexpr);
 
 		result = (Node *) make_op(pstate,
 								  a->name,
--- a/src/backend/parser/parse_expr.c
+++ b/src/backend/parser/parse_expr.c
@@ -931,7 +931,22 @@ rewrite_scope_identity_call(ParseState *pstate, Node **lexpr, Node **rexpr)
 	if (!(*lexpr) || !(*rexpr))
 		return;
 
-	if ((IsA(*lexpr, Var) && IsA(*rexpr, FuncExpr)))
+	if (IsA(*lexpr, Var) &&
+		IsA(*rexpr, FuncExpr) &&
+		strstr(get_func_name(((FuncExpr*) (*rexpr))->funcid), "like_escape") != NULL)
+    {
+        func_expr = (FuncExpr *) (*rexpr);
+        if((func_expr->args)->length==2){
+            Node *node = lsecond(func_expr->args);
+            if (IsA(node,Const) && ((Const *)(node))->constisnull)
+            {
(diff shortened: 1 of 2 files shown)


src/backend/parser/parse_expr.c
```diff
@@ -941,6 +956,23 @@ rewrite_scope_identity_call(ParseState *pstate, Node **lexpr, Node **rexpr)
 		col_expr = (Var*) *rexpr;
 		func_expr = (FuncExpr*) *lexpr;
 	}
+	else if (IsA(*rexpr, FuncExpr) &&
+            strstr(get_func_name(((FuncExpr*) (*rexpr))->funcid), "like_escape") != NULL)
```
ahmed-shameem: Why are we not checking any condition of lexpr here like above case? If this case is sufficient to prove that this the current query has LIKE clause then why do we need the above block at all? We can just check the rexpr and see whether like_escape function is called or not. If so, then we already have sufficient condition to make sure whether the query contains LIKE ESCAPE or not.

ahmed-shameem (changes requested):

src/backend/parser/parse_expr.c
```diff
@@ -931,7 +931,22 @@ rewrite_scope_identity_call(ParseState *pstate, Node **lexpr, Node **rexpr)
 	if (!(*lexpr) || !(*rexpr))
 		return;
 
-	if ((IsA(*lexpr, Var) && IsA(*rexpr, FuncExpr)))
+	if (IsA(*lexpr, Var) &&
```
forestkeeper: Why don't you combine this if clause together with another else clause, I think we really not care about what is the lexpr in here is ?

src/backend/utils/adt/like_match.c
```diff
@@ -350,15 +350,26 @@ do_like_escape(text *pat, text *esc)
...
+		if (sql_dialect == SQL_DIALECT_TSQL){
+            /*
+            * Escape string is empty, just show the error.
+            */
+            ereport(ERROR,
+                    (errcode(ERRCODE_INVALID_ESCAPE_SEQUENCE),
+                    errmsg("The invalid escape character \"\" was specified in a LIKE predicate."),
+                    errhint("Escape string must null or one character.")));
+        }
+		else{
```
forestkeeper: Can we not modify the original code logic with a else ?
just add the sql_dialect is enough

src/backend/utils/adt/like_match.c
```diff
@@ -350,15 +350,26 @@ do_like_escape(text *pat, text *esc)
...
-		/*
-		 * No escape character is wanted.  Double any backslashes in the
-		 * pattern to make them act like ordinary characters.
-		 */
-		while (plen > 0)
-		{
-			if (*p == '\\')
-				*r++ = '\\';
-			CopyAdvChar(r, p, plen);
+		if (sql_dialect == SQL_DIALECT_TSQL){
```
forestkeeper: pls use 
```
if ()
{
}
```
format, not using : 
```
if(){
}
```
format

src/backend/utils/adt/like_match.c
```diff
@@ -350,15 +350,26 @@ do_like_escape(text *pat, text *esc)
...
+            /*
+            * Escape string is empty, just show the error.
```
ahmed-shameem, line 355: Also, follow the format we use for comments. It should be:

/*
 * Escape string is empty, just show the error.
 * /

src/backend/utils/adt/like_match.c
```diff
@@ -350,15 +350,26 @@ do_like_escape(text *pat, text *esc)
 
 	if (elen == 0)
 	{
-		/*
-		 * No escape character is wanted.  Double any backslashes in the
```
ahmed-shameem: Also, please follow the format used here for comments in your changes.

src/backend/parser/parse_expr.c
RIC06X: I'm wondering if it is appropriate to write this escape null and like escape '' code in this rewrite_scope_identity_call function. The function's purpose is not to handle the escape null/"" case. I would suggest to create a new function and put all your escape related code there.

src/backend/parser/parse_expr.c
```diff
@@ -931,7 +931,22 @@ rewrite_scope_identity_call(ParseState *pstate, Node **lexpr, Node **rexpr)
...
+		IsA(*rexpr, FuncExpr) &&
+		strstr(get_func_name(((FuncExpr*) (*rexpr))->funcid), "like_escape") != NULL)
+    {
+        func_expr = (FuncExpr *) (*rexpr);
+        if((func_expr->args)->length==2){
+            Node *node = lsecond(func_expr->args);
+            if (IsA(node,Const) && ((Const *)(node))->constisnull)
```
RIC06X, line 941: if it is an empty string, "", this value ((Const *)(node))->constisnull will still be false, I'm not sure if how do you check empty string.

src/backend/utils/adt/like_match.c
RIC06X: if the escape string is an empty string, the elen will be 0. if the escape string is null, it will not go through the do_like_escape function. So I think you only need this sql_dialect check to solve this issue.

src/backend/parser/parse_expr.c
vikashprajapati-cell: We dont care about lexpr in here to check for like escape.

src/backend/parser/parse_expr.c
vikashprajapati-cell: I have renamed the function as discussed offline with Zhibai.

vikashprajapati-cell: > Why don't you combine this if clause together with another else clause, I think we really not care about what is the lexpr in here is ?

I have updated this in the PR revision.

src/backend/utils/adt/like_match.c
vikashprajapati-cell: Done

src/backend/utils/adt/like_match.c
vikashprajapati-cell: Updated

src/backend/utils/adt/like_match.c
vikashprajapati-cell: Done.

forestkeeper (approved):

src/backend/parser/parse_expr.c
```diff
@@ -931,6 +931,25 @@ rewrite_scope_identity_call(ParseState *pstate, Node **lexpr, Node **rexpr)
 	if (!(*lexpr) || !(*rexpr))
 		return;
 
+	func_expr = (FuncExpr *) (*rexpr);
+	if (IsA(*rexpr, FuncExpr) &&
+		strstr(get_func_name(func_expr->funcid), "like_escape") != NULL)
```
Deepesh125, line 936: why strstr?

src/backend/parser/parse_expr.c
```diff
@@ -931,6 +931,25 @@ rewrite_scope_identity_call(ParseState *pstate, Node **lexpr, Node **rexpr)
...
+	{
+		if((func_expr->args)->length==2)
```
Deepesh125, line 938: should use list_length()

src/backend/parser/parse_expr.c
```diff
@@ -931,6 +931,25 @@ rewrite_scope_identity_call(ParseState *pstate, Node **lexpr, Node **rexpr)
...
+		{
+			Node *node = lsecond(func_expr->args);
+			if (IsA(node,Const) && ((Const *)(node))->constisnull)
+			{
+				/*
+				 * This condition handles like escape null query,
+				 * only when using TSQL.
+				 */
+				*rexpr = (Node *)((func_expr)->args->elements[0]).ptr_value;
```
Deepesh125, line 947: this is not right way to access list element directly. Use appropriate API form list.h

src/backend/parser/parse_expr.c
```diff
@@ -919,7 +919,7 @@ exprIsNullConstant(Node *arg)
  * underlying function that returns an int type so the index is used.
  */
 static void
-rewrite_scope_identity_call(ParseState *pstate, Node **lexpr, Node **rexpr)
+bbf_rewrite_function_call(ParseState *pstate, Node **lexpr, Node **rexpr)
```
Deepesh125, line 922: overall we should use hook for such changes

src/backend/utils/adt/like_match.c
```diff
@@ -348,6 +348,17 @@ do_like_escape(text *pat, text *esc)
 	result = (text *) palloc(plen * 2 + VARHDRSZ);
 	r = VARDATA(result);
 
+	if (elen==0 && sql_dialect == SQL_DIALECT_TSQL)
```
Deepesh125, line 351: should hook be used here?

vikashprajapati-cell: I think the behaviour of query `select 1 where 'ABCD' LIKE 'AB[C]D' ESCAPE ''` is already implemented in `like_match.c` file in engine. And, i think it doesn't require extension to provide logic, so we don't need a hook in this case. This condition should suffice to handle the scenerio.

src/backend/parser/parse_expr.c
vikashprajapati-cell: I found strcmp() makes more sense while comparing the string, i will update this in next PR.

PR #193, opened 2023-08-24
Fixed handling of input arguments for ISNULL()

Deepesh125:
### Description

Previously, we were handling ISNULL using COALESCE semantics which is not right thing to do because return datatype of T-SQL ISNULL function is not same as COALESCE in all the cases. So this commit fixes this issue by appropriately calling hook to select the common type or say return type for ISNULL. Hook is implemented in babelfishpg_tsql function via PR: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1709

BABEL-917

### Check List

### DIFF
--- a/src/backend/parser/parse_expr.c
+++ b/src/backend/parser/parse_expr.c
@@ -2278,7 +2278,7 @@ transformCoalesceExpr(ParseState *pstate, CoalesceExpr *c)
 		newargs = lappend(newargs, newe);
 	}
 
-	if (select_common_type_hook && sql_dialect == SQL_DIALECT_TSQL && c->tsql_is_null)
+	if (sql_dialect == SQL_DIALECT_TSQL && select_common_type_hook && c->tsql_is_null)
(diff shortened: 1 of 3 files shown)


src/backend/parser/parse_expr.c
```diff
@@ -2278,7 +2278,10 @@ transformCoalesceExpr(ParseState *pstate, CoalesceExpr *c)
 		newargs = lappend(newargs, newe);
 	}
 
-	newc->coalescetype = select_common_type(pstate, newargs, "COALESCE", NULL);
+	if (sql_dialect == SQL_DIALECT_TSQL && select_common_type_hook && c->tsql_is_null)
```
KushaalShroff, line 2279: Are we continuing to use coalesce code path and handle the peculiarities of isnull separately anywhere we behave differently from coalesce?

Deepesh125: yes, we are not redesigning this ISNULL. Will continue using coalesce with the help of hook

KushaalShroff (approved):

PR #194, opened 2023-08-25
Enhance Babelfish dump/restore experience

rishabhtanwar29:
### Description
Following two enhancements are done:
1. Skip dumping default Babelfish objects to suppress ignorable errors during restore.
   * Introduced a new function `bbf_selectDumpableObject` which filters the objects
    we don't want to dump.
   * For users (global objects), improved filter condition and query to not dump default
   users. Additionally, entirely skip ALTER USER command if we just need to set
   * Added one helper function `is_ms_shipped` which calls `sys.`OBJECTPROPERTY`
   function to identify default Babelfish objects.
   NOSUPERUSER or NOREPLICATION.
2. Include database users and permissions for logical database level dump.
   * Improved query to fetch the Babelfish roles to also include all the users and roles
   of the specified logical database.
   * Fixed query to dump the data of `babelfish_authid_user_ext` catalog table to
   include all those users and roles we dumped in the above step.

Task: BABEL-4195, BABEL-4206

Extensions PR: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1780
 
### Check List


src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -126,6 +127,30 @@ isBabelfishDatabase(Archive *fout)
...
+static bool
+is_ms_shipped(DumpableObject *dobj, Archive *fout)
+{
+	PQExpBuffer	qry;
+	PGresult	*res;
+	bool    	ismsshipped = false;
+
+	/* Directly query sys.OBJECTPROPERTY function to know whether it is a system object or not. */
+	qry = createPQExpBuffer();
+	appendPQExpBuffer(qry, "SELECT sys.OBJECTPROPERTY(%u, 'ismsshipped');\n", dobj->catId.oid);
```
HarshLunagariya, line 144: Since we are relying on a system object , there can be possibility that due to faulty behaviour on OBJECTPROPERTY() - We may miss dumping some objects. If make sure this doesn't happen, Please verify that we have enough testing for ismsshipped.

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -186,52 +211,142 @@ dumpBabelGUCs(Archive *fout)
...
-			strcmp(sTypeInfo->dobj.name, "bool") == 0 &&
-			(strcmp(tTypeInfo->dobj.name, "bpchar") == 0 ||
-			 strcmp(tTypeInfo->dobj.name, "varchar") == 0))
-		cast->dobj.dump = DUMP_COMPONENT_NONE;
-}
+	switch (dobj->objType)
+	{
+		case DO_CAST:
+			{
+				CastInfo    	*cast = (CastInfo *) dobj;
```
surendravishnoi, line 306: to verify, we want to skip it for dump and pg upgrade?

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -186,52 +211,142 @@ dumpBabelGUCs(Archive *fout)
...
+			break;
+		case DO_NAMESPACE:
+			{
+				NamespaceInfo *nsinfo = (NamespaceInfo *) dobj;
+
+				if (fout->dopt->binary_upgrade)
+					return;
+
+				/* Do not dump the definition of default babelfish schemas */
+				if (strncmp(nsinfo->dobj.name, "babelfishpg", 11) == 0 ||
```
surendravishnoi: can a user create a namespace like babelfshpgxyz? that will pass this check

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -186,52 +211,142 @@ dumpBabelGUCs(Archive *fout)
...
-bbf_selectDumpableTableData(TableInfo *tbinfo, Archive *fout)
-{
-	if (!isBabelfishDatabase(fout) || fout->dopt->binary_upgrade)
-		return;
+				if (fout->dopt->binary_upgrade)
+					return;
 
-	if (isBabelfishConfigTable(tbinfo))
-		tbinfo->dobj.dump |= DUMP_COMPONENT_DATA;
+				if (strncmp(extinfo->dobj.name, "babelfishpg", 11) == 0)
```
surendravishnoi, line 412: same as above, there could be a extension with name babelfishpgxyz

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -186,52 +211,142 @@ dumpBabelGUCs(Archive *fout)
...
+
+				if (fout->dopt->binary_upgrade)
+					return;
+
+				/* Do not dump the definition of default babelfish schemas */
+				if (strncmp(nsinfo->dobj.name, "babelfishpg", 11) == 0 ||
+					strcmp(nsinfo->dobj.name, "master_dbo") == 0 ||
+					strcmp(nsinfo->dobj.name, "master_guest") == 0 ||
+					strcmp(nsinfo->dobj.name, "msdb_dbo") == 0 ||
+					strcmp(nsinfo->dobj.name, "master_guest") == 0 ||
```
2jungkook: `msdb_guest`?

rishabhtanwar29: Nice find, looks like I somehow forgot to include this change later. Thanks for pointing it out.

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: That's right, the only condition we are checking for this is that if it is a Babelfish database (at line no. 220).

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: Right, this I have taken from an engine check where we make babelfish extensions as protected extensions. Moreover, it is going to be quite rare for a user to have a custom extension with prefix `babelfishpg` in babelfish database. (It is also going to be harmless as user can recreate that extension after restore if needed.)

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: Yeah right, I guess this check is valid only for extensions only since as described in comment below, not dumping extension is going to be harmless but not dumping schema can be problematic.
The only reason to add this schema check is to exclude `CREATE SCHEMA` statement for `babelfish_telemetry` schema (which doesn't exist in OSS) so that there is no such error like:
```
ERROR:  schema "babelfishpg_telemetry" already exists
```
But now I have come to believe that getting this error is better than not dumping some user schema `babelfishxyz` so I have removed this schema check completely.

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -186,52 +211,143 @@ dumpBabelGUCs(Archive *fout)
...
+						strcmp(tTypeInfo->dobj.name, "varchar") == 0))
+					cast->dobj.dump = DUMP_COMPONENT_NONE;
+			}
+			break;
+		case DO_TABLE:
+			{
+				TableInfo *tbinfo = (TableInfo *) dobj;
+
+				if (fout->dopt->binary_upgrade)
+					return;
```
HarshLunagariya, line 339: Do we need this condition for CAST object as well?

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -828,15 +944,14 @@ addFromClauseForLogicalDatabaseDump(PQExpBuffer buf, TableInfo *tbinfo, bool is_
...
-		/*
-		 * Builtin db users will already be present in the target
-		 * server so no need to dump their catalog data.
-		 */
-		if (is_builtin_db)
-			appendPQExpBufferStr(buf, "LIMIT 0 ");
+						  "INNER JOIN sys.babelfish_sysdatabases b "
+						  "ON a.database_name = b.name COLLATE \"C\" "
+						  "WHERE b.dbid = %d "
+						  "AND a.rolname NOT IN ('dbo', 'db_owner', "
```
HarshLunagariya: Why are we including 'dbo' and 'db_owner' ? We shouldn't be including that. If it is needed then it should be migration mode dependent. because in single-db it will work, but on multi-db it won't.

src/bin/pg_dump/dumpall_babel_utils.c
```diff
@@ -57,145 +57,118 @@ isBabelfishDatabase(PGconn *conn)
...
+						  "FROM bbf_roles "
 						  "WHERE rolname !~ '^pg_' "
-						  "AND rolname IN ('dbo', 'db_owner', '%s_dbo', '%s_db_owner', '%s_guest') "
-						  "ORDER BY 2 ", role_catalog, role_catalog, escaped_bbf_db_name,
-						  escaped_bbf_db_name, escaped_bbf_db_name);
-
-		/* builtin db users will already be present in the target server so no need to dump them */
-		if (is_builtin_db)
-			appendPQExpBufferStr(buf, "LIMIT 0");
+						  "AND rolname NOT IN ('sysadmin', 'dbo', 'db_owner', "
```
HarshLunagariya: Same comment as last instance for use of 'dbo' and 'db_owner'.

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: Nope, this check is to exclude MVU scenario but CAST check is needed for both MVU and dump/restore.

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: Makes sense, although in worst case, we will dump a default object and user will get an harmless error that the object already exists.

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: Great find, thanks for pointing it out. Actually, now I realized that `dbo` and `db_owner` shouldn't be included here since we do want to dump them as they are object of a user database (single-db). The only objects we don't want to dump (we want to include here) are the ones related to builtin databases (master/msdb/tempdb). Removed in the next revision.

src/bin/pg_dump/dumpall_babel_utils.c
rishabhtanwar29: Same reason as in the reply above, removed both `dbo` and `db_owner` from here.

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -757,11 +873,18 @@ prepareForBabelfishDatabaseDump(Archive *fout, SimpleStringList *schema_include_
...
+	 */
 	appendPQExpBufferStr(query,
-						 "SELECT unnest(extconfig)::oid "
-						 "FROM pg_catalog.pg_extension "
-						 "WHERE extname = 'babelfishpg_tsql';");
+						 "WITH tableoids AS ("
+						 "SELECT unnest(extconfig)::oid AS id "
+						 "FROM pg_catalog.pg_extension WHERE extname = 'babelfishpg_tsql') "
+						 "SELECT id FROM tableoids WHERE id != 'sys.babelfish_configurations'::regclass " /* Exclude babelfish_configurations table */
+						 "UNION SELECT 'sys.babelfish_authid_user_ext'::regclass AS id "); /* Include babelfish_authid_user_ext table */
```
rishabhtanwar29, line 887: The above update is to improve this logic now although it will only be relevant when we support cross-version dump/restore. Currently, this has been added to allow TAP tests to run successfully.

surendravishnoi (approved):

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -757,11 +873,18 @@ prepareForBabelfishDatabaseDump(Archive *fout, SimpleStringList *schema_include_
...
+	 * below.
+	 */
 	appendPQExpBufferStr(query,
-						 "SELECT unnest(extconfig)::oid "
-						 "FROM pg_catalog.pg_extension "
-						 "WHERE extname = 'babelfishpg_tsql';");
+						 "WITH tableoids AS ("
+						 "SELECT unnest(extconfig)::oid AS id "
+						 "FROM pg_catalog.pg_extension WHERE extname = 'babelfishpg_tsql') "
+						 "SELECT id FROM tableoids WHERE id != 'sys.babelfish_configurations'::regclass " /* Exclude babelfish_configurations table */
```
HarshLunagariya, line 886: Why is this needed for only babelfish_configurations?

HarshLunagariya: As per offline discussion with Rishabh, It is not needed because configurations are added at the time of creation and entries are populated at creation time.

HarshLunagariya (approved):

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: Discussed offline.

PR #204, opened 2023-08-29
Block cross version and cross migration mode dump/restore

rishabhtanwar29:
### Description
Babelfish currently only supports dump and restore from databases with
same migration mode ( i.e. either singledb->singledb or multidb->multidb)
and same versions. This commit adds logic to dump a precheck so that the
restore process stops execution if the migration mode or version in the target
server is different than the one from which the dump file was created.

Task: BABEL-4055
 
Extensions PR: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1798
### Check List

### DIFF
--- a/src/bin/pg_dump/dump_babel_utils.c
+++ b/src/bin/pg_dump/dump_babel_utils.c
@@ -186,6 +186,85 @@ dumpBabelGUCs(Archive *fout)
 }
 
 /*
+ * dumpBabelRestoreChecks:
+ * Dumps Babelfish specific pre-checks which get executed at the
+ * beginning of restore to validate if restore can be performed
+ * or not.
+ */
+void
+dumpBabelRestoreChecks(Archive *fout)
+{
+	PGresult	*res;
+	char		*source_server_version;
+	char		*source_migration_mode;
+	PQExpBuffer	qry;
+
+	/* Skip if not Babelfish database or binary upgrade */
+	if (!isBabelfishDatabase(fout) || fout->dopt->binary_upgrade)
+		return;
+
+	/*
+	 * Cross version Babelfish dump/restore is not yet supported so
+	 * store the current server's version in the below procedure and
+	 * add logic to fail the restore if the target server version
+	 * differs from source server version.
+	 */
+	qry = createPQExpBuffer();
+	res = ExecuteSqlQueryForSingleRow(fout, "SHOW server_version");
+	source_server_version = pstrdup(PQgetvalue(res, 0, 0));
+
+	/*
+	 * Temporarily enable ON_ERROR_STOP so that whole restore script
+	 * execution fails if the following do block raises an error.
+	 */
+	appendPQExpBufferStr(qry, "\\set ON_ERROR_STOP on\n\n");
+	appendPQExpBuffer(qry,
+					  "DO $$"
+					  "\nDECLARE"
+					  "\n    target_server_version VARCHAR;"
+					  "\nBEGIN"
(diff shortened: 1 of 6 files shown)


src/bin/pg_dump/dumpall_babel_utils.c
```diff
@@ -56,6 +81,82 @@ isBabelfishDatabase(PGconn *conn)
...
 }
 
+/*
+ * dumpBabelRestoreChecks:
+ * Dumps Babelfish specific pre-checks which get executed at the
+ * beginning of restore to validate if restore can be performed
+ * or not.
+ */
+void
+dumpBabelRestoreChecks(FILE *OPF, PGconn *conn, int binary_upgrade)
```
surendravishnoi, line 91: can we avoid duplicate code by sharing utility function between dump and dumpall?

src/bin/pg_dump/dumpall_babel_utils.c
```diff
@@ -45,7 +70,7 @@ isBabelfishDatabase(PGconn *conn)
 	{
 		PGresult *res;
 		int		 ntups;
-		res = PQexec(conn, "SELECT extname FROM pg_extension WHERE extname = 'babelfishpg_tsql';");
+		res = executeQuery(conn, "SELECT extname FROM pg_extension WHERE extname = 'babelfishpg_tsql';");
```
surendravishnoi, line 73: why will such a failure happen? why do we want to fail the utility? this code is also applicable to non-babelfish? If yes, we should assume bbf_status = OFF in such cases

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -185,6 +185,85 @@ dumpBabelGUCs(Archive *fout)
...
+		return;
+
+	/*
+	 * Cross version Babelfish dump/restore is not yet supported so
+	 * store the current server's version in the below procedure and
+	 * add logic to fail the restore if the target server version
+	 * differs from source server version.
+	 */
+	qry = createPQExpBuffer();
+	res = ExecuteSqlQueryForSingleRow(fout, "SHOW server_version");
```
surendravishnoi, line 213: do we want to block even patch level difference?

src/bin/pg_dump/dumpall_babel_utils.c
rishabhtanwar29: Actually, pg_dump/pg_dumpall as standalone utilities and the function signature (arguments) and handling differs between both of them, so it might get over complicated to write a single function. Moreover, I think below piece of code is rarely going to change in future and eventually, we are going to remove it.

src/bin/pg_dump/dumpall_babel_utils.c
rishabhtanwar29: With this query only we will figure out if the server is a Babelfish or not. The `executeQuery` is a general wrapper for executing a query and handling the error condition. It's better to fail the utility in case server does not return a response (a general case) as we are otherwise probably going to fail at another place due to server not responding correctly.

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: I have not seen the cases where we have made any catalog changes between patch versions so I don't think it's necessary.

surendravishnoi: but that is what we are doing, any difference at all between source and target will lead to failure (even at patch level)

surendravishnoi (dismissed):

surendravishnoi (approved):

PR #205, opened 2023-08-29
Refinement of FOR JSON PATH & Nested Functionality

northaxosky:
Currently FOR JSON PATH is implemented in Babelfish using string manipulation. This change will implement FOR JSON PATH using PostgreSQL's JsonbValue and HTAB. The reason for this change is because the current implementation of FOR JSON PATH does not support the full functionality of the feature. Nested outputs determined by aliases in the SELECT statement are not supported. By implementing FOR JSON PATH using JsonbValue and HTAB, we provide an efficient solution that also supports nested json objects, while also setting the foundation for FOR JSON AUTO to be implemented.

**BABELFISH_EXTENSIONS CHANGES:**
- Switched from String Manipulation implementation of FOR JSON PATH to an implementation using JsonbValue.
- Wrapper function to add a value to JsonbValue and handle datatype checks.
- Supporting functions to determine path to insert a JsonbValue object and key for HashTable.
- Create json and insert into existing json functions that are used when dealing with nested json objects.
- Implemented Nested Output tracking using PostgreSQL's HTAB library (HashTable)
- Corrected the tests to showcase the new functionality of nested json objects
- Added another set of tests to more thoroughly test the nested json object functionality.

**POSTGRESQL_MODIFIED_FOR_BABELFISH CHANGES:**
- Created a public wrapper function in jsonb.h  to call add_jsonb() from forjson.c 

**TEST CASES:**
- Added a test that checks the functionality and accuracy of the nested json feature. Includes cases with multiple layers, deeply nested layers, values in different stages of json layers, nested null outputs.
- Organized version upgrade tests so previous versions use previous implementation of FOR JSON PATH and updated the schedule for each version.
- Modified existing test cases to correctly showcase the nested functionality if the given test uses alias that cause nested json objects to be created.
- Changed the expected outputs to include spaces as JsonbValue does when converted to a string.


src/include/utils/jsonb.h
```diff
@@ -382,11 +382,11 @@ extern int	compareJsonbContainers(JsonbContainer *a, JsonbContainer *b);
 extern JsonbValue *findJsonbValueFromContainer(JsonbContainer *sheader,
 											   uint32 flags,
 											   JsonbValue *key);
-extern JsonbValue *getKeyJsonValueFromContainer(JsonbContainer *container,
-												const char *keyVal, int keyLen,
+extern JsonbValue *getKeyJsonValueFromContainer(JsonbContainer *container,
```
Jakeowen1: nit: space at the end of this line and a few below it

Jakeowen1 (changes requested):

Jakeowen1 (approved):

src/backend/utils/adt/jsonb.c
```diff
@@ -598,6 +598,16 @@ JsonbToCStringWorker(StringInfo out, JsonbContainer *in, int estimated_len, bool
...
 }
 
+void
+jsonb_get_value(Datum val, bool is_null, JsonbValue *json, Oid val_type)
+{
+	JsonbInState state;
+	memset(&state, 0, sizeof(JsonbInState));
+
+	add_jsonb(val, is_null, &state, val_type, false);
+	*json = *(state.res);
```
2jungkook, line 608: It seems like `json` is palloc'd in `tsql_row_to_json()`. To prevent the possibility of memory overwrites, could you double check whether `json` and `state.res` have the same memory context?

Jakeowen1: The `jsonb_get_value` function is called by `tsql_row_to_json` which defines the `json` variable. The memory context is consistent in `tsql_row_to_json`  -  these GDB results were the same after `json` is declared and after `state` is declared 

```
(gdb) print MemoryContextStats(TopMemoryContext)
$5 = void
(gdb) print MemoryContextStats(CacheMemoryContext)
$6 = void
(gdb) print MemoryContextStats(CacheMemoryContext)
$7 = void
(gdb) print MemoryContextStats(PortalContext)
$8 = void
```

2jungkook (approved):

PR #216, opened 2023-09-13
Do not dump Babelfish initialize user

rishabhtanwar29:
### Description
When Babelfish initialize user is same between source and target servers
for dump/restore, the restore throws errors like "role ... already exists" and
"duplicate key value violates unique constraint..." etc while inserting catalog
entry for initialize user in babelfish_authid_login_ext catalog. These errors are
actually expected and makes sense as we know the user do already exists on
the target server but the whole restore gets rolled back in case the restore
process has been made transactional (`--single-transaction` option is used)
and there is no way for a user to fix this manually.

To overcome the above issue, this commit implements the logic to skip dumping
Babelfish initialize user so that there is no conflict when initialize user is same or
different between source and target servers. This is safe since no T-SQL objects
will have direct reference to initialize user and they will adapt to the new initialize
user on the target server.

The only case we have to handle is the "owner" name in `sys.babelfish_sysdatabases`
catalog table, which references to initialize user. For this, we will also skip dumping
the "owner" column and re-populate this column during restore with the owner of
current database (which is going to be initialize user of target Babelfish database).

Task: BABEL-4404 

Extensions PR: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/1841

### Check List

### DIFF
--- a/src/bin/pg_dump/dump_babel_utils.c
+++ b/src/bin/pg_dump/dump_babel_utils.c
@@ -35,6 +35,7 @@ static const CatalogId nilCatalogId = {0, 0};
 static char *escaped_bbf_db_name = NULL;
 static int bbf_db_id = 0;
 static SimpleOidList catalog_table_include_oids = {NULL, NULL};
+static char *babel_init_user = NULL;
 
 static char *getMinOid(Archive *fout);
 static bool isBabelfishConfigTable(TableInfo *tbinfo);
@@ -223,6 +224,7 @@ dumpBabelRestoreChecks(Archive *fout)
 	char		*source_server_version;
 	char		*source_migration_mode;
 	PQExpBuffer	qry;
+	ArchiveFormat format = ((ArchiveHandle *) fout)->format;
 
 	/* Skip if not Babelfish database or binary upgrade */
 	if (!isBabelfishDatabase(fout) || fout->dopt->binary_upgrade)
@@ -241,8 +243,10 @@ dumpBabelRestoreChecks(Archive *fout)
 	/*
 	 * Temporarily enable ON_ERROR_STOP so that whole restore script
 	 * execution fails if the following do block raises an error.
+	 * Note that it can only be used in plain text dump (archNull).
 	 */
-	appendPQExpBufferStr(qry, "\\set ON_ERROR_STOP on\n\n");
+	if (format == archNull)
+		appendPQExpBufferStr(qry, "\\set ON_ERROR_STOP on\n\n");
 	appendPQExpBuffer(qry,
 					  "DO $$"
 					  "\nDECLARE"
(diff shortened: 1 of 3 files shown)


src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -1278,19 +1315,13 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
  * Returns true if table in Babelfish Database is to be dumped with INSERT mode.
  * Currently we dump tables with sql_variant columns with INSERT operations to
  * correctly restore the metadata of the base datatype, which is not directly
- * posible with COPY statements. We also dump sys.babelfish_authid_login_ext
- * with INSERT statements so that if target database already has a login with
- * same name as in the source database, only that INSERT query with fail and won't
- * affect the other entries of the catalog table. 
+ * posible with COPY statements.
```
HarshLunagariya: s/posible/possible

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -1278,19 +1315,13 @@ fixCopyCommand(Archive *fout, PQExpBuffer copyBuf, TableInfo *tbinfo, bool isFro
...
-bool bbfIsDumpWithInsert(Archive *fout, TableInfo *tbinfo)
+bool
+bbfIsDumpWithInsert(Archive *fout, TableInfo *tbinfo)
 {
 	return (isBabelfishDatabase(fout) &&
-			(hasSqlvariantColumn(tbinfo) ||
-				pg_strcasecmp(fmtQualifiedDumpable(tbinfo),
-					quote_all_identifiers ?
-					"\"sys\".\"babelfish_authid_login_ext\"" :
-					"sys.babelfish_authid_login_ext") == 0));
```
HarshLunagariya, line 1293: Why are we removing this?

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -890,6 +895,18 @@ prepareForBabelfishDatabaseDump(Archive *fout, SimpleStringList *schema_include_
...
+	resetPQExpBuffer(query);
+	
+	/*
+	 * Find out initialize user of current Babelfish database.
+	 */
+	appendPQExpBufferStr(query, "SELECT r.rolname FROM pg_roles r "
+						 "INNER JOIN pg_database d ON r.oid = d.datdba "
+						 "WHERE d.datname = current_database()");
+	res = ExecuteSqlQueryForSingleRow(fout, query->data);
+	babel_init_user = pstrdup(PQgetvalue(res, 0, 0));
```
HarshLunagariya: Can we reuse getBabelfishInitUser() here?

rishabhtanwar29: Actually that function is defined in pg_dumpall, which has separate code base from pg_dump as both are standalone binaries.

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: Actually this change was previously done to forcefully dump babelfish_authid_login_ext data with INSERT statements so that even if one INSERT fails for existing babelfish init user, other INSERTs should still go through. But now that we have decided to not dump babelfish init user, we can safely dump this table with COPY as well. Basically we don't wan't to pose any restriction on the table data will get dumped for this table.

HarshLunagariya (approved):

shardgupta (approved):

PR #217, opened 2023-09-18
BABEL-4411 set transaction isolation cmd change

tanscorpio7:
### Description
Currently babelfish transactions start with session default isolation level. Users can not change this isolation level after transaction has started. On running the set transaction isolation inside a transaction, current transaction isolation does not change but all subsequent transaction will pick up this new isolation level.

Changes in this PR:

Allow changing the current transaction isolation level if no DML query has been run inside transaction.
Abort the transaction if set transaction isolation level is run after a DML query in a transaction block unless the new isolation level is same as current isolation level.
Introduce an escape hatch to ignore any invalid set transaction isolation level stmt so that transaction does not abort.

#### Engine Change:
Add check hook to guc default_transaction_isolation (all errors should be taken care here, that is assign hook should never fail)
If the conditions for setting isolation level are true returns true
If conditions are false and escape hatch is not set (strict - default value) - abort transaction with error
but if escape hatch is set to ignore for same situation - we simply ignore the set tran isolation command by setting the new isolation level to current isolation level. Which is apparently the same as ignoring the cmd.

#### Extension Change:
Introduce escape hatch for set transaction isolation 
New attribute in PLtsql_stmt_execsql structure. This attribute is needed to identify if the stmt is set transaction isolation during initialisation of internal savepoint and if set, we skip the internal save point for the current stmt.
This needs to be done since postgres does not allow setting isolation level inside a save point.

### Issues Resolved

BABEL-4411

### Check List

### DIFF
--- a/src/backend/utils/misc/guc.c
+++ b/src/backend/utils/misc/guc.c
@@ -12263,16 +12263,10 @@ assign_default_transaction_isolation(int newval, void *extra)
 	if (guc_newval_hook)
 	{
 		char		*s;
-		A_Const		*t;
 		s = transaction_isolation_to_string(newval);
-		t = makeNode(A_Const);;
-		t->val.sval.type = T_String;
-		t->val.sval.sval = s;
-		t->location = -1;
-		if( newval != XactIsoLevel )
-			SetPGVariable("transaction_isolation", list_make1(t), true);
+		set_config_option("transaction_isolation", s, PGC_USERSET,
+							PGC_S_SESSION, GUC_ACTION_LOCAL, true, 0, false);
 		(*guc_newval_hook)("default_transaction_isolation", false, NULL, newval);
-		pfree(t);
 	}
 }
 
--- a/src/backend/utils/misc/guc.c
+++ b/src/backend/utils/misc/guc.c
@@ -12264,8 +12264,8 @@ assign_default_transaction_isolation(int newval, void *extra)
 	{
 		char		*s;
 		s = transaction_isolation_to_string(newval);
-		set_config_option("transaction_isolation", s, PGC_USERSET,
-							PGC_S_SESSION, GUC_ACTION_LOCAL, true, 0, false);
+		SetConfigOption("transaction_isolation", s,
+							PGC_USERSET, PGC_S_SESSION);
 		(*guc_newval_hook)("default_transaction_isolation", false, NULL, newval);
 	}
 }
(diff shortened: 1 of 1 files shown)


src/backend/utils/misc/guc.c
```diff
@@ -12259,7 +12261,13 @@ static void
 assign_default_transaction_isolation(int newval, void *extra)
 {
 	if (guc_newval_hook)
+	{
```
surendravishnoi, line 12264: we should put this code into hook as well

src/backend/utils/misc/guc.c
```diff
@@ -13075,4 +13083,51 @@ guc_set_stack_value(struct config_generic *gconf, config_var_value *val)
...
+		case XACT_READ_COMMITTED:
+			return "read committed";
+		case XACT_READ_UNCOMMITTED:
+			return "read uncommitted";
+	}
+	return "(unrecognized)";
+}
+
+static bool
+check_default_transaction_isolation(int *newval, void **extra, GucSource source)
```
surendravishnoi, line 13104: any reason for having this as a separate function? can we put it new_val hook?

src/backend/utils/misc/guc.c
```diff
@@ -13075,4 +13083,51 @@ guc_set_stack_value(struct config_generic *gconf, config_var_value *val)
...
+				ignore_isolation = GetConfigOption("babelfishpg_tsql.escape_hatch_set_transaction_isolation_level", false, false);
+
+				if(ignore_isolation && !strcmp(ignore_isolation, "ignore"))
+				{
+					*newval = XactIsoLevel;
+					return true;
+				}
+				else
+				{
+					GUC_check_errmsg("SET TRANSACTION ISOLATION failed because of setting from invalid state, set escape hatch to silently ignore");
```
surendravishnoi, line 13124: we should provide the name of escape hatch

src/backend/utils/misc/guc.c
```diff
@@ -13075,4 +13083,51 @@ guc_set_stack_value(struct config_generic *gconf, config_var_value *val)
...
+
+		if ( newXactIsoLevel != XactIsoLevel && IsTransactionState() )
+		{
+			if( FirstSnapshotSet || IsSubTransaction() || (newXactIsoLevel == XACT_SERIALIZABLE && RecoveryInProgress()))
+			{
+				ignore_isolation = GetConfigOption("babelfishpg_tsql.escape_hatch_set_transaction_isolation_level", false, false);
+
+				if(ignore_isolation && !strcmp(ignore_isolation, "ignore"))
+				{
+					*newval = XactIsoLevel;
```
surendravishnoi, line 13119: if we put everything in a single hook, we can simply skip setting transaction isolation level

src/backend/utils/misc/guc.c
tanscorpio7: So having a check hook will make sure that we set both 'default_transaction_Isolation' and 'transaction_isolation' together and never just 'default_transaction_Isolation'

src/backend/utils/misc/guc.c
tanscorpio7: So, this was needed to make sure that default and transaction level both are always same.
We would never set just one of them 
And postgres suggests that assign hook never fails

PR #226, opened 2023-10-04
Do not dump default babelfishpg_telemetry schema

rishabhtanwar29:
### Description
Skip dumping default babelfishpg_telemetry schema as it would
already exists on the target server.

### Check List

### DIFF
--- a/src/bin/pg_dump/dump_babel_utils.c
+++ b/src/bin/pg_dump/dump_babel_utils.c
@@ -314,7 +314,7 @@ bbf_selectDumpableObject(DumpableObject *dobj, Archive *fout)
 				ExtensionInfo	*ext = findOwningExtension(cast->dobj.catId);
 
 				/* Skip if cast is not a member of babelfish extension */
-				if (ext == NULL || strcmp(ext->dobj.name, "babelfishpg_common") != 0)
+				if (ext == NULL || strncmp(ext->dobj.name, "babelfishpg_common", 18) != 0)
 					return;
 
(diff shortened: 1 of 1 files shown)


src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -314,7 +314,7 @@ bbf_selectDumpableObject(DumpableObject *dobj, Archive *fout)
 				ExtensionInfo	*ext = findOwningExtension(cast->dobj.catId);
 
 				/* Skip if cast is not a member of babelfish extension */
-				if (ext == NULL || strcmp(ext->dobj.name, "babelfishpg_common") != 0)
+				if (ext == NULL || strncmp(ext->dobj.name, "babelfishpg_common", 18) != 0)
```
surendravishnoi: what is the rationale behind this? with this any name with babelfish_common as prefix will match, is that desired beahvior?

rishabhtanwar29: The main reason was that `strcmp` is considered unsafe c function and would come in the security findings but I just looked at the tool again and looks like it does not consider this as unsafe. So I guess we can skip this.

surendravishnoi: using strncmp is ok but if we do not compare length, we can get incorrect semantics.
strcmp(str1, str2) is not same as strncmp(str1, str2, strlen(str2) (str1 and str2 will match as long as str2 is a substring of str1, which is incorrect)

surendravishnoi (approved):

PR #227, opened 2023-10-05
Schema Resolution for multiple functions in single stmt

tanscorpio7:
### Description

In SQL Server the default schema inside  functions is same as functions schema, followed by the actual query default schema. We previously handled this behaviour on a statement level which worked correctly when only one schema & one function was used in the statement. When multiple functions and schemas are specified or the schema is specified with some other object along with a non schema specified function, this approach becomes ineffective, since we would resolve everything to the one schema.

To solve this we must resolve the schema (pg search path) at function execution level. We should only do this for user defined functions. All procedures & triggers will be excluded from the change since their schema resolution follows different strategy and is already implemented else where in the code. Cross DB functions call are not allowed in babelfish, so need not to look at those cases. System defined functions (sys, pg_catalog, ...) should be excluded from this change since they are expected to run in postgres environment.

The implementation uses the SET option available for PG function definitions. This SET cmd is picked up from PG catalog and locally set at run time(only for function execution). We do not store this information in the catalog itself. Instead we search the function owner schema and call the SET command during function initialisation.
But instead of adding the new search path to proconfig attribute of functions, we set it directly using the search_path global variable, to avoid performance drop.

We also store the new search path in fcache->prosearchpath to avoid multiple calls.
And we skip setting search path if language is not pltsql 

ex: Select s2.f();

s2.f() {
   -- first we search objects in s2
   -- then we search in database default schema s2.dbo
}

### Issues Resolved

[BABEL-4442]
 
### Check List

### DIFF
--- a/src/backend/utils/fmgr/fmgr.c
+++ b/src/backend/utils/fmgr/fmgr.c
@@ -708,7 +708,7 @@ fmgr_security_definer(PG_FUNCTION_ARGS)
 	int			non_tsql_proc_count = 0;
 	void	   *newextra = NULL;
 	char 	   *cacheTupleProcname = NULL;
-	char	   *old_search_path = false;
+	char	   *old_search_path = NULL;
 
 	if (!fcinfo->flinfo->fn_extra)
 	{
@@ -895,7 +895,6 @@ fmgr_security_definer(PG_FUNCTION_ARGS)
 			if(sql_dialect == tsql_dialect && old_search_path)
 			{
 				namespace_search_path = old_search_path;
-				elog(LOG, "-------------> %s", old_search_path);
 				assign_search_path(old_search_path, newextra);
 			}
 			sql_dialect = sql_dialect_value_old;
@@ -913,7 +912,6 @@ fmgr_security_definer(PG_FUNCTION_ARGS)
 		if(sql_dialect == tsql_dialect && old_search_path)
 		{
 			namespace_search_path = old_search_path;
-			elog(LOG, "-------------> %s", old_search_path);
 			assign_search_path(old_search_path, newextra);
 		}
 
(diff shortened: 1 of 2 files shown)


src/backend/utils/fmgr/fmgr.c
```diff
@@ -789,7 +790,20 @@ fmgr_security_definer(PG_FUNCTION_ARGS)
 	if (set_sql_dialect && IsTransactionState())
 	{
 		if ((fcache->prolang == pltsql_lang_oid) || (fcache->prolang == pltsql_validator_oid))
+		{
 			sql_dialect_value = tsql_dialect;
+			if (set_local_schema_for_func_hook)
+			{
+				char 	*new_search_path = NULL;
+				new_search_path = (*set_local_schema_for_func_hook)(fcinfo->flinfo->fn_oid);
```
surendravishnoi: this will not work, perf impact will be not acceptable. We are searching two catalogs and changing search path. This will create big penalties for function call. Let us see if we can optimize this by using fcache (store search path there). Also, do perf experiments to see how much penalty we hit due to this.

surendravishnoi (commented):
have we done some perf run to see the impact of this change?

src/backend/utils/fmgr/fmgr.c
```diff
@@ -881,16 +904,27 @@ fmgr_security_definer(PG_FUNCTION_ARGS)
...
+		}
 
 		PG_RE_THROW();
 	}
 	PG_END_TRY();
 
 	fcinfo->flinfo = save_flinfo;
 
-	if (set_sql_dialect)
+	if(old_search_path)
```
surendravishnoi: if (

src/backend/utils/fmgr/fmgr.c
```diff
@@ -786,6 +802,13 @@ fmgr_security_definer(PG_FUNCTION_ARGS)
 						GUC_ACTION_SAVE);
 	}
 
+	if (fcache->prosearchpath)
+	{
+			old_search_path = namespace_search_path;
```
surendravishnoi: indentation

surendravishnoi (dismissed):

src/backend/utils/fmgr/fmgr.c
```diff
@@ -707,6 +709,18 @@ fmgr_security_definer(PG_FUNCTION_ARGS)
...
+
+	if (get_func_language_oids_hook)
+		get_func_language_oids_hook(&pltsql_lang_oid, &pltsql_validator_oid);
+	else
+	{
+		pltsql_lang_oid = InvalidOid;
+		pltsql_validator_oid = InvalidOid;
+	}
+
+	// get_language_procs("pltsql", &pltsql_lang_oid, &pltsql_validator_oid);
```
shalinilohia50: Nit: Remove this unnecessary line.

shalinilohia50 (approved):

surendravishnoi (approved):

PR #232, opened 2023-10-06
OID Wraparound

time-change:
### Description

This change adds a separate OID buffer for temp objects. Since temp objects are already session-local, we can allocate a separate buffer of OIDs set aside only for temp objects. This same buffer range can be used across all sessions.

 
### Check List


src/backend/catalog/toasting.c
```diff
@@ -198,7 +198,7 @@ create_toast_table(Relation rel, Oid toastOid, Oid toastIndexOid,
 	/*
 	 * Create the toast table and its index
 	 */
-	if (sql_dialect == SQL_DIALECT_TSQL && RelationIsBBFTableVariable(rel))
+	if (sql_dialect == SQL_DIALECT_TSQL && (RelationIsBBFTableVariable(rel) || RelationIsBBFTempTable(rel)))
```
surendravishnoi: is this the right thing to do? I guess creating as @pg_toast will make it a table variable. Do we want the same thing for temp tables?

src/backend/utils/misc/queryenvironment.c
```diff
@@ -878,15 +878,21 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
 					CacheInvalidateHeapTuple(catalog_rel, newtup, NULL);
 					break;
 				case ENR_OP_UPDATE:
-					oldtup = lfirst(lc);
-					lfirst(lc) = heap_copytuple(tup);
-					CacheInvalidateHeapTuple(catalog_rel, oldtup, tup);
+					if (lc)
```
surendravishnoi: what use case is there for lc being NULL?

src/include/utils/rel.h
```diff
@@ -700,4 +700,11 @@ extern void RelationDecrementReferenceCount(Relation rel);
 	 strlen((relation)->rd_rel->relname.data) >= 1 && \
 	 (relation)->rd_rel->relname.data[0] == '@')
 
+#define RelationIsBBFTempTable(relation) \
+	((relation)->rd_rel->relpersistence == RELPERSISTENCE_TEMP && \
+	 (relation)->rd_rel->relname.data && \
+	 strlen((relation)->rd_rel->relname.data) >= 1 && \
+	 (relation)->rd_rel->relname.data[0] == '#')
```
surendravishnoi: is this a safe check? can PG have tables with # in temp persistence?

src/backend/access/transam/varsup.c
```diff
@@ -514,6 +516,71 @@ ForceTransactionIdLimitUpdate(void)
...
+
+			/* If the OID range would wraparound, start from beginning instead. */
+			if (tempOidStart + temp_oid_buffer_size < tempOidStart) 
+			{
+				tempOidStart = FirstNormalObjectId;
+			}
+			temp_oid_buffer_start = tempOidStart + INT_MIN;
+			ShmemVariableCache->tempOidStart = tempOidStart;
+			ShmemVariableCache->tempOidBufferSize = temp_oid_buffer_size;
+			TempVariableCache = (VariableCache) palloc0(sizeof(*TempVariableCache));
```
surendravishnoi: what is the use of this temp variable cache?

src/backend/access/transam/varsup.c
```diff
@@ -514,6 +516,71 @@ ForceTransactionIdLimitUpdate(void)
...
+
+			TempVariableCache->nextOid = (Oid) tempOidStart;
+			nextTempOid = (Oid) tempOidStart;
+
+			result = tempOidStart;
+		}
+		LWLockRelease(OidGenLock);
+	}
+
+	real_temp_oid_buffer_start = temp_oid_buffer_start + INT_MIN;
```
surendravishnoi: doesn't temp_oid_buffer_start already have INT_MIN added above?

src/backend/access/transam/varsup.c
```diff
@@ -514,6 +516,71 @@ ForceTransactionIdLimitUpdate(void)
...
+			}
+			temp_oid_buffer_start = tempOidStart + INT_MIN;
+			ShmemVariableCache->tempOidStart = tempOidStart;
+			ShmemVariableCache->tempOidBufferSize = temp_oid_buffer_size;
+			TempVariableCache = (VariableCache) palloc0(sizeof(*TempVariableCache));
+
+			TempVariableCache->nextOid = (Oid) tempOidStart;
+			nextTempOid = (Oid) tempOidStart;
+
+			result = tempOidStart;
```
surendravishnoi: this is unnecessary

src/backend/access/transam/varsup.c
```diff
@@ -514,6 +516,71 @@ ForceTransactionIdLimitUpdate(void)
...
+		}
+		LWLockRelease(OidGenLock);
+	}
+
+	real_temp_oid_buffer_start = temp_oid_buffer_start + INT_MIN;
+
+	/*
+	 * Check for wraparound of the temp OID buffer.
+	 */
+	if (nextTempOid >= (Oid) (real_temp_oid_buffer_start + temp_oid_buffer_size) || nextTempOid < (Oid) real_temp_oid_buffer_start)
```
surendravishnoi: which of these are INT_MIN adjusted? can they we compared?

src/backend/catalog/catalog.c
```diff
@@ -558,10 +665,31 @@ GetNewRelFileNode(Oid reltablespace, Relation pg_class, char relpersistence)
 
 		/* Generate the OID */
 		if (pg_class)
-			rnode.node.relNode = GetNewOidWithIndex(pg_class, ClassOidIndexId,
+		{
+			if (relpersistence == RELPERSISTENCE_TEMP && sql_dialect == SQL_DIALECT_TSQL)
+			{
+				/* Temp tables use temp OID logic */
+				rnode.node.relNode = GetNewTempOidWithIndex(pg_class, ClassOidIndexId,
```
surendravishnoi: why have we written this new method? wouldn't GetNewOidWithIndex handle both cases?

robverschoor (changes requested):
Please add test cases that address these cases:

- exceeding 64K temp object in a session
- a wraparound scenario
- a scenario whereby the 64K range overlaps with an already-existing OID

src/backend/access/transam/varsup.c
time-change: will remove. It was leftover from the original POC, now we just use the static int.

time-change:
> Please add test cases that address these cases:
> 
> * exceeding 64K temp object in a session
> * a wraparound scenario
> * a scenario whereby the 64K range overlaps with an already-existing OID

I do plan to add these test cases (see quip doc for full test plan) - as mentioned, we'll need to add new integration tests. So these cases probably can't be part of this PR. One of the issues for example, creating over 64k temp objects in a session will time out in GitHub actions. With our own integration tests we can have more flexibility here. Especially with wraparound, if we're talking about just temp OID wraparound then that will be the same issue, we need to create enough temp objects to wrap around the buffer. General OID wraparound in postgres is even harder to trigger, we'd need to create billions of OIDs. Overlap is difficult to guarantee as well, it would basically necessitate postgres OID wraparound to test.

src/include/utils/rel.h
time-change: no, PG can't have '#' in table names

src/backend/catalog/catalog.c
time-change: GetNewTempOidWithIndex calls GetNewTempObjectId instead of GetNewObjectId. We could merge the two, but it would require a refactoring of GetNewOidWithIndex to see relpersistence (intended for the generated OID), right? This way is less invasive to engine code IMO. The relation passed in is a system relation so it will be permanent.

lejaokri:
This is a pretty sizeable code change and I am wondering if it is possible to have some concurrency tests as part of this PR? I am not sure if any of the Github Actions have concurrency tests in them, if not, can we do some manual tests just to verify? Describe the concurrency tests and the results in the Test Plan.

src/backend/catalog/toasting.c
time-change: the index for the toast table needs to be in ENR or it will cause cache lookup errors. changD the toast_prefix to match temp table naming convention

src/backend/utils/init/postinit.c
```diff
@@ -1141,6 +1141,8 @@ InitPostgres(const char *in_dbname, Oid dboid,
 	/* Initialize this backend's session state. */
 	InitializeSession();
 
+	TempVariableCache = (VariableCache) palloc0(sizeof(*TempVariableCache));
```
lejaokri: I could be missing it but I dont see where `TempVariableCache` is used.

src/backend/utils/misc/queryenvironment.c
time-change: There is a case in table_variable_xact_nested where lc is NULL causing a crash during cleanup. I will look into it more.

src/backend/access/transam/varsup.c
```diff
@@ -514,6 +516,67 @@ ForceTransactionIdLimitUpdate(void)
...
+			nextTempOid = ShmemVariableCache->tempOidStart;
+		}
+		else
+		{
+			/* We need to pick a new start for the buffer range. */
+			tempOidStart = ShmemVariableCache->nextOid;
+			if (ShmemVariableCache->oidCount < temp_oid_buffer_size)
+				ShmemVariableCache->oidCount = 0;
+			else
+				ShmemVariableCache->oidCount -= temp_oid_buffer_size;
```
lejaokri: What is the reason behind this logic?

src/backend/access/transam/varsup.c
```diff
@@ -514,6 +516,67 @@ ForceTransactionIdLimitUpdate(void)
...
+
+			/* If the OID range would wraparound, start from beginning instead. */
+			if (tempOidStart + temp_oid_buffer_size < tempOidStart) 
+			{
+				tempOidStart = FirstNormalObjectId;
+			}
+			temp_oid_buffer_start = tempOidStart + INT_MIN;
```
lejaokri: What is the main idea of doing + INT_MIN and - INT_MIN in some places?

src/backend/access/transam/varsup.c
time-change: ShmemVariableCache->oidCount gets decremented for each normal OID generated, but we need to take into account the number of OIDs we've reserved for temp objects, right?

src/backend/access/transam/varsup.c
time-change: Here we need to translate real OIDs into temp_oid_buffer_start, which is storing OIDs in an int, so we have to slide all the values back. 

Below real_temp_oid_buffer_start is an Oid, so we're just doing the opposite.

src/backend/catalog/toasting.c
```diff
@@ -201,6 +201,9 @@ create_toast_table(Relation rel, Oid toastOid, Oid toastIndexOid,
 	if (sql_dialect == SQL_DIALECT_TSQL && RelationIsBBFTableVariable(rel))
 		pg_toast_prefix = "@pg_toast";
 
+	if (sql_dialect == SQL_DIALECT_TSQL && RelationIsBBFTempTable(rel))
+		pg_toast_prefix = "#pg_toast";
```
lejaokri: I think it's not as simple as Table Variables. All Table Variables are in ENR, but not all temp tables are in ENR. Furthermore, temp table and its corresponding toast table may not have the same columns so you may end up with temp table not in ENR but toast table in ENR or vise versa. 

Consider the following tests which now fails with this code but clean before.
```
CREATE TYPE typa FROM int
GO

CREATE TYPE typb FROM nvarchar(100)
GO

-- Temp table wont be in ENR because of the CheckTempTableHasDependencies in 
-- heap_create_with_catalog but the toast table is in ENR. 
-- So you may need additional changes in heap_create_with_catalog().
CREATE TABLE #temp_3(a typa, b typb)
GO

DROP TABLE #temp_3
GO

-- This fails now but it used to be clean.
CREATE TABLE #temp_3(a typa, b typb)
GO

DROP TYPE typa
DROP TYPE typb
GO
```

src/backend/utils/misc/queryenvironment.c
lejaokri: The lc is NULL is related to this https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/232#discussion_r1366272681

lejaokri: We should remove the lc is NULL check since the oid change should not affect the logic here.
Unless there is a separate bug you're trying to fix.

I think once the issue in my comment https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/232#discussion_r1366272681 is fixed then you do not need to add the if(lc) check here anymore.

src/include/utils/rel.h
surendravishnoi: create table "#t1"(c1 int) works

src/backend/access/transam/varsup.c
```diff
@@ -514,6 +515,67 @@ ForceTransactionIdLimitUpdate(void)
...
+		if (OidIsValid(ShmemVariableCache->tempOidStart) && (ShmemVariableCache->tempOidBufferSize == temp_oid_buffer_size))
+		{
+			/* If ShmemVariableCache->tempOidStart is valid, it could also mean the buffer size GUC has been altered, and we need to pick a new start. */
+			temp_oid_buffer_start = ShmemVariableCache->tempOidStart + INT_MIN;
+			nextTempOid = ShmemVariableCache->tempOidStart;
+		}
+		else
+		{
+			/* We need to pick a new start for the buffer range. */
+			tempOidStart = ShmemVariableCache->nextOid;
```
lejaokri: Does this logic need to handle the case `if (ShmemVariableCache->nextOid < ((Oid) FirstNormalObjectId))` similar to `GetNewObjectId()`? Consider the case when `ShmemVariableCache->nextOid == 0` then we will end up using oids below `FirstNormalObjectId`

src/include/utils/rel.h
time-change: ack, I will leave a comment, but the parent check should still be safe since we're checking sql_dialect as well

src/backend/catalog/index.c
```diff
@@ -950,8 +950,16 @@ index_create(Relation heapRelation,
 		}
 		else
 		{
-			indexRelationId =
-				GetNewRelFileNode(tableSpaceId, pg_class, relpersistence);
+			if (is_enr)
+			{
+				/* Index OIDs must be kept in normal OID range. */
+				indexRelationId = GetNewRelFileNode(tableSpaceId, pg_class, RELPERSISTENCE_PERMANENT);
```
lejaokri: Why is this? This change would make the code in `GetNewRelFileNode()` go to a different code path.(see the switch statement in `GetNewRelFileNode()`)

src/backend/catalog/catalog.c
```diff
@@ -558,10 +625,31 @@ GetNewRelFileNode(Oid reltablespace, Relation pg_class, char relpersistence)
...
+			if (relpersistence == RELPERSISTENCE_TEMP && sql_dialect == SQL_DIALECT_TSQL)
+			{
+				/* Temp tables use temp OID logic */
+				rnode.node.relNode = GetNewTempObjectId();
+			}
+			else
+			{
+				rnode.node.relNode = GetNewObjectId();
+			}
+		}
```
lejaokri: The code will look less invasive if we move most of the new logic generating OID into hook functions

src/backend/access/transam/varsup.c
lejaokri: Ok thanks. nitpick: may be more readable to introduce new macros.

```
-- you can change the name ofcourse

#define OID_TO_BUFFER_START_LOCATION(oid) ((oid) + INT_MIN)
#define BUFFER_START_LOCATION_TO_OID(bufstart) ((bufstart) - INT_MIN)

```

src/backend/access/transam/varsup.c
lejaokri: I see. Thanks! Please add a comment here.

lejaokri (commented):
I am done reviewing for now.

src/backend/catalog/index.c
time-change: Seems like the OID change is altering deletion order, causing crash in performMultipleDeletions. I am looking into this (related to one of the test cases in 3168)

src/backend/access/transam/varsup.c
```diff
@@ -514,6 +518,72 @@ ForceTransactionIdLimitUpdate(void)
...
+
+			/*
+			 * Decrement ShmemVariableCache->oidCount to take into account the new buffer we're allocating
+			 */
+			if (ShmemVariableCache->oidCount < temp_oid_buffer_size)
+				ShmemVariableCache->oidCount = 0;
```
surendravishnoi: should be reconcile regular next oid here rather than doing it inside getnextoid?

src/backend/access/transam/varsup.c
```diff
@@ -570,6 +640,12 @@ GetNewObjectId(void)
 		}
 	}
 
+		/* If we're in tempOid buffer range, skip ahead. */
+	if (OidIsValid(temp_oid_buffer_start)
```
surendravishnoi: temp_oid_buffer_start is local per backend. What if a different backend has acquired temp ids? Shouldn't we check  ShmemVariableCache->tempOidStart instead?

src/backend/access/transam/varsup.c
```diff
@@ -514,6 +518,72 @@ ForceTransactionIdLimitUpdate(void)
...
+	if (RecoveryInProgress())
+		elog(ERROR, "cannot assign OIDs during recovery");
+
+	if (!OidIsValid((Oid) temp_oid_buffer_start)) /* InvalidOid means it needs assigning */
+	{
+		/* First check to see if another connection has already picked a start, then update. */
+		LWLockAcquire(OidGenLock, LW_EXCLUSIVE);
+		if (OidIsValid(ShmemVariableCache->tempOidStart) && (ShmemVariableCache->tempOidBufferSize == temp_oid_buffer_size))
+		{
+			/* If ShmemVariableCache->tempOidStart is valid, it could also mean the buffer size GUC has been altered, and we need to pick a new start. */
```
surendravishnoi: can diff backends have diff value for temp_oid_buffer_size? how will that work?

src/backend/access/transam/varsup.c
```diff
@@ -514,6 +518,72 @@ ForceTransactionIdLimitUpdate(void)
...
+			temp_oid_buffer_start = OID_TO_BUFFER_START(ShmemVariableCache->tempOidStart);
+			nextTempOid = ShmemVariableCache->tempOidStart;
+		}
+		else
```
surendravishnoi: if we pick new range due to buffer size change, are we throwing away current range? also different backends will pick different ranges if buffer size is diff?

src/include/access/transam.h
```diff
@@ -267,6 +274,7 @@ extern bool TransactionStartedDuringRecovery(void);
 
 /* in transam/varsup.c */
 extern PGDLLIMPORT VariableCache ShmemVariableCache;
+extern PGDLLIMPORT VariableCache TempVariableCache;
```
surendravishnoi: unnecessary

src/backend/catalog/catalog.c
```diff
@@ -496,6 +497,71 @@ GetNewOidWithIndex(Relation relation, Oid indexId, AttrNumber oidcolumn)
...
+
+		/* see notes above about using SnapshotAny */
+		scan = systable_beginscan(relation, indexId, true,
+								  SnapshotAny, 1, &key);
+
+		collides = HeapTupleIsValid(systable_getnext(scan));
+
+		systable_endscan(scan);
+
+		/*
```
surendravishnoi: comment is obsolete

src/backend/catalog/catalog.c
```diff
@@ -496,6 +497,71 @@ GetNewOidWithIndex(Relation relation, Oid indexId, AttrNumber oidcolumn)
...
+		 * Log that we iterate more than GETNEWOID_LOG_THRESHOLD but have not
+		 * yet found OID unused in the relation. Then repeat logging with
+		 * exponentially increasing intervals until we iterate more than
+		 * GETNEWOID_LOG_MAX_INTERVAL. Finally repeat logging every
+		 * GETNEWOID_LOG_MAX_INTERVAL unless an unused OID is found. This
+		 * logic is necessary not to fill up the server log with the similar
+		 * messages.
+		 */
+		if (OidIsValid((Oid) temp_oid_buffer_start) && retries >= temp_oid_buffer_size)
```
surendravishnoi: if oid buffer size is very large, should we also log with some interval?

src/backend/catalog/catalog.c
```diff
@@ -583,7 +671,14 @@ GetNewRelFileNode(Oid reltablespace, Relation pg_class, char relpersistence)
 			collides = false;
 		}
 
+		if (OidIsValid((Oid) temp_oid_buffer_start) && tries > temp_oid_buffer_size)
```
surendravishnoi: in case we could not get oids in current range, should we try for a new range?

src/backend/access/transam/varsup.c
time-change: Not at the moment, temp_oid_buffer_size is currently a fixed size. In the future we may add support for user changing this value, for now I will remove this statement since that's been removed as part of this PR. 

Allowing users to change this would be possible in the future, but we would need to check how cleanly GUC changes are passed along to other backends, and we need to coordinate with CP as well to introduce a new GUC.

src/backend/catalog/catalog.c
time-change: Buffer size is fixed, but we can log if number of used OIDs is very large - say over 80% utilization

src/backend/catalog/index.c
time-change: I think we could hold this specific change for a later PR. I will continue to work on it, but it seems a little bit more complicated than I expected. Many if not all dependencies will need to use temp OIDs in certain cases cases - pg_constraint, pg_attrdef, toasting, types. This appears to me to be due in part to deletion order, but there may be other underlying issues as well. So indexes and dependencies will continue to use normal OIDs. Temp tables, TV, toast, will use temp oid.

src/backend/catalog/catalog.c
time-change: I think this could only happen after wraparound.

src/backend/access/transam/varsup.c
```diff
@@ -24,12 +26,16 @@
...
 #include "storage/pmsignal.h"
 #include "storage/proc.h"
+#include "utils/guc.h"
 #include "utils/syscache.h"
 
 
 /* Number of OIDs to prefetch (preallocate) per XLOG write */
 #define VAR_OID_PREFETCH		8192
 
+#define OID_TO_BUFFER_START(oid) 		((oid) + INT_MIN)
```
surendravishnoi: why do we need to do +/- INT_MIN? we can keep all involved vars as uint32 for simplicity

time-change: The value temp_oid_buffer_start is an int, which is why the Oids needs to be properly converted. But we must treat it like a uint32 by adjusting it because it might store an oid buffer start value greater than INT_MAX. 

temp_oid_buffer_start is an int because it's stored as an internal-only GUC; gucs don't support unsigned ints.

src/backend/utils/misc/guc.c
```diff
@@ -694,6 +694,12 @@ int			ssl_renegotiation_limit;
 int			huge_pages;
 int			huge_page_size;
 
+/*
+ * OIDs are stored as uint32, so we will add INT_MIN to match the range.
+ */
+int			temp_oid_buffer_start;
+int			temp_oid_buffer_size;
+
```
surendravishnoi, line 702: Instead of using these GUCs directly, add two methods to get values which will convert int to uint32 and return

src/backend/access/transam/varsup.c
surendravishnoi: We can keep all variables as unit32. If GUC is int, we just convert it to uint32 when accessing its value. That way tempoid logic is not polluted with INT->UINT conversion.

surendravishnoi: I will hold further review till we do this change as it would simply logic and review

src/backend/catalog/catalog.c
lejaokri: Can this be done?

src/backend/access/transam/varsup.c
```diff
@@ -514,6 +520,76 @@ ForceTransactionIdLimitUpdate(void)
...
+				tempOidStart = FirstNormalObjectId;
+
+			temp_oid_buffer_start = OID_TO_BUFFER_START(tempOidStart);
+			ShmemVariableCache->tempOidStart = tempOidStart;
+			ShmemVariableCache->tempOidBufferSize = temp_oid_buffer_size;
+
+			nextTempOid = (Oid) tempOidStart;
+
+			/* Skip nextOid ahead to end of range here as well.  */
+			ShmemVariableCache->nextOid = (Oid) (tempOidStart + temp_oid_buffer_size);
```
surendravishnoi: we are always consuming temp_oid_buffer_size worth of temp ids here. What if ShmemVariableCache->oidCount is < temp_oid_buffer_size? is oidCount not the upper limit for one assignment?

src/backend/access/transam/varsup.c
```diff
@@ -514,6 +520,76 @@ ForceTransactionIdLimitUpdate(void)
...
+				ShmemVariableCache->oidCount = 0;
+			else
+				ShmemVariableCache->oidCount -= temp_oid_buffer_size;
+
+			/*
+			 * If ShmemVariableCache->nextOid is below FirstNormalObjectId then we can start at FirstNormalObjectId here and
+			 * GetNewObjectId will return the right value on the next call.  
+			 */
+			if (tempOidStart < FirstNormalObjectId)
+				tempOidStart = FirstNormalObjectId;
```
surendravishnoi: How does oidCount reconcile here? we are skipping OIDs from next up to first normal oid. Shouldn't we deduct those from oidCount?

src/backend/access/transam/varsup.c
```diff
@@ -514,6 +520,76 @@ ForceTransactionIdLimitUpdate(void)
...
+			/* Skip nextOid ahead to end of range here as well.  */
+			ShmemVariableCache->nextOid = (Oid) (tempOidStart + temp_oid_buffer_size);
+		}
+		LWLockRelease(OidGenLock);
+	}
+
+	/*
+	 * Check for wraparound of the temp OID buffer.
+	 */
+	if (nextTempOid >= (Oid) (BUFFER_START_TO_OID + temp_oid_buffer_size)
```
surendravishnoi: we should have test cases for both conditions
1) When we exhaust current allocation
2) When we wraparound current allocation

src/backend/access/transam/varsup.c
```diff
@@ -514,6 +520,76 @@ ForceTransactionIdLimitUpdate(void)
...
 
+Oid
+GetNewTempObjectId(void)
+{
+	Oid			result;
+	Oid			tempOidStart;
+	static Oid 	nextTempOid = InvalidOid;
+
+	/* safety check, we should never get this far in a HS standby */
+	if (RecoveryInProgress())
```
surendravishnoi: we should have test case for this

src/backend/access/transam/varsup.c
```diff
@@ -514,6 +520,76 @@ ForceTransactionIdLimitUpdate(void)
...
+
+			/*
+			 * If ShmemVariableCache->nextOid is below FirstNormalObjectId then we can start at FirstNormalObjectId here and
+			 * GetNewObjectId will return the right value on the next call.  
+			 */
+			if (tempOidStart < FirstNormalObjectId)
+				tempOidStart = FirstNormalObjectId;
+
+			/* If the OID range would wraparound, start from beginning instead. */
+			if (tempOidStart + temp_oid_buffer_size < tempOidStart)
```
surendravishnoi: we should have test for this

src/backend/access/transam/varsup.c
surendravishnoi: also, we should have test for this which hits this condition

src/backend/access/transam/varsup.c
```diff
@@ -514,6 +520,76 @@ ForceTransactionIdLimitUpdate(void)
...
+		}
+		else
+		{
+			/* We need to pick a new start for the buffer range. */
+			tempOidStart = ShmemVariableCache->nextOid;
+
+			/*
+			 * Decrement ShmemVariableCache->oidCount to take into account the new buffer we're allocating
+			 */
+			if (ShmemVariableCache->oidCount < temp_oid_buffer_size)
```
surendravishnoi: do we have a test which exercises both if/else conditions?

src/include/utils/rel.h
surendravishnoi: comments say check for SQL dialect but actual dialect check is not there. We should add tests where we create PG side tables (regular and temp) with #name and make sure that there are not regressions. Please also check if we can create #tables on babelfish side or not (syntax leak)

src/include/access/transam.h
```diff
@@ -252,6 +252,13 @@ typedef struct VariableCacheData
 	 */
 	TransactionId oldestClogXid;	/* oldest it's safe to look up in clog */
 
+	/*
+	 * These fields are also protected by OidGenLock. For tempOidStart, Shmem will
+	 * be the source of truth, as another process may have gotten there first and 
+	 * updated the start. For tempOidBufferSize, the GUC will be the source of truth.
```
surendravishnoi: this comment is confusing, what do we mean when we say GUC will be the source of truth?

src/include/utils/rel.h
```diff
@@ -700,4 +700,12 @@ extern void RelationDecrementReferenceCount(Relation rel);
 	 strlen((relation)->rd_rel->relname.data) >= 1 && \
 	 (relation)->rd_rel->relname.data[0] == '@')
 
+/* PG Tables can begin with '#' as well, so check sql_dialect additionally */
+#define RelationIsBBFTempTable(relation) \
```
surendravishnoi: I see we are checking for three things more places
SQL dialect, is BBF temp table and is part of ENR. Can we replace is BFF temp table check with just rel persistence check? What additional cases we are covering by checking for '#'?

src/backend/catalog/catalog.c
```diff
@@ -496,6 +497,76 @@ GetNewOidWithIndex(Relation relation, Oid indexId, AttrNumber oidcolumn)
...
+	SysScanDesc scan;
+	ScanKeyData key;
+	bool		collides;
+	uint64		retries = 0;
+
+	/* Only system relations are supported */
+	Assert(IsSystemRelation(relation));
+
+	/* In bootstrap mode, we don't have any indexes to use. No temp objects in bootstrap mode. */
+	if (IsBootstrapProcessingMode())
```
surendravishnoi: is this even a valid use case? we call this function only for temp relationships, it does not make sense to return regular object id

src/backend/catalog/catalog.c
```diff
@@ -496,6 +497,76 @@ GetNewOidWithIndex(Relation relation, Oid indexId, AttrNumber oidcolumn)
...
+		return GetNewObjectId();
+
+	/*
+	 * We should never be asked to generate a new pg_type OID during
+	 * pg_upgrade; doing so would risk collisions with the OIDs it wants to
+	 * assign.  Hitting this assert means there's some path where we failed to
+	 * ensure that a type OID is determined by commands in the dump script.
+	 */
+	Assert(!IsBinaryUpgrade || RelationGetRelid(relation) != TypeRelationId);
```
surendravishnoi: is second condition in assert applicable to temp tables?

src/backend/access/transam/varsup.c
```diff
@@ -514,6 +520,76 @@ ForceTransactionIdLimitUpdate(void)
...
+{
+	Oid			result;
+	Oid			tempOidStart;
+	static Oid 	nextTempOid = InvalidOid;
+
+	/* safety check, we should never get this far in a HS standby */
+	if (RecoveryInProgress())
+		elog(ERROR, "cannot assign OIDs during recovery");
+
+	if (!OidIsValid((Oid) temp_oid_buffer_start)) /* InvalidOid means it needs assigning */
```
surendravishnoi: I think we should use different logic to check for un assigned temp oid, may be reserver INT_MIN as unassigned value?

src/backend/catalog/catalog.c
```diff
@@ -496,6 +497,76 @@ GetNewOidWithIndex(Relation relation, Oid indexId, AttrNumber oidcolumn)
...
+		/*
+		 * Log that we iterate more than GETNEWOID_LOG_THRESHOLD but have not
+		 * yet found OID unused in the relation. Then repeat logging with
+		 * exponentially increasing intervals until we iterate more than
+		 * GETNEWOID_LOG_MAX_INTERVAL. Finally repeat logging every
+		 * GETNEWOID_LOG_MAX_INTERVAL unless an unused OID is found. This
+		 * logic is necessary not to fill up the server log with the similar
+		 * messages.
+		 */
+		if (OidIsValid((Oid) temp_oid_buffer_start) && retries >= temp_oid_buffer_size)
```
surendravishnoi: what does it mean to have an invalid temp_oid_buffer_start here? is there a test case where we do not hit both if/else

src/backend/catalog/catalog.c
surendravishnoi: do we have a test case for this condition?

src/backend/catalog/catalog.c
```diff
@@ -558,10 +630,127 @@ GetNewRelFileNode(Oid reltablespace, Relation pg_class, char relpersistence)
...
+			 * Here we have a little bit of a dilemma: if errno is something
+			 * other than ENOENT, should we declare a collision and loop? In
+			 * practice it seems best to go ahead regardless of the errno.  If
+			 * there is a colliding file we will get an smgr failure when we
+			 * attempt to create the new relation file.
+			 */
+			collides = false;
+		}
+
+		if (OidIsValid((Oid) temp_oid_buffer_start) && tries > temp_oid_buffer_size)
```
surendravishnoi: we are getting oids for both temp and regular object. tries count will increase even when there is collision for regular object ids. How is this condition sufficient to stop mix-up between regular table and temp table retries?

src/backend/catalog/index.c
```diff
@@ -950,8 +950,16 @@ index_create(Relation heapRelation,
 		}
 		else
 		{
-			indexRelationId =
-				GetNewRelFileNode(tableSpaceId, pg_class, relpersistence);
+			if (is_enr)
+			{
+				/* Index OIDs must be kept in normal OID range. */
+				indexRelationId = GetNewPermanentRelFileNode(tableSpaceId, pg_class, relpersistence);
```
surendravishnoi: rather than adding a duplicate function, would it be easier to add a parameter to GetNewRelFileNode to indicate the same?

src/include/utils/guc.h
```diff
@@ -279,6 +279,9 @@ extern PGDLLIMPORT int temp_file_limit;
 
 extern PGDLLIMPORT int num_temp_buffers;
 
+extern PGDLLIMPORT int temp_oid_buffer_start;
+extern PGDLLIMPORT int temp_oid_buffer_size;
```
surendravishnoi, line 283: there should be a way to disable the feature, based on buffer size may be

src/backend/access/transam/varsup.c
```diff
@@ -514,6 +520,76 @@ ForceTransactionIdLimitUpdate(void)
 	return false;
 }
 
+Oid
+GetNewTempObjectId(void)
```
surendravishnoi: we should add functionality to extension via hooks wherever possible, this will reduce the chances of regressing engine code

src/backend/catalog/index.c
time-change: I favored this method over adding a new parameter to avoid refactoring many parts of engine code with the new function

src/backend/access/transam/varsup.c
time-change: oidCount is decremented by the if statement above. In the scenario you're describing, oidCount will be set to 0. The next time GetNewObjectId is called, it will handle oidCount = 0. 

So the current behavior will have the buffer be a fixed 65k size - if there's fewer than that many OIDs actually available (say the last 10k OIDs in the buffer), then there is a chance we place the new OID buffer (since it will wraparound) in a zone of completely used OIDs and we may go into general OID wraparound behavior about 10k OIDs earlier than is expected. In my opinion this is an extreme edge case, since an application that would consume all the OIDs in the buffer likely won't be too affected by this case (which would only happen if the entire database consume the entire OID buffer before ever creating a temp table in Babelfish) - they would have to clean up OIDs eventually. 

The potential fixes for this edge case I think all raise questions of their own - if we limit the temp_oid_buffer_size to the size of the remaining ShmemVariableCache->oidCount, then we run the risk of having an extremely small buffer size with no way for the customer to fix it. We could disable the feature completely in the case where oidCount is less than the buffer size. However both of these solutions would still require some form of operator intervention.

src/backend/catalog/catalog.c
time-change: It's just a sanity check to ensure that the temp_oid_buffer_start is initialized - but should be unnecessary as we are always calling GetNewTempObjectId before this check.

src/include/access/transam.h
time-change: This comment is left over from when temp_oid_buffer_size was user-adjustable. I will delete this comment

src/backend/access/transam/varsup.c
time-change: One way we could test this is to take off the limit on temp_oid_buffer_size. Then we can have it be essentially OID_MAX which would trigger this. The downside is that leaves the door open for potentially breaking normal OID usage if the temp_oid_buffer_size is tuned too large.

Moreover, how would we be able to verify the correctness?

I could test this manually thru GDB/engine code change. Automated is a different story

src/backend/access/transam/varsup.c
time-change: The Oids below FirstNormalObjectId are reserved, so they can't be used as part of the normal oid buffer; oidCount accounts for this. If we hit this statement, we must be in wraparound (since tempOidStart was grabbed from ShmemVariableCache->nextOid). 

Similar to the above - this scenario is theoretically testable manually through gdb/code changes. Automating this is difficult because we would have to consume the entire OID buffer to simulate wraparound.

src/backend/access/transam/varsup.c
time-change: As above, can test manually, automated testing is difficult due to requiring wraparound in this scenario.

src/backend/catalog/catalog.c
time-change: Done

src/backend/access/transam/varsup.c
time-change: Already implemented

src/backend/access/transam/varsup.c
time-change: A test case for this would require setting up a hot standby server. And as in the comment, we shouldn't reach this codepath even in that case. I added notes in the quip document.

src/backend/access/transam/varsup.c
```diff
@@ -13,6 +13,8 @@
 
 #include "postgres.h"
 
+#include <unistd.h>
```
surendravishnoi: why these changes?

src/backend/utils/misc/guc.c
```diff
@@ -694,6 +694,12 @@ int			ssl_renegotiation_limit;
 int			huge_pages;
 int			huge_page_size;
 
+/*
+ * OIDs are stored as uint32, so we will add INT_MIN to match the range.
```
surendravishnoi, line 698: only keep what is required on engine side

src/backend/catalog/index.c
```diff
@@ -950,8 +950,12 @@ index_create(Relation heapRelation,
 		}
 		else
 		{
-			indexRelationId =
-				GetNewRelFileNode(tableSpaceId, pg_class, relpersistence);
+			/* Index OIDs must be kept in normal OID range. */
```
lejaokri: nit:
better to comment why index needs to use normal oid range

src/backend/commands/cluster.c
```diff
@@ -734,8 +734,14 @@ make_new_heap(Oid OIDOldHeap, Oid NewTableSpace, Oid NewAccessMethod,
 	 * a shared rel.  However, we do make the new heap mapped if the source is
 	 * mapped.  This simplifies swap_relation_files, and is absolutely
 	 * necessary for rebuilding pg_class, for reasons explained there.
+	 *
+	 * We also must ensure that these temp tables are properly named in TSQL
+	 * so that the metadata is properly cleaned up after in this function.
 	 */
-	snprintf(NewHeapName, sizeof(NewHeapName), "pg_temp_%u", OIDOldHeap);
+	if (sql_dialect == SQL_DIALECT_TSQL && relpersistence == RELPERSISTENCE_TEMP && get_ENR_withoid(currentQueryEnv, OIDOldHeap, ENR_TSQL_TEMP))
+		snprintf(NewHeapName, sizeof(NewHeapName), "#pg_temp_%u", OIDOldHeap);
```
lejaokri, line 742: I dont think we're getting here for table variables so I think it should be fine to assume we're temp table and automatically prepend '#'. Normally I would suggest to add an `Assert(!RelationIsBBFTableVariable(OldHeap);`.

src/backend/commands/cluster.c
```diff
@@ -1603,6 +1609,8 @@ finish_heap_swap(Oid OIDOldHeap, Oid OIDNewHeap,
 			/* rename the toast table ... */
 			if (sql_dialect == SQL_DIALECT_TSQL && RelationIsBBFTableVariable(newrel))
 				pg_toast_prefix = "@pg_toast";
+			else if (sql_dialect == SQL_DIALECT_TSQL && newrel->rd_rel->relpersistence == RELPERSISTENCE_TEMP && get_ENR_withoid(currentQueryEnv, newrel->rd_id, ENR_TSQL_TEMP))
```
lejaokri, line 1612: same as above

src/include/access/transam.h
```diff
@@ -252,6 +252,13 @@ typedef struct VariableCacheData
 	 */
 	TransactionId oldestClogXid;	/* oldest it's safe to look up in clog */
 
+	/*
+	 * These fields are also protected by OidGenLock. For tempOidStart, Shmem will
+	 * be the source of truth, as another process may have gotten there first and 
+	 * updated the start.
+	 */
```
lejaokri, line 259: just fyi to keep an eye once this is merged to internal gitfarm as this struct could be different internally

src/backend/catalog/catalog.c
```diff
@@ -558,10 +569,23 @@ GetNewRelFileNode(Oid reltablespace, Relation pg_class, char relpersistence)
 
 		/* Generate the OID */
 		if (pg_class)
-			rnode.node.relNode = GetNewOidWithIndex(pg_class, ClassOidIndexId,
+		{
+			/* Temp tables use temp OID logic */
+			if (USE_BBF_OID_BUFFER)
```
surendravishnoi: we can do it once outside loop and use the result

src/backend/catalog/index.c
```diff
@@ -950,8 +950,12 @@ index_create(Relation heapRelation,
...
+			if (is_enr && GetNewPermanentRelFileNode_hook)
+				indexRelationId = GetNewPermanentRelFileNode_hook(tableSpaceId, pg_class, relpersistence);
```
surendravishnoi: this approach has following issue
In future if there is any change in GetNewRelFileNode, it will not come into GetNewPermannent... function and subtle bugs can creep in. I think it would be better to modify function with one additional argument. We need to change at 3-4 other places only. One benefit of this is also that in future if someone adds a new call to GetNewRelFileNode, they need to pass a new argument and it will be detected at merge time.
Now for function GetNewRelFileNode, we have another option of keeping it unchanged with just a hook call at beginning into extension with modified logic. This will make engine side changes to bare minimum. However, it also has disadvantage that a change on engine side might get missed on corresponding hook on extension side.

src/include/utils/rel.h
surendravishnoi: have we added this testing?

src/backend/utils/misc/guc.c
time-change: temp_oid_buffer_size is referenced in catalog.c

lejaokri (dismissed):
I dont have any more comment

src/backend/utils/activity/pgstat_relation.c
```diff
@@ -26,6 +26,9 @@
 #include "utils/rel.h"
 #include "utils/timestamp.h"
 
+#include "parser/parser.h"
+#include "utils/queryenvironment.h"
```
surendravishnoi, line 30: do we need this header?

src/backend/access/transam/varsup.c
```diff
@@ -27,9 +27,6 @@
 #include "utils/syscache.h"
 
 
-/* Number of OIDs to prefetch (preallocate) per XLOG write */
```
surendravishnoi: we should not move the define, rather we should add a wrapper function to access it

src/backend/utils/activity/pgstat_relation.c
time-change: yes. sql_dialect and RELPERSISTENCE_TEMP below

lejaokri (approved):

time-change:
Closing, this is merged in 16.x only for now.

PR #234, opened 2023-10-09
Create DB Performance improvements

tanscorpio7:
### Description

In Multi DB mode, as the number of databases increases, so does the time to create the next new DB.
This is because we create three internal roles for each new DB and internally when run the DB subcommands, multiple calls to roles_is_member_of("sysadmin") is made. Now that output of this list contains all the three roles of every db created. This is the major reason for the perfomance degradation of CREATE DB command.

We fix this in three different places.

1. getAvailDbid - this functions makes a call to nextval function, which by default checks for current user's permission and makes a call to roles_is_member_of. Instead we could call the **nextval_internal** which is the same function but with the additional option of check permissions flag which we will set to false. To double check we can just ensure that the current user is "sysadmin" when getAvailDbid is called. (Currently we only call this when user is sysadmin)

2. Set temporary user when creating schema - when we create the dbo and guest schema for the new database, the create schema function fetches all the roles that current role is member of (recursively) to check if if current role can actually become the target schema owner role. To bypass this we can **assume the newdb_dbo role** when creating these schemas. In this case all the roles that newdb_dbo is member of will be fetched, but this list is much smaller than sysadmin.

3. Select best grantor - Select best grantor first fetches the roles_list that sysadmin is member of and then start checking for permissions. But sysadmin is always the first to be checked. That is sysadmin is always top of the roles_list. 
We can add a quick check to this. That is, first check if current role is sysadmin and can it give us all the permission needed. If yes, simply return. Note** This does not change any behaviour since this will anyway be done in the first loop after fetching roles_list. We are instead running the first loop before fetching the whole list.

4. Set newdb_dbo user when creating sysdatabases view in new db

| Create DB              | After | Before |
| :----------------: | :------: | :----: |
| DBs10        |   2ms  | 5ms |
| DBs500     |   3ms   | 163ms |
| DBs1000    |  4ms   | 740ms |
| DBs3000   |  9ms   | NA(~18secs) |

### Issues Resolved

[BABEL-3869](https://jira.rds.a2z.com/browse/BABEL-3869)

### DIFF
--- a/src/backend/utils/adt/acl.c
+++ b/src/backend/utils/adt/acl.c
@@ -33,7 +33,6 @@
 #include "funcapi.h"
 #include "lib/qunique.h"
 #include "miscadmin.h"
-#include "parser/parser.h"
 #include "utils/acl.h"
 #include "utils/array.h"
 #include "utils/builtins.h"
--- a/src/backend/utils/adt/acl.c
+++ b/src/backend/utils/adt/acl.c
@@ -5120,14 +5120,8 @@ select_best_grantor(Oid roleId, AclMode privileges,
 		return;
 	}
 
-	/*
-	 * Otherwise we have to do a careful search to see if roleId has the
-	 * privileges of any suitable role.  Note: we can hang onto the result of
-	 * roles_is_member_of() throughout this loop, because aclmask_direct()
-	 * doesn't query any role memberships.
-	 */
-
-	if(roleId == get_role_oid("sysadmin", true))
+	if(sql_dialect == SQL_DIALECT_TSQL &&
+		roleId == get_role_oid("sysadmin", true))
 	{
 		AclMode		sysadmin_privs;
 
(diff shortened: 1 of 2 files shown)


src/backend/utils/adt/acl.c
```diff
@@ -5126,6 +5126,22 @@ select_best_grantor(Oid roleId, AclMode privileges,
 	 * roles_is_member_of() throughout this loop, because aclmask_direct()
 	 * doesn't query any role memberships.
 	 */
+
+	if(roleId == get_role_oid("sysadmin", true))
```
shalinilohia50: We can add a check for TSQL dialect here.

Deepesh125: I agree here but should use hook here

shalinilohia50: +1

forestkeeper (approved):

Deepesh125 (approved):

PR #258, opened 2023-11-18
Added Hook for Geospatial function handling

Anikait143:
### Description

- Added PostParseFunctionRefHook to support TSQL syntax for geospatial functions.

 
### Issues Resolved

Task: BABEL- 4370
 
### Check List

### DIFF
--- a/src/backend/parser/parse_expr.c
+++ b/src/backend/parser/parse_expr.c
@@ -1540,12 +1540,12 @@ transformFuncCall(ParseState *pstate, FuncCall *fn)
 	}
 
 	/*
-	 * Now we give the Post Func ref hook to check whether the function matches any of the Geospatial functions and handle them separately
+	 * Now we give the Pre Func ref hook to check whether the function matches any of the Geospatial functions and handle them separately
 	 * STAsText(), STAsBinary(), STDistance()
 	 */
-	if (pstate->p_post_funcref_hook != NULL)
+	if (pstate->p_pre_funcref_hook != NULL)
 	{
-		targs = pstate->p_post_funcref_hook(pstate, fn, targs);
+		targs = pstate->p_pre_funcref_hook(pstate, fn, targs);
 	}
 
 	/* ... and hand off to ParseFuncOrColumn */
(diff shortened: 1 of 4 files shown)


src/include/parser/parse_node.h
```diff
@@ -89,6 +89,7 @@ typedef enum ParseExprKind
  */
 typedef Node *(*PreParseColumnRefHook) (ParseState *pstate, ColumnRef *cref);
 typedef Node *(*PostParseColumnRefHook) (ParseState *pstate, ColumnRef *cref, Node *var);
+typedef List *(*PostParseFunctionRefHook) (ParseState *pstate, FuncCall *fn, List *fargs);
```
time-change: i wonder if the naming is correct, `PostParseFunctionRefHook`. since we aren't technically calling the hook post func processing, it's before the call to `ParseFuncOrColumn`.

src/backend/parser/parse_expr.c
```diff
@@ -1539,6 +1539,15 @@ transformFuncCall(ParseState *pstate, FuncCall *fn)
...
 	}
 
+	/*
+	 * Now we give the Post Func ref hook to check whether the function matches any of the Geospatial functions and handle them separately
+	 * STAsText(), STAsBinary(), STDistance()
+	 */
+	if (pstate->p_post_funcref_hook != NULL)
+	{
+		targs = pstate->p_post_funcref_hook(pstate, fn, targs);
+	}
```
kuntalghosh: I think this hook will be useful. Please create a JIRA for moving the checksum code under this hook and assign it to the owner.

src/include/parser/parse_node.h
Anikait143: It is default convention we have also followed it in transformColumnRef function

src/backend/parser/parse_expr.c
Anikait143: I will do that once this is merged

KushaalShroff: If we do the change in ANTLR then we wont require a hook

src/include/parser/parse_node.h
rishabhtanwar29: IMO, this should actually be `PreParseFunctionRefHook` since this steps precedes the actual function parsing logic.

src/backend/parser/parse_expr.c
Anikait143: Actually we require catalogue lookup for these queries so we can't do these changes in parser side that's why we chose the analyser phase for these handlings.

src/include/parser/parse_node.h
Anikait143: Done

PR #260, opened 2023-11-21
Fixed error code difference occurring due to parallel worker enforced

Deepesh125:
### Description

When any error raised from parallel worker then Postgres by default does not share message_id (part of edata) to leader worker. Babelfish's error mapping framework relies on message_id to map to correct T-SQL error code. But since it is no more available to leader worker, T-SQL error code would not be found and default error code will be used. This will also affect the transactional behaviour of given error. 

In order to fix this issue, we need to store if given worker is spawned in the context of Babelfish or not to share error
message_id. Hence, this changes made changes in two data structure namely Port and FixedParallelState to store context
about Babelfish
 
### Issues Resolved
BABEL-4539

### Check List
Existing JDBC test cases should be sufficient to test these changes. 
Also, regression test suite provided by vanilla Postgres runs fine with these changes.

### DIFF
--- a/src/backend/access/transam/parallel.c
+++ b/src/backend/access/transam/parallel.c
@@ -96,14 +96,14 @@ typedef struct FixedParallelState
 	TimestampTz stmt_ts;
 	SerializableXactHandle serializable_xact_handle;
 
+	/* Track if parallel worker is spawned in the context of Babelfish */
+	bool babelfish_context;
+
 	/* Mutex protects remaining fields. */
 	slock_t		mutex;
 
 	/* Maximum XactLastRecEnd of any worker. */
 	XLogRecPtr	last_xlog_end;
-
-	/* Track if parallel worker is spawned in the context of Babelfish */
-	bool babelfish_context;
 } FixedParallelState;
 
 /*
@@ -335,9 +335,9 @@ InitializeParallelDSM(ParallelContext *pcxt)
 	fps->xact_ts = GetCurrentTransactionStartTimestamp();
 	fps->stmt_ts = GetCurrentStatementStartTimestamp();
 	fps->serializable_xact_handle = ShareSerializableXact();
+	fps->babelfish_context = MyProcPort->is_tds_conn;
 	SpinLockInit(&fps->mutex);
 	fps->last_xlog_end = 0;
-	fps->babelfish_context = MyProcPort->is_tds_conn;
 	shm_toc_insert(pcxt->toc, PARALLEL_KEY_FIXED, fps);
 
 	/* We can skip the rest of this if we're not budgeting for any workers. */
--- a/src/backend/access/transam/parallel.c
+++ b/src/backend/access/transam/parallel.c
@@ -31,7 +31,6 @@
 #include "libpq/pqmq.h"
 #include "miscadmin.h"
 #include "optimizer/optimizer.h"
-#include "parser/parser.h"
 #include "pgstat.h"
 #include "storage/ipc.h"
 #include "storage/predicate.h"
--- a/src/backend/access/transam/parallel.c
+++ b/src/backend/access/transam/parallel.c
@@ -102,6 +102,9 @@ typedef struct FixedParallelState
 
 	/* Maximum XactLastRecEnd of any worker. */
 	XLogRecPtr	last_xlog_end;
+
+	/* Track if parallel worker is spawned in the context of Babelfish */
+	bool babelfish_context;
 } FixedParallelState;
 
 /*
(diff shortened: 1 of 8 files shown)


src/backend/libpq/pqmq.c
```diff
@@ -310,6 +311,16 @@ pq_parse_errornotice(StringInfo msg, ErrorData *edata)
...
 				edata->funcname = pstrdup(value);
 				break;
+			case PG_DIAG_MESSAGE_ID:
+				if (sql_dialect == SQL_DIALECT_TSQL)
+				{
+					edata->message_id = pstrdup(value);
+				}
+				else
+				{
+					elog(ERROR, "Unexpected message_id found");
```
surendravishnoi: why are we giving error for APG here? also does this code applies to clients as well? they are not aware of SQL dialect

src/backend/utils/error/elog.c
```diff
@@ -1703,6 +1704,8 @@ ThrowErrorData(ErrorData *edata)
 	if (edata->backtrace)
 		newedata->backtrace = pstrdup(edata->backtrace);
 	/* assume message_id is not available */
+	if (edata->message_id)
```
surendravishnoi: why this change? message id is global at-least for given postgres worker

src/backend/utils/error/elog.c
```diff
@@ -3250,6 +3253,15 @@ send_message_to_frontend(ErrorData *edata)
 			err_sendstring(&msgbuf, edata->funcname);
 		}
 
+		if (sql_dialect == SQL_DIALECT_TSQL)
+		{
+			if (edata->message_id)
```
surendravishnoi, line 3264: how do we make sure that all clients will be able to handle additional message id in error messgage

src/backend/libpq/pqmq.c
KushaalShroff: For PG anyway we were throwing `elog(ERROR, "unrecognized error field code: %d", (int) code);` before

Deepesh125: which clients are you referring to? Client here would be leader node (not psql that you might be referring here)

src/backend/utils/error/elog.c
Deepesh125: message_id is not global, its part of edata

src/backend/utils/error/elog.c
Deepesh125: I think I will add another check (IsParallelWorker()) here to ensure that only parallel worker sends this message_id and client (leader node in this case) will be able to handle it through dialect.

src/backend/utils/error/elog.c
surendravishnoi: message ids are static strings, so there is one copy at process level. that is why we do not copy them, assigning pointer is sufficient. What is reasoning behind this change?

src/backend/utils/error/elog.c
surendravishnoi: are we sure that parallel workers have consistent mapping with respect to dialect with leader node? what if error happens inside a procedure executed with PG dialect or vice versa and leader has different dialect?

src/backend/libpq/pqmq.c
surendravishnoi: we should validate here as well that this is indeed in context of parallel worker (I am not sure if that is even possible). Also, code treats message id as static strings. If we are using str dup, we need to make sure that assumption is not broken when we simply assign pointers.

src/backend/utils/error/elog.c
Deepesh125: First copy is happening inside temporary context (which will be resetted at the end of handling parallel message) and second copy is happening inside error context which would be preserved at process level and will be used everywhere for error manipulation. So there is no harm in keeping this.

src/backend/utils/error/elog.c
Deepesh125: As discussed offline, we can keep this message_id irrespective of dialect.

src/include/libpq/libpq-be.h
```diff
@@ -250,6 +250,7 @@ typedef struct Port
 	SSL		   *ssl;
 	X509	   *peer;
 #endif
+	bool		is_tds_conn;
```
surendravishnoi, line 253: this information is available on TSQL side, we should add hook rather than adding a new member here.

src/backend/access/transam/parallel.c
```diff
@@ -101,6 +101,9 @@ typedef struct FixedParallelState
 
 	/* Maximum XactLastRecEnd of any worker. */
 	XLogRecPtr	last_xlog_end;
+
+	/* Track if parallel worker is spawned in the context of Babelfish */
+	bool babelfish_context;
```
surendravishnoi: if this is not in the context of mutex, move it before mutex

Deepesh125: Thats good point, updated accordingly. Thanks

src/include/libpq/libpq-be.h
Deepesh125: We rather keep it as it is for some other reasons.

surendravishnoi (approved):

PR #302, opened 2024-02-04
Fixed crash when trying to drop ENR relations during subtransaction.

lejaokri:
### Description

Fixed crash when trying to drop ENR relations during subtransaction.

Unable to drop ENR tables in subtransaction because invalidation
messages was never accepted failing the check in PrepareInvalidationState().

This is because ENR relations skip GetCurrentCommandId(true) hence
making CommandCounterIncrement() practically a no-op.

Also fix a Day 1 issue when caller of pltsql_clean_table_variables()
skips some important error handling when
pltsql_clean_table_variables
fails.

Task: BABEL-4737
 
### Issues Resolved

### Check List

### DIFF
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -27,6 +27,7 @@
 #include "access/tupdesc.h"
 #include "access/htup_details.h"
 #include "access/relscan.h"        /* SysScan related */
+#include "access/xact.h"           /* GetCurrentCommandId */
 #include "catalog/catalog.h"
 #include "catalog/dependency.h"
 #include "catalog/indexing.h"
@@ -867,8 +868,13 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
 			default:
 				break;
 		}
+
 		/* add/drop/update is possible. Do it now. */
-		if (ret) {
+		if (ret)
+		{
+			/* Tell CommandCounterIncrement() we are about to create some inval messages */
+			GetCurrentCommandId(true);
+
 			Assert(queryEnv->memctx);
 			oldcxt = MemoryContextSwitchTo(queryEnv->memctx);
 			switch (op) {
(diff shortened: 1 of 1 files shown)


src/backend/utils/misc/queryenvironment.c
```diff
@@ -878,14 +884,23 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
 					CacheInvalidateHeapTuple(catalog_rel, newtup, NULL);
 					break;
 				case ENR_OP_UPDATE:
+					/*
```
surendravishnoi, line 887: this is a known issue with temp tables/table variables. Why do we want to invalidate cache in the first place? Temp tables/variables are session local and should not impact any one else. So, we should not be even doing this unless we have entries in the catalog and not in ENR.

lejaokri: ENR still uses the syscache. We need to invalidate our **own** syscache, otherwise we would keep finding the tuple in the cache even after dropping/altering the table.

lejaokri: This also happens if I remove `CacheInvalidateHeapTuple` after creation.

```
[clouddesk3] % sqlcmd -S localhost -U jdbc_user -P 12345678
1> create table #temp(c1 int)
2> insert into #temp values(1)
3> go
relation "#temp" does not exist
```

surendravishnoi: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/134, please take a look how we can avoid shared memory invalidation. Also, perf test data is limited as we should have tried 2000 to 4000 connections

surendravishnoi (approved):

src/backend/utils/misc/queryenvironment.c
lejaokri: just for the record, #134 is a follow on for this

PR #306, opened 2024-02-08
Add pg_attrdef to ENR for BBF table variables and temp tables

Jakeowen1:
### Description

Table Variables and Temp Tables can have identity columns which
adds entries to pg_attrdef. If the table is in ENR, then pg_attrdef
entry should also be in ENR otherwise we would have catalog
inconsistency.
 
### Issues Resolved

Task: BABEL-4752
 
### Check List

### DIFF
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -758,6 +758,9 @@ ENRAddUncommittedTupleData(EphemeralNamedRelation enr, Oid catoid, ENRTupleOpera
 		case SequenceRelationId:
 			list_ptr = &enr->md.uncommitted_cattups[ENR_CATTUP_SEQUENCE];
 			break;
+		case AttrDefaultRelationId:
+			list_ptr = &enr->md.uncommitted_cattups[ENR_CATTUP_ATTR_DEF_REL];
+			break;
 		default:
 			/* Shouldn't reach here */
 			elog(ERROR, "Did not find matching ENR catalog entry for catoid %d", catoid);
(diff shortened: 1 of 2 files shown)


src/backend/utils/misc/queryenvironment.c
```diff
@@ -588,6 +590,15 @@ bool ENRgetSystableScan(Relation rel, Oid indexId, int nkeys, ScanKey key, List
 					}
 				}
 			}
+			else if (reloid == AttrDefaultRelationId) {
+				if (indexId == AttrDefaultOidIndexId) {
```
lejaokri: I think `AttrDefaultRelationId` has two index. We need to support both.
```
53 DECLARE_UNIQUE_INDEX(pg_attrdef_adrelid_adnum_index, 2656, AttrDefaultIndexId, on pg_attrdef using btree(adrelid oid_ops, adnum int2_ops));
 54 DECLARE_UNIQUE_INDEX_PKEY(pg_attrdef_oid_index, 2657, AttrDefaultOidIndexId, on pg_attrdef using btree(oid oid_ops));
```

src/backend/utils/misc/queryenvironment.c
```diff
@@ -37,6 +37,7 @@
 #include "catalog/pg_type.h"
 #include "catalog/pg_depend.h"
 #include "catalog/pg_sequence.h"
+#include "catalog/pg_attrdef.h"
```
surendravishnoi, line 41: We should think of adding some sanity checks in ENR functions. Specifically, if we do not find an ENR entry, for given context and catalog, is it expected or we should assert/raise and error. We need not to do that as part of this change. Otherwise, looks good. I assume we will add sufficient test to take care of
- DDLs, DMLs and queries which will need to access/update the ENR catalog to work properly for default value columns.

time-change: marking this as resolved, as we noted this in a JIRA for followup

time-change:
Now that ENR Rollback is fixed, this change should be unblocked.

lejaokri (approved):

2jungkook (approved):

PR #310, opened 2024-02-21
index created for UNIQUE table/column constraint in pltsql should be NULLS FIRST

tanscorpio7:
### Description

pltsql always defaults to NULLS FIRST ordering when ASC for index. One case which was not covered yet was UNIQUE constraint in CREATE TABLE command, which still creates an index with SORTBY_NULLS_DEFAULT. (PG default is NULLS LAST in ASC)
This UNIQUE constraint could be a column constraint or table constraint and PG does not support NULLS ORDERING for UNIQUE constraint in CREATE TABLE.
`CREATE TABLE name (col1 type UNIQUE)`    -- column constraint
or
`CREATE TABLE name (col1 type1, col2 type2, UNIQUE (col1))`     -- table constraint

To fix this we change the NULLS ORDERING for unique index created from table/column constraints.
The hook is placed in transformIndexConstraint which is responsible for converting the constraint node to create index node for create stmts.
We modify the NULLS ORDERING value here.

Also only index which have amcanorder set to true can accept ordering. Right now only b-tree are supported as unique index and b-tree does support ordering.
 
### Issues Resolved

[BABEL-3571]
 
#### Sign Off

### Check List

### DIFF
--- a/src/backend/parser/parse_utilcmd.c
+++ b/src/backend/parser/parse_utilcmd.c
@@ -2593,7 +2593,7 @@ transformIndexConstraint(Constraint *constraint, CreateStmtContext *cxt)
 				iparam->ordering = i->ordering;
 			}
 
-			if (pltsql_unique_constraint_nulls_ordering_hook)
+			if (sql_dialect != SQL_DIALECT_TSQL && pltsql_unique_constraint_nulls_ordering_hook)
 			{
 				iparam->nulls_ordering = (* pltsql_unique_constraint_nulls_ordering_hook) (constraint->contype, iparam->ordering);
 			}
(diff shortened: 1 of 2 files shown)


src/backend/parser/parse_utilcmd.c
```diff
@@ -2591,6 +2592,12 @@ transformIndexConstraint(Constraint *constraint, CreateStmtContext *cxt)
 				IndexElem * i = (IndexElem *) lfirst(lc);
 				iparam->ordering = i->ordering;
 			}
+
+			if (pltsql_unique_constraint_nulls_ordering_hook)
```
forestkeeper: Should use T_SQL Dialect in here

jsudrik (approved):
Approved based on the other reviews.

PR #321, opened 2024-02-27
Added missing checks for parent queryEnv in certain ENR functions

time-change:
### Description

During the verification for the fix for BABEL-4498 (aka babelfish-for-postgresql/babelfish_extensions#2159), we found that an unexpected error would be thrown due to incorrectly value being saved to cache. This is due to conflicting visibility in certain cases with temp tables - some functions in queryenvironment.c are correctly searching through parentEnv during ENR lookup, and it was missing in others. This change corrects the functions which were not searching through parentEnv.
 

Test cases added in babelfish-for-postgresql/babelfish_extensions#2395
### Issues Resolved

Missing temp table visibility in certain checks. 
 
### Check List

### DIFF
--- a/src/backend/executor/nodeNamedtuplestorescan.c
+++ b/src/backend/executor/nodeNamedtuplestorescan.c
@@ -102,7 +102,7 @@ ExecInitNamedTuplestoreScan(NamedTuplestoreScan *node, EState *estate, int eflag
 	scanstate->ss.ps.state = estate;
 	scanstate->ss.ps.ExecProcNode = ExecNamedTuplestoreScan;
 
-	enr = get_ENR(estate->es_queryEnv, node->enrname);
+	enr = get_ENR(estate->es_queryEnv, node->enrname, true);
 	if (!enr)
 		elog(ERROR, "executor could not find named tuplestore \"%s\"",
 			 node->enrname);
--- a/src/backend/executor/nodeNamedtuplestorescan.c
+++ b/src/backend/executor/nodeNamedtuplestorescan.c
@@ -102,7 +102,7 @@ ExecInitNamedTuplestoreScan(NamedTuplestoreScan *node, EState *estate, int eflag
 	scanstate->ss.ps.state = estate;
 	scanstate->ss.ps.ExecProcNode = ExecNamedTuplestoreScan;
 
-	enr = get_ENR(estate->es_queryEnv, node->enrname, true);
+	enr = get_ENR(estate->es_queryEnv, node->enrname, false);
 	if (!enr)
 		elog(ERROR, "executor could not find named tuplestore \"%s\"",
 			 node->enrname);
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -239,12 +239,11 @@ EphemeralNamedRelation
 get_ENR_withoid(QueryEnvironment *queryEnv, Oid id, EphemeralNameRelationType type)
 {
 	ListCell   *lc;
-	QueryEnvironment *qe = queryEnv;
 
 	if (queryEnv == NULL)
 		return NULL;
 
-	foreach(lc, qe->namedRelList)
+	foreach(lc, queryEnv->namedRelList)
 	{
 		EphemeralNamedRelation enr = (EphemeralNamedRelation) lfirst(lc);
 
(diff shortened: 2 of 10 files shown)


src/backend/utils/misc/queryenvironment.c
```diff
@@ -215,19 +236,26 @@ get_ENR(QueryEnvironment *queryEnv, const char *name)
  * Same as get_ENR() but just search for relation oid
  */
 EphemeralNamedRelation
-get_ENR_withoid(QueryEnvironment *queryEnv, Oid id, EphemeralNameRelationType type)
+get_ENR_withoid(QueryEnvironment *queryEnv, Oid id, EphemeralNameRelationType type, bool search)
```
surendravishnoi: why we need search param here? Oids cannot have duplicate. When do we want to avoid finding oid in parent envs?

src/backend/utils/misc/queryenvironment.c
```diff
@@ -189,23 +196,37 @@ bool has_existing_enr_relations()
...
+ * search - true if we want to include parent queryEnvs (such as in 
+ * 			babelfish_get_enr_list, object_id).
+ * 
+ * 			Other such places (like in HeapCreateWithCatalog, SPI)
+ * 			allow duplicate ENR names in the top level queryEnv, so 
+ * 			it must be false there.
  */
 EphemeralNamedRelation
-get_ENR(QueryEnvironment *queryEnv, const char *name)
+get_ENR(QueryEnvironment *queryEnv, const char *name, bool search)
```
surendravishnoi, line 208: can we limit this special search to table variable and temp tables only? we can find that using name? also, do we have use cases where enr names could be duplicate for table variables/temp tables in same session but at different nesting level

time-change: yeah, enr names can be duplicate in same session in places like the SPI logic above, so those places the flags must be set to false to avoid registering duplicate ENRs. 

in addition, in those places the name may still begin with @ or # so we can't just add name check in the get_ENR function, it depends on the caller (if i'm correctly understanding what you mean by limiting the search to TV/TT only by using name)

src/backend/executor/nodeNamedtuplestorescan.c
```diff
@@ -102,7 +102,7 @@ ExecInitNamedTuplestoreScan(NamedTuplestoreScan *node, EState *estate, int eflag
 	scanstate->ss.ps.state = estate;
 	scanstate->ss.ps.ExecProcNode = ExecNamedTuplestoreScan;
 
-	enr = get_ENR(estate->es_queryEnv, node->enrname);
+	enr = get_ENR(estate->es_queryEnv, node->enrname, true);
```
lejaokri: This one changes behaviour. It used to check only the current level but it checks recursively now. Is this intended? This looks like it's called in generic ENR codepath outside our use case in babelfish

src/backend/utils/misc/queryenvironment.c
```diff
@@ -127,7 +127,7 @@ get_visible_ENR_metadata(QueryEnvironment *queryEnv, const char *refname)
 	if (queryEnv == NULL)
 		return NULL;
 
-	enr = get_ENR(queryEnv, refname);
+	enr = get_ENR(queryEnv, refname, true);
```
lejaokri: We're also changing behaviour here. Is this intended? asking because `get_visible_ENR_metadata` is called in a generic codepath.

src/backend/utils/misc/queryenvironment.c
surendravishnoi: can you give one concrete example where enr names are duplicate in same session?
can you explain in more detail how name check is not sufficient and why caller context is important, please provide one use case to explain it

time-change: sure. the example I have is when running a test for BABEL-4672. without setting `search = false` as above, the following will cause a server crash

```
CREATE TRIGGER tr_emp_salary_instead_insert ON emp_salary
INSTEAD OF INSERT
AS
INSERT INTO emp_salary VALUES(2, 2000);
GO

CREATE TRIGGER tr_emp_salary_after_insert ON emp_salary
AFTER INSERT
AS
INSERT INTO emp_salary VALUES(3, 3000);
GO

INSERT INTO emp_salary VALUES (1,1000);
GO
```

Reasoning here is that it uses a default name (such as 'updated' or 'inserted') for the ENR. If it's nested, then we will run into a name collision if we do a search through all parent queryEnvs. One example of a crash caused by that is in `register_ENR` by hitting `Assert(get_ENR(queryEnv, enr->md.name, true) == NULL);`. 

Even if we set the search parameter in the above `Assert` to be `false`, we will still fail assertions in other parts of the trigger code.

lejaokri (approved):
I dont have any more comment

PR #345, opened 2024-04-18
Fix behavioural differences due to return type and handle constant string literal inputs differently for T-SQL COALESCE function

basasairohan:
### Description

Behavioural differences mentioned below have been found out better PG Coalesce and T-SQL Coalesce function.
1. Consistency in return datatype i.e., for the same set of input datatypes T-SQL should return consistent return datatype as shown in below example
```
-- T-SQL
select coalesce(cast(8 as smallint), cast(1 as bit))
go
~~START~~
smallint
8
~~END~~
select coalesce(cast(1 as bit), cast(8 as smallint))
go
~~START~~
smallint
1
~~END~~

-- BBF

select coalesce(cast(8 as smallint), cast(1 as bit))
go
~~START~~
smallint
8
~~END~~
select coalesce(cast(1 as bit), cast(8 as smallint))
go
~~START~~
bit
1
~~END~~
```
2. When the input is constant string literal i.e., T-SQL should take short circuit evaluation meaning unnecessary arguments should not be evaluated
```
-- T-SQL
1> SELECT COALESCE(NULL, 1, 2, 'I am a string')
2> go
           
-----------
          1

(1 rows affected)

-- BBF
1> SELECT COALESCE(NULL, 1, 2, 'I am a string')
2> go
Msg 22, Level 16, State 1, Server BABEL, Line 1
invalid input syntax for integer: "I am a string"
```

This commit contains the following changes : 

1. Whenever the caller is T-SQL COALESCE, then call the select_common_type_hook with 'TSQL_COALESCE' context.
2. If the input is a constant string literal, first we convert the UNKNOWN node to VARCHAR node and then coerce it to the common type. This change also handles empty or white space strings according to the way T-SQL handles.

### Issues Resolved

[BABEL-726]

Extension PR : https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2507

Signed-off-by: Sai Rohan Basa [bsrohan@amazon.com](mailto:bsrohan@amazon.com)

 
### Check List

### DIFF
--- a/src/backend/parser/parse_expr.c
+++ b/src/backend/parser/parse_expr.c
@@ -2316,8 +2316,8 @@ transformCoalesceExpr(ParseState *pstate, CoalesceExpr *c)
 
 	if (sql_dialect == SQL_DIALECT_TSQL && select_common_type_hook && c->tsql_is_null)
 		newc->coalescetype = select_common_type(pstate, newargs, "ISNULL", NULL);
-	else if(sql_dialect == SQL_DIALECT_TSQL && select_common_type_hook && c->tsql_coalesce_func)
-		newc->coalescetype = select_common_type(pstate, newargs, "TSQL_COALESCE", NULL);
+	// else if(sql_dialect == SQL_DIALECT_TSQL && select_common_type_hook && c->tsql_coalesce_func)
+	// 	newc->coalescetype = select_common_type(pstate, newargs, "TSQL_COALESCE", NULL);
 	else
 		newc->coalescetype = select_common_type(pstate, newargs, "COALESCE", NULL);
 	/* coalescecollid will be set by parse_collate.c */
--- a/src/backend/parser/parse_expr.c
+++ b/src/backend/parser/parse_expr.c
@@ -2316,8 +2316,8 @@ transformCoalesceExpr(ParseState *pstate, CoalesceExpr *c)
 
 	if (sql_dialect == SQL_DIALECT_TSQL && select_common_type_hook && c->tsql_is_null)
 		newc->coalescetype = select_common_type(pstate, newargs, "ISNULL", NULL);
-	// else if(sql_dialect == SQL_DIALECT_TSQL && select_common_type_hook && c->tsql_coalesce_func)
-	// 	newc->coalescetype = select_common_type(pstate, newargs, "TSQL_COALESCE", NULL);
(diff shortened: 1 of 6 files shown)


src/backend/parser/parse_expr.c
```diff
@@ -2325,6 +2327,36 @@ transformCoalesceExpr(ParseState *pstate, CoalesceExpr *c)
...
+			{
+				if (val[i] != ' ')
+					break;
+			}
+
+			if (i != -1)
+				e = coerce_to_common_type(pstate, e,
+									 		VARCHAROID,
+									 		"COALESCE");
+		}
```
thephantomthief: Instead of adding this logic in the engine code, we should define a function in our babelfish extensions which contains this logic and just call that function.

src/backend/parser/parse_expr.c
```diff
@@ -2325,6 +2327,36 @@ transformCoalesceExpr(ParseState *pstate, CoalesceExpr *c)
...
+			if (val != NULL)
+				i = strlen(val) - 1;
+
+			/*
+			 *	Additional handling for empty or white space string literals as
+			 *	T-SQL treats an empty string literal as 0 in certain datatypes
+			 */
+			for (; i >= 0; i--)
+			{
+				if (val[i] != ' ')
```
thephantomthief: use `isspace()`

src/backend/parser/parse_expr.c
```diff
@@ -2325,6 +2328,18 @@ transformCoalesceExpr(ParseState *pstate, CoalesceExpr *c)
...
 		Node	   *newe;
+		Oid		   etype = exprType(e);
+
+		/*
+		 *	T-SQL treats constant string literals as VARCHAR. Hence,
+		 *	coercing into VARCHAR before coercing it to the common type.
+		 */
+		if (sql_dialect == SQL_DIALECT_TSQL && !c->tsql_is_null && 
+			etype == UNKNOWNOID && IsA(e, Const) &&
+			handle_constant_literals_hook)
```
thephantomthief: Let's only check for dialect and `handle_constant_literals_hook` here. The rest of the things should be moved inside the hook implementation. This will reduce overhead on a regular PG query

thephantomthief (approved):

jsudrik (approved):
Approved based on the other reviews.

PR #348, opened 2024-04-23
Add transactional support for ENR temp tables

time-change:
### Description

ENR temp tables were not sensitive to transactional behavior. We add it here by introducing a new structure which will track uncommitted ENR tuples and process them accordingly, depending on if the active transaction is committed or rolled back. 

Tests are added in babelfish-for-postgresql/babelfish_extensions#2522
 
### Issues Resolved

BABEL-4864
 
### Check List

### DIFF
--- a/src/backend/access/transam/xact.c
+++ b/src/backend/access/transam/xact.c
@@ -380,10 +380,6 @@ IsTransactionState(void)
 {
 	TransactionState s = CurrentTransactionState;
 
-	/* During ENR Rollback, we may need to call relation_open during TRANS_ABORT. */
-	if (temp_table_xact_support && ENRInRollback() && s->state == TRANS_ABORT)
-		return true;
-
 	/*
 	 * TRANS_DEFAULT and TRANS_ABORT are obviously unsafe states.  However, we
 	 * also reject the startup/shutdown states TRANS_START, TRANS_COMMIT,
@@ -636,9 +632,7 @@ AssignTransactionId(TransactionState s)
 	/* Assert that caller didn't screw up */
 	Assert(!FullTransactionIdIsValid(s->fullTransactionId));
 
-	/* In ENR Rollback, we may need to call relation_open during TRANS_ABORT. */
-	if (temp_table_xact_support && !ENRInRollback() && s->state != TRANS_ABORT)
-		Assert(s->state == TRANS_INPROGRESS);
+	Assert(s->state == TRANS_INPROGRESS);
 
 	if (AbortCurTransaction)
 	{
@@ -2265,6 +2259,9 @@ CommitTransaction(void)
 	/* Commit updates to the relation map --- do this as late as possible */
 	AtEOXact_RelationMap(true, is_parallel_worker);
 
+	if (temp_table_xact_support)
+		ENRCommitChanges(currentQueryEnv);
+
 	/*
 	 * set the current transaction state information appropriately during
 	 * commit processing
@@ -2772,6 +2769,9 @@ AbortTransaction(void)
 		elog(WARNING, "AbortTransaction while in %s state",
 			 TransStateAsString(s->state));
 	Assert(s->parent == NULL);
+
+	if (temp_table_xact_support)
+		ENRRollbackChanges(currentQueryEnv);
 
 	/*
 	 * set the current transaction state information appropriately during the
@@ -5131,6 +5131,9 @@ AbortSubTransaction(void)
 	if (s->state != TRANS_INPROGRESS)
 		elog(WARNING, "AbortSubTransaction while in %s state",
 			 TransStateAsString(s->state));
+
+	if (temp_table_xact_support)
+		ENRRollbackSubtransaction(s->subTransactionId, currentQueryEnv);
 
 	s->state = TRANS_ABORT;
 
--- a/src/backend/access/transam/xact.c
+++ b/src/backend/access/transam/xact.c
@@ -631,7 +631,6 @@ AssignTransactionId(TransactionState s)
 
 	/* Assert that caller didn't screw up */
 	Assert(!FullTransactionIdIsValid(s->fullTransactionId));
-
 	Assert(s->state == TRANS_INPROGRESS);
 
 	if (AbortCurTransaction)
--- a/src/backend/access/transam/xact.c
+++ b/src/backend/access/transam/xact.c
@@ -380,6 +380,10 @@ IsTransactionState(void)
 {
 	TransactionState s = CurrentTransactionState;
 
+	/* During ENR Rollback, we may need to call relation_open during TRANS_ABORT. */
+	if (temp_table_xact_support && ENRInRollback() && s->state == TRANS_ABORT)
+		return true;
+
 	/*
 	 * TRANS_DEFAULT and TRANS_ABORT are obviously unsafe states.  However, we
 	 * also reject the startup/shutdown states TRANS_START, TRANS_COMMIT,
@@ -631,7 +635,10 @@ AssignTransactionId(TransactionState s)
 
 	/* Assert that caller didn't screw up */
 	Assert(!FullTransactionIdIsValid(s->fullTransactionId));
-	Assert(s->state == TRANS_INPROGRESS);
+
+	/* In ENR Rollback, we may need to call relation_open during TRANS_ABORT. */
+	if (temp_table_xact_support && !ENRInRollback() && s->state != TRANS_ABORT)
+		Assert(s->state == TRANS_INPROGRESS);
 
 	if (AbortCurTransaction)
 	{
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -1099,20 +1099,7 @@ void ENRDropEntry(Oid id)
 	/* If we are dropping a committed ENR, wait until COMMIT to free it. */
 	if (temp_table_xact_support && enr->md.is_bbf_temp_table)
 	{
-		ListCell *lc;
 		enr->md.dropped_subid = GetCurrentSubTransactionId();
-		foreach(lc, currentQueryEnv->dropped_namedRelList)
-		{
-			EphemeralNamedRelation tmp_enr = (EphemeralNamedRelation) (lc);
-
-			if (strcmp(enr->md.name, tmp_enr->md.name) == 0 && tmp_enr->md.dropped_subid == enr->md.dropped_subid)
-			{
-				currentQueryEnv->dropped_namedRelList = foreach_delete_current(currentQueryEnv->dropped_namedRelList, lc);
-				pfree(tmp_enr->md.name);
-				pfree(tmp_enr);
-				break;
-			}
-		}
 		currentQueryEnv->dropped_namedRelList = lappend(currentQueryEnv->dropped_namedRelList, enr);
 	}
 	else
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -42,6 +42,7 @@
 #include "catalog/pg_index_d.h"
 #include "parser/parser.h"      /* only needed for GUC variables */
 #include "utils/inval.h"
+#include "utils/guc.h"
 #include "utils/syscache.h"
 #include "utils/queryenvironment.h"
 #include "utils/rel.h"
@@ -164,7 +165,6 @@ unregister_ENR(QueryEnvironment *queryEnv, const char *name)
 
 /*
  * Return the list of ENRs registered in the current query environment.
- * Skip over dropped ENRs in the current environment.
  */
 List *get_namedRelList()
 {
@@ -172,12 +172,7 @@ List *get_namedRelList()
 	QueryEnvironment *qe = currentQueryEnv;
 	while (qe)
 	{
-		ListCell *lc;
-		foreach(lc, qe->namedRelList)
-		{
-			EphemeralNamedRelation enr = (EphemeralNamedRelation) lfirst(lc);
-			relList = lappend(relList, enr);
-		}
+		relList = list_concat(relList, qe->namedRelList);
 		qe = qe->parentEnv;
 	}
 	return relList;
@@ -1006,7 +1001,7 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
 			switch (op) {
 				case ENR_OP_ADD:
 					newtup = heap_copytuple(tup);
-					if (enr->md.is_committed && !enr->md.is_table_variable)
+					if (enr->md.is_committed && !enr->md.is_table_variable && temp_table_xact_support)
 						ENRAddUncommittedTupleData(enr, catalog_oid, op, newtup);
 					*list_ptr = list_insert_nth(*list_ptr, insert_at, newtup);
 					CacheInvalidateHeapTuple(catalog_rel, newtup, NULL);
@@ -1022,13 +1017,13 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
 					* which is much better than crashing.
 					*/
 					oldtup = lfirst(lc);
-					if (enr->md.is_committed && !enr->md.is_table_variable)
+					if (enr->md.is_committed && !enr->md.is_table_variable && temp_table_xact_support)
 						ENRAddUncommittedTupleData(enr, catalog_oid, op, oldtup);
 					CacheInvalidateHeapTuple(catalog_rel, oldtup, tup);
 					lfirst(lc) = heap_copytuple(tup);
 					break;
 				case ENR_OP_DROP:
-					if (enr->md.is_committed && !enr->md.is_table_variable)
+					if (enr->md.is_committed && !enr->md.is_table_variable && temp_table_xact_support)
 						ENRAddUncommittedTupleData(enr, catalog_oid, op, tup);
 					if (!skip_cache_inval)
 						CacheInvalidateHeapTuple(catalog_rel, tup, NULL);
@@ -1066,7 +1061,7 @@ bool ENRdropTuple(Relation rel, HeapTuple tup)
 	/*
 	 * If we've already dropped it in this transaction but haven't committed, pretend the drop is done. 
 	 */
-	if (ENRTupleIsDropped(rel, tup))
+	if (temp_table_xact_support && ENRTupleIsDropped(rel, tup))
 		return true;
 	return _ENR_tuple_operation(rel, tup, ENR_OP_DROP, false);
 }
@@ -1094,7 +1089,7 @@ void ENRDropEntry(Oid id, QueryEnvironment *queryEnv)
 		return;
 
 	/* If we want to drop a committed ENR, then we should mark it as such and drop it on COMMIT */
-	if (enr->md.is_committed)
+	if (temp_table_xact_support && enr->md.is_committed)
 	{
 		oldcxt = MemoryContextSwitchTo(queryEnv->memctx);
 		queryEnv->namedRelList = list_delete(queryEnv->namedRelList, enr);
@@ -1514,7 +1509,7 @@ extern void ENRDropCatalogEntry(Relation catalog_relation, Oid relid)
 				htup = list_nth(*list_ptr, 0);
 				*list_ptr = list_delete_ptr(*list_ptr, htup);
 
-				if (enr->md.is_committed)
+				if (temp_table_xact_support && enr->md.is_committed)
 						ENRAddUncommittedTupleData(enr, catalog_oid, ENR_OP_DROP, htup);
 
 				CacheInvalidateHeapTuple(catalog_relation, htup, NULL);
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -52,6 +52,7 @@
 struct QueryEnvironment
 {
 	List	   *namedRelList;
+	List	   *dropped_namedRelList;
 	struct QueryEnvironment *parentEnv;
 	MemoryContext	memctx;
 };
@@ -80,6 +81,7 @@ create_queryEnv2(MemoryContext cxt, bool top_level)
 	if (top_level) {
 		queryEnv = topLevelQueryEnv;
 		queryEnv->namedRelList = NIL;
+		queryEnv->dropped_namedRelList = NIL;
 		queryEnv->parentEnv = NULL;
 		queryEnv->memctx = cxt;
 	} else {
@@ -122,7 +124,7 @@ get_visible_ENR_metadata(QueryEnvironment *queryEnv, const char *refname)
 		return NULL;
 
 	enr = get_ENR(queryEnv, refname, false);
-	if (enr && !enr->md.pending_drop)
+	if (enr)
 		return &(enr->md);
 
 	return NULL;
@@ -140,7 +142,7 @@ register_ENR(QueryEnvironment *queryEnv, EphemeralNamedRelation enr)
 	Assert(enr != NULL);
 	Assert(get_ENR(queryEnv, enr->md.name, false) == NULL);
 
-	if (enr->md.name[0] == '@')
+	if (enr->md.name[0] != '#')
 		enr->md.is_table_variable = true;
 
 	queryEnv->namedRelList = lappend(queryEnv->namedRelList, enr);
@@ -174,8 +176,7 @@ List *get_namedRelList()
 		foreach(lc, qe->namedRelList)
 		{
 			EphemeralNamedRelation enr = (EphemeralNamedRelation) lfirst(lc);
-			if (!enr->md.pending_drop)
-				relList = lappend(relList, enr);
+			relList = lappend(relList, enr);
 		}
 		qe = qe->parentEnv;
 	}
@@ -225,7 +226,7 @@ get_ENR(QueryEnvironment *queryEnv, const char *name, bool search)
 		{
 			EphemeralNamedRelation enr = (EphemeralNamedRelation) lfirst(lc);
 
-			if (strcmp(enr->md.name, name) == 0 && !enr->md.pending_drop)
+			if (strcmp(enr->md.name, name) == 0)
 				return enr;
 		}
 		if (!search)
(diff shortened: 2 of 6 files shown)


src/backend/access/transam/xact.c
```diff
@@ -380,6 +380,10 @@ IsTransactionState(void)
 {
 	TransactionState s = CurrentTransactionState;
 
+	/* During ENR Rollback, we may need to call relation_open during TRANS_ABORT. */
```
surendravishnoi: comment clearly says that TRANS_ABORT is unsafe. How do we reach to conclusion that it is safe for our use case?

src/backend/access/transam/xact.c
```diff
@@ -631,7 +635,10 @@ AssignTransactionId(TransactionState s)
 
 	/* Assert that caller didn't screw up */
 	Assert(!FullTransactionIdIsValid(s->fullTransactionId));
-	Assert(s->state == TRANS_INPROGRESS);
+
+	/* In ENR Rollback, we may need to call relation_open during TRANS_ABORT. */
+	if (temp_table_xact_support && !ENRInRollback() && s->state != TRANS_ABORT)
```
surendravishnoi: this is very risky. There must be a reason for not assigning new transaction ids while transaction is in aborted state

src/backend/utils/misc/queryenvironment.c
```diff
@@ -146,6 +145,12 @@ register_ENR(QueryEnvironment *queryEnv, EphemeralNamedRelation enr)
 	Assert(enr != NULL);
 	Assert(get_ENR(queryEnv, enr->md.name, false) == NULL);
 
+	if (enr->md.name[0] != '#')
```
surendravishnoi: It is not safe to assume that anything not starting with # must be a table variable? or we are just tracking all non temp table objects? if second, please use appropriate name for the same

src/backend/utils/misc/queryenvironment.c
```diff
@@ -669,13 +674,129 @@ find_enr(Form_pg_depend entry)
...
+		default:
+			/* Shouldn't reach here */
+			elog(ERROR, "Did not find matching ENR catalog entry for catoid %d", catoid);
+			break;
+	}
+
+	/*
+	 * Extra UPDATEs can be skipped because we don't care about intermediate updates - only the initial state.
+	 */
+	if (op == ENR_OP_UPDATE && *list_ptr != NIL)
```
surendravishnoi: how will we handle sub transaction in such cases? even though an update is intermediate in transaction but could be first one in current sub transaction

src/backend/utils/misc/queryenvironment.c
```diff
@@ -978,8 +1111,469 @@ void ENRDropEntry(Oid id)
...
+ * In other words, search through the list of uncommitted tuples for a match, and return if true.
+ */
+bool
+ENRTupleIsDropped(Relation rel, HeapTuple tup)
+{
+	Oid catalog_oid;
+	ListCell *lc;
+	QueryEnvironment *queryEnv = currentQueryEnv;
+
+	if (sql_dialect != SQL_DIALECT_TSQL)
```
surendravishnoi: why are we relying on just dialect here? shouldn't we look into what kind of object it is?

src/backend/utils/misc/queryenvironment.c
```diff
@@ -978,8 +1111,469 @@ void ENRDropEntry(Oid id)
...
+		ListCell *lc2;
+		EphemeralNamedRelation enr = (EphemeralNamedRelation) lfirst(lc);
+
+		if (enr->md.enrtype != ENR_TSQL_TEMP)
+			continue;
+
+		/* Obtain oid to search by from tuple*/
+		switch (catalog_oid)
+		{
+			case RelationRelationId:
```
surendravishnoi: this is code duplication with existing logic, may be we can refactor to return tuple and caller can do relevant operation later

src/backend/utils/misc/queryenvironment.c
```diff
@@ -978,8 +1111,469 @@ void ENRDropEntry(Oid id)
...
+}
+
+/*
+ * ENRCommitChanges
+ *
+ * 1. Free uncommitted tuple data from each ENR, then mark it as committed.
+ * 2. Finish freeing ENRs that were dropped in this transaction.
+ */
+void
+ENRCommitChanges(QueryEnvironment *queryEnv)
```
surendravishnoi, line 1120: what if there a failure after this point? would a rollback happen and how will we do undo in that case?

src/backend/utils/misc/queryenvironment.c
```diff
@@ -978,8 +1111,469 @@ void ENRDropEntry(Oid id)
...
+	/* Add back any ENRs that must be restored. */
+	foreach (lc, queryEnv->dropped_namedRelList)
+	{
+		EphemeralNamedRelation enr = (EphemeralNamedRelation) lfirst(lc);
+
+		Assert(enr->md.enrtype == ENR_TSQL_TEMP);
+		Assert(!enr->md.is_table_variable);
+
+		/* If this was dropped in the current subtransaction, then we must restore it. */
+		if (enr->md.dropped_subid != InvalidSubTransactionId && enr->md.dropped_subid >= subid) // 4
```
surendravishnoi, line 1306: how is dropped subid work here? we have 1, 2... 10 drops in 10 savepoints, how will we know up to which point we want to cleanup with single dropped subid?

src/backend/utils/misc/queryenvironment.c
```diff
@@ -978,8 +1111,469 @@ void ENRDropEntry(Oid id)
...
+	 */
+	foreach(lc, queryEnv->namedRelList)
+	{
+		Relation rel;
+		EphemeralNamedRelation enr = (EphemeralNamedRelation) lfirst(lc);
+
+		if (enr->md.enrtype != ENR_TSQL_TEMP || enr->md.is_table_variable)
+			continue;
+
+		if (!enr->md.is_committed && enr->md.created_subid >= subid)
```
surendravishnoi, line 1324: same as above, how will single sub id let us know which ENRs to drop?

src/backend/utils/misc/queryenvironment.c
```diff
@@ -978,8 +1111,469 @@ void ENRDropEntry(Oid id)
...
+			List *uncommitted_cattups = enr->md.uncommitted_cattups[i];
+			ListCell *lc2;
+
+			foreach(lc2, uncommitted_cattups)
+			{
+				ENRUncommittedTuple uncommitted_tup = (ENRUncommittedTuple) lfirst(lc2);
+				bool skip_cache_inval = false;
+
+				if (!OidIsValid(uncommitted_tup->catalog_oid) || !HeapTupleIsValid(uncommitted_tup->tup))
+					elog(ERROR, "ROLLBACK failed! Invalid temp table entries were found.");
```
surendravishnoi: what happens when we encounter this error? a rollback cannot fail

src/backend/utils/misc/queryenvironment.c
```diff
@@ -978,8 +1111,469 @@ void ENRDropEntry(Oid id)
...
+
+	foreach(lc, queryEnv->namedRelList)
+	{
+		EphemeralNamedRelation enr = (EphemeralNamedRelation) lfirst(lc);
+
+		if (enr->md.enrtype != ENR_TSQL_TEMP || enr->md.is_table_variable)
+			continue;
+
+		ENRDeleteUncommittedTupleData(InvalidSubTransactionId, enr);
+		enr->md.is_committed = true;
```
surendravishnoi, line 1138: we mark it committed here, what happens when a new transaction starts?

src/backend/utils/misc/queryenvironment.c
```diff
@@ -978,8 +1111,469 @@ void ENRDropEntry(Oid id)
...
+		if (enr->md.enrtype != ENR_TSQL_TEMP || enr->md.is_table_variable)
+			continue;
+
+		if (!enr->md.is_committed)
+		{
+			enrs_to_drop = lappend(enrs_to_drop, enr);
+			continue;
+		}
+
+		/* ROLLBACK block. */
```
surendravishnoi: what if there are multiple table versions all committed? like
begin;
create; drop; create; drop; create; drop; create;
commit
why are we keeping all versions and not just one?

src/backend/utils/misc/queryenvironment.c
time-change: it is the second, I will change the name

src/backend/access/transam/xact.c
time-change: See discussion above

src/backend/access/transam/xact.c
time-change: Moved them to AbortTransaction, right before we set `s->state = TRANS_ABORT`. Functionally this is similar, as the Xact callbacks are invoked only and directly from AbortTransaction/AbortSubTransaction. The difference here is that, this code can be invoked regardless of whether our xact_callback functions are installed.

So if it seems necessary and this approach works, we could introduce a bool in _PG_init where the xact callbacks are registered so that the logic remains consistent. For ex, I'm not sure whether the subxact callbacks get removed if Babelfish is uninstalled from an instance, but this code would remain there regardless.

src/backend/utils/misc/queryenvironment.c
time-change: That's true, with subtransaction support this is no longer true, so removed this line to track intermediate updates.

src/backend/utils/misc/queryenvironment.c
time-change: ENRTupleIsDropped is only called from ENRdropTuple - so we don't have information other than the target Relation and HeapTuple, I'm not sure what else we can check for at this point in the function

src/backend/utils/misc/queryenvironment.c
time-change: If it is committed, then it was created outside of the current transaction. We track whether an enr is committed so that during ROLLBACK, we can be sure to drop any ENRs created in the current transaction.

src/backend/utils/misc/queryenvironment.c
time-change: I don't follow the question here - this is in the ROLLBACK block. I'm not sure how multiple table versions can be committed without running into name collision.

src/backend/utils/misc/queryenvironment.c
time-change: There is no way for an ENR to have its dropped subid altered. If the ENR is dropped in a subtransaction, then the subxactid is recorded, and the ENR is moved into the dropped_relid list - this is effectively the same as actually dropping it, but the information sticks around until commit/rollback. If the ENR is later recreated (ie, same name), a **new ENR is created**. Then if that new ENR is dropped, it is moved into the dropped_relid list with the new subtransaction id in which it was created/dropped.

Otherwise, if the same ENR is added back to namedRelList as part of rollback, you can see that the dropped_subid is reset. Then if it is dropped in a later subtransaction, it will be updated accordingly. 

Let me know if I misunderstood the question

src/backend/utils/misc/queryenvironment.c
time-change: See comment above

src/backend/utils/misc/queryenvironment.c
time-change: It should not fail, because this would mean a violated assumption or some issue with memory. But I thought it wouldn't be appropriate to emit a FATAL error to close the server since we are dealing with temp objects here - any bad state could be forcibly cleaned up by opening a new session.

src/backend/utils/misc/queryenvironment.c
time-change: If a failure happens at this point, a rollback would be called and changes in the transaction would be rolled back as usual. 

If any sort of failure happens during ENRRollback though, we would get a crash as it's called during AbortTransaction.

src/backend/utils/misc/queryenvironment.c
lejaokri: I think we will crash anyway.

For instance, try setting `uncommitted_tup->catalog_oid=0` then the code goes to error codepath over and over again until we hit a maximum.

ie
```
ereport(ERROR) -> triggers error handling
error handling goes to AbortTransaction
AbortTransaction calls ENRRollbackChanges
```

src/backend/utils/misc/queryenvironment.c
surendravishnoi: also, can we assume anything starting with # must be a temp table? Please validate that trigger inserted/deleted tables cannot start with #

src/backend/utils/misc/queryenvironment.c
```diff
@@ -146,6 +145,12 @@ register_ENR(QueryEnvironment *queryEnv, EphemeralNamedRelation enr)
 	Assert(enr != NULL);
 	Assert(get_ENR(queryEnv, enr->md.name, false) == NULL);
 
+	if (enr->md.name[0] != '#')
+		enr->md.is_not_bbf_temp_table = true;
+
+	if (!enr->md.is_not_bbf_temp_table && GetCurrentSubTransactionId() != InvalidSubTransactionId)
```
surendravishnoi: why not call it is_bbf_temp_table? !not seems non-intuitive

src/backend/utils/misc/queryenvironment.c
```diff
@@ -669,13 +674,123 @@ find_enr(Form_pg_depend entry)
...
+static void
+ENRAddUncommittedTupleData(EphemeralNamedRelation enr, Oid catoid, ENRTupleOperationType op, HeapTuple tup)
+{
+	ENRUncommittedTuple uncommitted_tup;
+	List **list_ptr = NULL;
+	
+	if (in_enr_rollback)
+		return;
+	
+	uncommitted_tup = (ENRUncommittedTuple) palloc0(sizeof(ENRUncommittedTupleData));
```
surendravishnoi, line 687: what memory context we are using here? we need to make sure is at transaction level

src/backend/utils/misc/queryenvironment.c
```diff
@@ -669,13 +674,123 @@ find_enr(Form_pg_depend entry)
...
+		foreach(lc, enr->md.uncommitted_cattups[i])
+		{
+			ENRUncommittedTuple uncommitted_tup = (ENRUncommittedTuple) lfirst(lc);
+
+			if (subid != InvalidSubTransactionId && uncommitted_tup->subid < subid)
+				continue;
+
+			heap_freetuple(uncommitted_tup->tup);
+			pfree(uncommitted_tup);
+			if (uncommitted_tup->subid >= subid)
```
surendravishnoi: this check is unnecessary

src/backend/utils/misc/queryenvironment.c
surendravishnoi: can we refactor it?

src/backend/utils/misc/queryenvironment.c
surendravishnoi: Point is that we have already got rid of undo data from ENR here. So even though ENRollback is called, it has no undo data to do actual rollback

src/backend/utils/misc/queryenvironment.c
```diff
@@ -978,8 +1105,469 @@ void ENRDropEntry(Oid id)
...
+		return false;
+
+	catalog_oid = RelationGetRelid(rel);
+
+	foreach(lc, queryEnv->namedRelList)
+	{
+		ListCell *lc2;
+		EphemeralNamedRelation enr = (EphemeralNamedRelation) lfirst(lc);
+
+		if (enr->md.enrtype != ENR_TSQL_TEMP)
```
surendravishnoi: why are we not checking for is_not_bbf_temp_table here?

src/backend/utils/misc/queryenvironment.c
```diff
@@ -669,13 +674,123 @@ find_enr(Form_pg_depend entry)
...
+				continue;
+
+			heap_freetuple(uncommitted_tup->tup);
+			pfree(uncommitted_tup);
+			if (uncommitted_tup->subid >= subid)
+				enr->md.uncommitted_cattups[i] = foreach_delete_current(enr->md.uncommitted_cattups[i], lc);
+		}
+		if (subid == InvalidSubTransactionId)
+		{
+			list_free(enr->md.uncommitted_cattups[i]);
```
surendravishnoi, line 773: will list free take care of all elements inside list?

src/backend/utils/misc/queryenvironment.c
surendravishnoi: got it

src/backend/utils/misc/queryenvironment.c
```diff
@@ -951,15 +1073,20 @@ bool ENRaddTuple(Relation rel, HeapTuple tup)
  */
 bool ENRdropTuple(Relation rel, HeapTuple tup)
 {
-	return _ENR_tuple_operation(rel, tup, ENR_OP_DROP);
+	/*
+	 * If we've already dropped it in this transaction but haven't committed, pretend the drop is done. 
+	 */
+	if (temp_table_xact_support && ENRTupleIsDropped(rel, tup))
```
surendravishnoi: which use case is this? why would we call dropped on an already dropped tuple?

src/backend/utils/misc/queryenvironment.c
surendravishnoi: in above example, we will have one entry in named rel list and 3 entries in dropped named list. We are transferring all dropped entries into named list. Why are we doing that? we just need the first one from the list which is committed, others can be dropped immediately

src/backend/utils/misc/queryenvironment.c
```diff
@@ -978,8 +1105,469 @@ void ENRDropEntry(Oid id)
...
+			List *uncommitted_cattups = enr->md.uncommitted_cattups[i];
+			ListCell *lc2;
+
+			foreach(lc2, uncommitted_cattups)
+			{
+				ENRUncommittedTuple uncommitted_tup = (ENRUncommittedTuple) lfirst(lc2);
+				bool skip_cache_inval = false;
+
+				if (!OidIsValid(uncommitted_tup->catalog_oid) || !HeapTupleIsValid(uncommitted_tup->tup))
+					elog(ERROR, "ROLLBACK failed! Invalid temp table entries were found.");
```
surendravishnoi: this will result in server crash.

src/backend/utils/misc/queryenvironment.c
```diff
@@ -978,8 +1105,469 @@ void ENRDropEntry(Oid id)
...
+				ENRUncommittedTuple uncommitted_tup = (ENRUncommittedTuple) lfirst(lc2);
+				bool skip_cache_inval = false;
+
+				if (!OidIsValid(uncommitted_tup->catalog_oid) || !HeapTupleIsValid(uncommitted_tup->tup))
+					elog(ERROR, "ROLLBACK failed! Invalid temp table entries were found.");	
+
+				if (uncommitted_tup->subid < subid)
+					continue;
+
+				if (uncommitted_tup->catalog_oid == IndexRelationId && uncommitted_tup->optype == ENR_OP_ADD)
```
surendravishnoi: this code can be refactored and shared with rollback handling

src/backend/utils/misc/queryenvironment.c
time-change: yes, the trigger table names "inserted" and "deleted" are hardcoded in

src/backend/utils/misc/queryenvironment.c
time-change: this is in queryEnv->memCtx, it is called in _ENR_tuple_operation. same ctx as Rollback/Commit

src/backend/utils/misc/queryenvironment.c
time-change: list_free takes care of the list and all of the ListCell pointers in the list; the elements inside the list were freed in the previous loop

src/backend/utils/misc/queryenvironment.c
time-change: Yes, then functionally it would be a noop. I don't think theres anything wrong with that, for ex. an empty transaction block is fine: 

```
BEGIN transaction
ROLLBACK
```

src/backend/utils/misc/queryenvironment.c
time-change: ack, added a fix in ENRDropEntry

src/backend/utils/misc/queryenvironment.c
time-change: Modified this to just continue - that way we can still attempt to clean up as much as possible and avoid infinite looping in error codepath.

src/backend/access/transam/xact.c
```diff
@@ -631,6 +631,7 @@ AssignTransactionId(TransactionState s)
 
 	/* Assert that caller didn't screw up */
 	Assert(!FullTransactionIdIsValid(s->fullTransactionId));
+
```
surendravishnoi: avoid unnecessary changes

src/backend/utils/misc/queryenvironment.c
surendravishnoi: I am talking about this case
begin;
add ENR
commit -> there is failure after we delete uncommitted data, now rollback will happen. As part of rollback we need to remove ENR but we cannot as we have removed corresponding uncommitted data in failed commit

src/backend/utils/misc/queryenvironment.c
```diff
@@ -978,8 +1103,287 @@ void ENRDropEntry(Oid id)
...
+		
+		Form_pg_index idx_form = (Form_pg_index) GETSTRUCT(uncommitted_tup->tup);
+		EphemeralNamedRelation tmp_enr = get_ENR_withoid(queryEnv, idx_form->indrelid, ENR_TSQL_TEMP);
+		if (tmp_enr)
+		{
+			skip_cache_inval = true;
+		}
+	}
+
+	rel = relation_open(uncommitted_tup->catalog_oid, AccessExclusiveLock);
```
surendravishnoi: do we really need to ope relation in exclusive mode?

src/backend/utils/misc/queryenvironment.c
```diff
@@ -669,13 +674,123 @@ find_enr(Form_pg_depend entry)
 	return NULL;
 }
 
+bool 
+ENRInRollback()
```
surendravishnoi: where do we use this function?

src/backend/utils/misc/queryenvironment.c
```diff
@@ -978,8 +1103,287 @@ void ENRDropEntry(Oid id)
...
+			continue;
+
+		if (!enr->md.is_committed)
+		{
+			enrs_to_drop = lappend(enrs_to_drop, enr);
+			continue;
+		}
+
+		/* ROLLBACK block. */
+		in_enr_rollback = true;
```
surendravishnoi: instead of setting variable, preferably pass a bool indicating the same to skip undo during rollback

src/backend/utils/misc/queryenvironment.c
```diff
@@ -978,8 +1103,287 @@ void ENRDropEntry(Oid id)
...
+	/* If we are dropping a committed ENR, wait until COMMIT to free it. */
+	if (temp_table_xact_support && enr->md.is_bbf_temp_table)
+	{
+		ListCell *lc;
+		enr->md.dropped_subid = GetCurrentSubTransactionId();
+		foreach(lc, currentQueryEnv->dropped_namedRelList)
+		{
+			EphemeralNamedRelation tmp_enr = (EphemeralNamedRelation) (lc);
+
+			if (strcmp(enr->md.name, tmp_enr->md.name) == 0)
```
surendravishnoi: this is incorrect, we remove old entries only if they are in same transaction/sub-transaction

src/backend/utils/misc/queryenvironment.c
surendravishnoi: my suggestion was to avoid adding them to named list and rather drop directly here. Otherwise, we are first adding them to name list, later putting them into ends_to_drop list and drop later

src/backend/utils/misc/queryenvironment.c
time-change: So it is true that if there is a failure after this point that requires a rollback, we won't be able to rollback the metadata properly.

However, I don't see a clean solution for this. If we leave the data around in case it needs to be rolled back AFTER this point in the commit, well then it still needs to be cleaned up at some point. And then the question can be raised - what happens if we hit an error after the cleanup happens then? IMO, it is alright to leave this edge case around because:

1) If we have a failure after ENRDeleteUncommittedTupleData, in our code it could only be during the pfree calls to queryEnv->dropped_namedRelList. If those calls fail for some reason, it would be a memory issue that causes server restart, so the bad metadata would be cleaned up. 

2) If we have a failure after ENRDeleteUncommittedTupleData but it is NOT in our code, then it is in Postgres native transaction handling code, (ie. some point in CommitTransaction in xact.c). I am not sure what PG would do in that scenario as well, since there are a few calls before the section where it is "too late" to abort a transaction inside CommitTransaction. 

3) Ultimately if that does somehow happen, the customer will be able to reach a clean slate by opening a new session.

src/backend/utils/misc/queryenvironment.c
time-change: The original example you gave was 

begin
create; drop; create; drop; create; drop; create
commit.

so the optimization works here where we won't track those intermediate dropped ENRs in dropped_namedRelList.

but with ROLLBACK, i think the overhead is minimal. since none of these are marked as committed, the code will skip processing and then add them to the dropped ENR list anyways. so the difference is just dropping them slightly earlier in this call instead of dropping them at the end.

src/backend/utils/misc/queryenvironment.c
```diff
@@ -978,8 +1095,279 @@ void ENRDropEntry(Oid id)
...
+	/* If we are dropping a committed ENR, wait until COMMIT to free it. */
+	if (temp_table_xact_support && enr->md.is_bbf_temp_table)
+	{
+		ListCell *lc;
+		enr->md.dropped_subid = GetCurrentSubTransactionId();
+		foreach(lc, currentQueryEnv->dropped_namedRelList)
+		{
+			EphemeralNamedRelation tmp_enr = (EphemeralNamedRelation) (lc);
+
+			if (strcmp(enr->md.name, tmp_enr->md.name) == 0 && tmp_enr->md.dropped_subid == enr->md.dropped_subid)
```
surendravishnoi: given that we are handling this during rollback itself, we can skip this optimisation in favour of simplicity

surendravishnoi (dismissed):

src/backend/utils/misc/queryenvironment.c
time-change: Ack, fixed. Noting that surendra has approved PR (in case this gets lost due to force push).

lejaokri (approved):
Looking at notes seems Surendra has approved this one.

forestkeeper (approved):

PR #353, opened 2024-04-29
Fix searchpath handling for tsql udfs inside fmgr

tanscorpio7:
### Description

Currently we override the search_path for pltsql UDFs directly using the GUC variable & avoid set_config.
This leads to crash if there was another set_config('search_path... when we were in this overridden state.
Reason for crash is that the new set_config call may push the current value (the one we set in fmgr) onto
GUC stack. Now it pssoble that we exit the function call and released fcache, but fcache->prosearchpath
is still referenced in GUC stack.

To fix this we fallback to using set_config for setting the search_path & skip the check hook for search path
when setting from inside fmgr which helps with performance.
Also increase the GUC nest level since GUC_ACTION_SAVE is not possible on top of GUC_ACTION_SET
(user set), which is possible in this case. New nest level also means that we do not have to reset the search
path at the end of execution and let AtEOXact_GUC(commit=true) handle it.

We also move the set search path block in fmgr inside TRY, to ensure we reset the skip check search path
variable in case of errors.

### Issues Resolved

[BABEL-4877]

### Sign Off

### Check List

By submitting this pull request, I confirm that my contribution is under the terms of the Apache 2.0 and PostgreSQL licenses, and grant any person obtaining a copy of the contribution permission to relicense all or a portion of my contribution to the PostgreSQL License solely to contribute all or a portion of my contribution to the PostgreSQL open source project.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/babelfish-for-postgresql/babelfish_extensions/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/src/backend/catalog/namespace.c
+++ b/src/backend/catalog/namespace.c
@@ -4450,6 +4450,7 @@ check_search_path(char **newval, void **extra, GucSource source)
 	char	   *rawname;
 	List	   *namelist;
 
+	/* quick exit for babelfish when setting search path in fmgr_security_definer */
 	if (sql_dialect == SQL_DIALECT_TSQL && set_local_schema_for_func_hook
 		&& pltsql_check_search_path == false)
 	{
--- a/src/backend/catalog/namespace.c
+++ b/src/backend/catalog/namespace.c
@@ -4450,6 +4450,12 @@ check_search_path(char **newval, void **extra, GucSource source)
 	char	   *rawname;
 	List	   *namelist;
 
+	if (sql_dialect == SQL_DIALECT_TSQL && set_local_schema_for_func_hook
+		&& pltsql_check_search_path == false)
+	{
+		pltsql_check_search_path = true;
+		return true;
+	}
 	/* Need a modifiable copy of string */
 	rawname = pstrdup(*newval);
 
--- a/src/backend/utils/fmgr/fmgr.c
+++ b/src/backend/utils/fmgr/fmgr.c
@@ -709,7 +709,7 @@ fmgr_security_definer(PG_FUNCTION_ARGS)
 	int			non_tsql_proc_count = 0;
 	void	   *newextra = NULL;
 	char 	   *cacheTupleProcname = NULL;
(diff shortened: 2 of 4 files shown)


src/backend/utils/fmgr/fmgr.c
```diff
@@ -803,9 +804,12 @@ fmgr_security_definer(PG_FUNCTION_ARGS)
 
 	if (fcache->prosearchpath)
 	{
-		old_search_path = namespace_search_path;
-		namespace_search_path = fcache->prosearchpath;
-		assign_search_path(fcache->prosearchpath, newextra);
+		pltsql_save_nestlevel = NewGUCNestLevel();
+		pltsql_check_search_path = false;
+		(void) set_config_option("search_path", fcache->prosearchpath,
```
surendravishnoi: see if we can move it to try/catch

src/backend/catalog/namespace.c
```diff
@@ -4450,6 +4450,13 @@ check_search_path(char **newval, void **extra, GucSource source)
 	char	   *rawname;
 	List	   *namelist;
 
+	/* quick exit for babelfish when setting search path in fmgr_security_definer */
+	if (sql_dialect == SQL_DIALECT_TSQL && set_local_schema_for_func_hook
```
surendravishnoi, line 4454: validate if dialect check is correct here in case of PG to TSQL call

surendravishnoi (commented):
Please add problem description and solution in commit message

src/backend/catalog/namespace.c
tanscorpio7: we set the dialect in fmgr before calling set_cofig searchpath.
and we only change search path for tsql user defined functions. so dialect will be tsql.

src/backend/utils/fmgr/fmgr.c
tanscorpio7: moved inside try block

surendravishnoi (approved):

PR #359, opened 2024-05-09
[OSS-ONLY] Fix crash in InitializeParallelDSM due to MyProcPort not set

ahmed-shameem:
### Description

Crash in InitializeParallelDSM during VACUUM by a bg worker

InitializeParallelDSM was crashing when a cron job was scheduled to vacuum a table, index or index build/reindexing as there are done by maintenance worker. This change was pushed into serval branches. 

This is happenning because for maintenance worker MyProcPort is not set which is why we were getting segfault. With this commit, we have updated the condition to check whether MyProcPort is set or not.
 

Signed-off-by: Shameem Ahmed [shmeeh@amazon.com](mailto:shmeeh@amazon.com)
 
### Check List

### DIFF
--- a/src/backend/tcop/postgres.c
+++ b/src/backend/tcop/postgres.c
@@ -4592,7 +4592,10 @@ PostgresMain(const char *dbname, const char *username)
 			ReportChangedGUCOptions();
 
 			if (MyProcPort && MyProcPort->protocol_config->fn_send_ready_for_query)
+			{
+				Assert(MyProcPort != NULL);
 				MyProcPort->protocol_config->fn_send_ready_for_query(whereToSendOutput);
+			}
 			else
 				ReadyForQuery(whereToSendOutput);
 			send_ready_for_query = false;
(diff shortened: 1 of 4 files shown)


src/backend/tcop/postgres.c
```diff
@@ -4592,7 +4592,10 @@ PostgresMain(const char *dbname, const char *username)
 			ReportChangedGUCOptions();
 
 			if (MyProcPort && MyProcPort->protocol_config->fn_send_ready_for_query)
+			{
+				Assert(MyProcPort != NULL);
```
rishabhtanwar29: Unnecessary assert here as we are already checking if `MyProcPort` is not null in the condition above.
Additionally, we also need to add a check for `MyProcPort` in `ReadCommand` isn't it?

ahmed-shameem: Right, got it confused with something else. Updated in latest rev

rishabhtanwar29 (approved):

shardgupta (approved):

PR #371, opened 2024-05-16
Fix BBF PIVOT cannot use CTE as source table

RIC06X:
Fixed BBF PIVOT stmt cannot use CTE as source table

### Description

Fixed BBF PIVOT stmt cannot use CTE as source table
 
### Issues Resolved

BABEL-4959 
### Check List

### DIFF
--- a/src/backend/parser/analyze.c
+++ b/src/backend/parser/analyze.c
@@ -1440,7 +1440,7 @@ transformSelectStmt(ParseState *pstate, SelectStmt *stmt)
 	qry->commandType = CMD_SELECT;
 
 	/* process the WITH clause independently of all else */
-	if (stmt->withClause && !stmt->isPivot)
+	if (stmt->withClause && ( sql_dialect == SQL_DIALECT_PG || transform_pivot_clause_hook == NULL || !stmt->isPivot))
 	{
 		qry->hasRecursive = stmt->withClause->recursive;
 		qry->cteList = transformWithClause(pstate, stmt->withClause);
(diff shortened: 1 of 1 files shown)


src/backend/parser/analyze.c
```diff
@@ -1440,7 +1440,7 @@ transformSelectStmt(ParseState *pstate, SelectStmt *stmt)
 	qry->commandType = CMD_SELECT;
 
 	/* process the WITH clause independently of all else */
-	if (stmt->withClause)
+	if (stmt->withClause && !stmt->isPivot)
```
tanscorpio7: should we first check for hook here ? we can probably reuse transform_pivot_clause_hook

xhfanhe (dismissed):

src/backend/parser/analyze.c
forestkeeper: So does it mean skip transform select Pivot && withClause ?

RIC06X: If in that case, then we will skip the intoClause check.

RIC06X: Yes, this will skip the withClause  transformation for the wrapper sql select ... bbf_pivot(), where we need the withClause is the src_sql, so we moved the withClause from the wrapper sql to src_sql and delayed its transformation

tanscorpio7: I didn't mean actually calling the hook here. Just a safety check. I was suggesting something like
```if (stmt->withClause && (transform_pivot_clause_hook == NULL || !(dialect == tsql && stmt->isPivot)))```
This just provides surety that this change does not come into picture if babelfish isn't installed.

RIC06X: Thanks, will update here

tanscorpio7 (approved):

forestkeeper (approved):

PR #373, opened 2024-05-20
Fix differences in T-SQL COALESCE function while calling the function with variables and constants together

basasairohan:
### Description

This commit will fix the following issue with T-SQL COALESCE function : 

1. When the arguments contain the mix of variables and constants as shown below, the below query should return 4 as it is the first non-null value. But BBF throws error as we try to evaluate the constant ('dfidhfgi') at the compile time in this case.
```
DECLARE @abc int = 4 ;
select COALESCE(@abc, 'dfidhfgi')
```
 
### Issues Resolved

BABEL-726
Signed-off-by: Sai Rohan Basa [bsrohan@amazon.com](mailto:bsrohan@amazon.com)
 
### Check List

### DIFF
--- a/src/backend/optimizer/util/clauses.c
+++ b/src/backend/optimizer/util/clauses.c
@@ -3281,13 +3281,9 @@ eval_const_expressions_mutator(Node *node,
 					 * are evaluated at runtime. Hence, skipping eval_const_expressions_mutator()
 					 * at this step.
 					 */
-					if (sql_dialect == SQL_DIALECT_TSQL && !coalesceexpr->tsql_is_null)
-					{
+					if (sql_dialect == SQL_DIALECT_TSQL && !coalesceexpr->tsql_is_null 
+						&& lfirst(arg) != NULL && ((Node *) lfirst(arg))->type == T_CoerceViaIO)
 						e = (Node *) lfirst(arg);
-						if (e->type != T_CoerceViaIO)
-							e = eval_const_expressions_mutator((Node *) lfirst(arg),
-													   			context);
-					}
 					else
 						e = eval_const_expressions_mutator((Node *) lfirst(arg),
 													   		context);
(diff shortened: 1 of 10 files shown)


src/backend/optimizer/util/clauses.c
```diff
@@ -3276,8 +3276,16 @@ eval_const_expressions_mutator(Node *node,
 				{
 					Node	   *e;
 
-					e = eval_const_expressions_mutator((Node *) lfirst(arg),
-													   context);
+					if (sql_dialect == SQL_DIALECT_TSQL && !coalesceexpr->tsql_is_null)
```
thephantomthief: Please check with @surendravishnoi if this logic should be encapsulated in a hook. It looks fine to me but I wonder if this might lead to merge conflicts at time of community merges

surendravishnoi: we can simply the condition put all checks in one line
if (not TSQL || tsql is null || type is coerce)
   eval const expr mutator

also add a comment explaining why we skip this line for given conditions

basasairohan: We can not put all the checks in a single line as variable `e` would be NULL while checking the condition.

src/backend/optimizer/util/clauses.c
```diff
@@ -3276,6 +3276,11 @@ eval_const_expressions_mutator(Node *node,
 				{
 					Node	   *e;
 
+					/*
+					 * T-SQL Coalesce is handled differently, constant expressions
+					 * are evaluated at the runtime. Hence, skipping eval_const_expressions_mutator()
```
thephantomthief: Nit: `the` is not needed

thephantomthief (approved):

src/backend/optimizer/util/clauses.c
surendravishnoi: I mean check like this
if (sql_dialext != SQL_DIALECT_TSQL || coalesceexpr->tsql_is_null || (lfirst(arg) == NULL) || (((Node *) lfirst(arg))->type ==  T_CoerceViaIO)
     e = eval_const_expressions_mutator((Node *) lfirst(arg), ...

surendravishnoi (approved):

PR #375, opened 2024-05-22
Don't take locks for Babelfish Temp Tables.

Sairakan:
### Description

[Describe what this change achieves]
 
### Issues Resolved

### Check List


src/backend/catalog/dependency.c
```diff
@@ -1569,13 +1569,15 @@ AcquireDeletionLock(const ObjectAddress *object, int flags)
 {
 	if (object->classId == RelationRelationId)
 	{
+		if (flags & PERFORM_DELETION_CONCURRENT_LOCK)
+			LockRelationOid(object->objectId, AccessShareLock);
```
time-change: I don't think this is viable. Eventually on RO we will not be able to take locks that exist in shared memory. So it should either be NoLock or a new local-only lock struct. Or maybe see if it's possible to define a new LockMode specifically for temp tables that makes it look like a lock is taken to the rest of engine code, but we don't touch any shared memory structs (if something like that is possible).

PR #384, opened 2024-06-06
Disable shared cache invalidation for temp table

time-change:
### Description

Since we have moved temp table metadata into local memory, we don't need to add cache invalidations to shared memory when they concern temp tables. 

Manual Testing: 

Tracking with `pg_waldump` to see whether any cache inval messages show up in WAL when we create temp tables/manipulate metadata showed none are created. 
 
### Issues Resolved

BABEL-5012

This change is similar to a previous patch from Surendra that we ended up not committing - originally the idea was that it might improve performance and we didn't end up finding a significant improvement in performance. Turns out we need the change for proper concurrency support with temp tables. The only difference is that we also exclude those inval messages from going into WAL 
 
### Check List

### DIFF
--- a/src/backend/access/heap/heapam.c
+++ b/src/backend/access/heap/heapam.c
@@ -2007,7 +2007,7 @@ heap_insert(Relation relation, HeapTuple tup, CommandId cid,
 	 * the heaptup data structure is all in local memory, not in the shared
 	 * buffer.
 	 */
-	CacheInvalidateHeapTuple(relation, heaptup, NULL);
+	CacheInvalidateHeapTuple(relation, heaptup, NULL, false);
 
 	/* Note: speculative insertions are counted too, even if aborted later */
 	pgstat_count_heap_insert(relation, 1);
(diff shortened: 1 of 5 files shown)


src/backend/access/heap/heapam.c
```diff
@@ -2007,7 +2007,7 @@ heap_insert(Relation relation, HeapTuple tup, CommandId cid,
 	 * the heaptup data structure is all in local memory, not in the shared
 	 * buffer.
 	 */
-	CacheInvalidateHeapTuple(relation, heaptup, NULL);
+	CacheInvalidateHeapTuple(relation, heaptup, NULL, false);
```
surendravishnoi, line 2010: this change is very intrusive on engine side. We should think of alternatives where we have minimal changes on code outside query environment.

src/backend/utils/cache/inval.c
```diff
@@ -792,10 +796,13 @@ AcceptInvalidationMessages(void)
  *		Initialize inval data for the current (sub)transaction.
  */
 static void
-PrepareInvalidationState(void)
+PrepareInvalidationState(bool isEnr)
 {
 	TransInvalidationInfo *myInfo;
 
+	if (transInvalInfo != NULL && !isEnr)
```
surendravishnoi, line 803: how is this correct? transInvalInfo is transaction level structure while invalidation is relation specific. How can I mark local only in this structure where my transaction can have mix of both temp and regular tables?

src/backend/access/heap/heapam.c
time-change: agree that we should try to make this change as unintrusive as possible. Main issue is that most of the abstraction around inval messages doesn't have enough knowledge of whether it is being invoked on a temp-table related invalidation or not; so our code will likely need to touch many layers to work regardless of the approach.

time-change:
Closing as we'll proceed with the approach in #411

PR #386, opened 2024-06-11
Fix engine crash in tsql_openjson_with_get_subjsonb

Jakeowen1:
### Description
This change resolves two engine crashes in `tsql_openjson_with_get_subjsonb`. The first crash was related to the way that `strict` and `lax` modes were handled in openjson column paths. In the case where a user did not put a space between `strict/lax` and the column name, it would cause the column type to be left NULL, resulting in a NULL entry into strlen and a crash. This change handles strict/lax modes properly, and adds a null check for column type to prevent crashes. The second crash was caused by a null value being passed into a json result list. In cases where a given json path did not exist, we would represent this by create a list of one null element. This element would be passed into the `JsonbValueToJsonb` function and result in crash. This commit removes the null entry from the list to prevent this crash from happening. 
 
### Issues Resolved
BABEL-3820
 
### Check List

### DIFF
--- a/src/backend/utils/adt/jsonpath_exec.c
+++ b/src/backend/utils/adt/jsonpath_exec.c
@@ -526,13 +526,27 @@ tsql_openjson_with_columnize(Jsonb *jb, char *col_info)
 			asjson;
 	col_path = NULL; col_type = NULL; strict = false; as = false; asjson = false;
 	token = strtok(col_info, " ");
+
+	if (strcmp(token, "strict") == 0)
+	{
+		strict = true;
+		if(strlen(token) == 6)
+			token = strtok(NULL, " ");
+		else
+			token = token + 5;
+	}
+	else if (strcmp(token, "lax") == 0)
+	{
+		strict = false;
+		if(strlen(token) == 3)
+			token = strtok(NULL, " ");
+		else
+			token = token + 2;
+	}
+
 	while (token != NULL)
 	{
-		if (strncmp(token, "strict", 6) == 0)
-			strict = true;
-		else if (strncmp(token, "lax", 3) == 0)
-			strict = false;
-		else if (col_path == NULL)
+		if (col_path == NULL)
 			col_path = token;
 		else if (col_type == NULL)
 			col_type = token;
@@ -543,7 +557,7 @@ tsql_openjson_with_columnize(Jsonb *jb, char *col_info)
 		token = strtok(NULL, " ");
 	}
 
-	if (strlen(col_type) >= 3) /* Get column size restriction, if it exists */
+	if (col_type && strlen(col_type) >= 3) /* Get column size restriction, if it exists */
 	{
 		token = strtok(col_type, "(");
 		if (token)
@@ -577,6 +591,14 @@ tsql_openjson_with_columnize(Jsonb *jb, char *col_info)
 
 	/* get tuple set using executeJsonPath */
 	jp = DatumGetJsonPathP(DirectFunctionCall1(jsonpath_in, CStringGetDatum(col_path)));
+
+	/*
+	 * In case where there is no space between 'strict/lax' and the column path
+	 * jsonpath_in will automatically set the header to strict. We need to change
+	 * the header to prevent unintended 'strict' calls to openjson
+	 */
+	if(!strict)
+		jp->header = JSONPATH_LAX + 1;
 
 	(void) executeJsonPath(jp, vars, jb, false, &found, false);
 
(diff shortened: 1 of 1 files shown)


src/backend/utils/adt/jsonpath_exec.c
```diff
@@ -578,6 +592,14 @@ tsql_openjson_with_columnize(Jsonb *jb, char *col_info)
 	/* get tuple set using executeJsonPath */
 	jp = DatumGetJsonPathP(DirectFunctionCall1(jsonpath_in, CStringGetDatum(col_path)));
 
+	/*
+	 * In case where there is no space between 'strict/lax' and the column path
+	 * jsonpath_in will automatically set the header to strict. We need to change
+	 * the header to prevent unintended 'strict' calls to openjson
+	 */
+	if(!strict)
+		jp->header = JSONPATH_LAX + 1;
```
Sairakan: What is the meaning of `JSONPATH_LAX + 1`? Is it trying to set both the `JSONPATH_LAX` and `JSONPATH_VERSION` flags? In that case, it's probably better to use something like `JSONPATH_LAX & JSONPATH_VERSION`

Sairakan: Or more likely, just use `jp->header |= JSONPATH_LAX`, unless that is not giving the desired behavior

Jakeowen1: Yeah `jp->header |= JSONPATH_LAX` is the same behavior im looking for, I can update the code. The point is to get it to pass a couple check statements such as this in the `executeJsonPath` function
`cxt.laxMode = (path->header & JSONPATH_LAX) != 0;` which will set it to lax mode rather than keeping it as strict, and this assert in the `jspInit` function `Assert((js->header & ~JSONPATH_LAX) == JSONPATH_VERSION);`

Sairakan (approved):

src/backend/utils/adt/jsonpath_exec.c
```diff
@@ -486,8 +486,17 @@ tsql_openjson_with_get_subjsonb(PG_FUNCTION_ARGS)
 	vars = (Jsonb *) DirectFunctionCall1(jsonb_in, CStringGetDatum("{}"));
 	(void) executeJsonPath(jp, vars, jb, false, &found, false);
 
-	if (JsonValueListLength(&found) >= 1)
-		sub_jb = JsonbValueToJsonb(JsonValueListHead(&found));
+	if (JsonValueListLength(&found) >= 1) {
+		JsonbValue *jv = JsonValueListHead(&found);
```
forestkeeper, line 490: Is that jv = NULL guaranteed only can be at list head ?

src/backend/utils/adt/jsonpath_exec.c
```diff
@@ -526,13 +535,27 @@ tsql_openjson_with_columnize(Jsonb *jb, char *col_info)
 			asjson;
 	col_path = NULL; col_type = NULL; strict = false; as = false; asjson = false;
 	token = strtok(col_info, " ");
+
+	if (strncmp(token, "strict", 6) == 0)
```
forestkeeper: Is it possible to have lower case and upper case issue in here ?

forestkeeper: If not , pls add a test case of STRICT upper case

src/backend/utils/adt/jsonpath_exec.c
```diff
@@ -872,10 +903,10 @@ executeItemOptUnwrapTarget(JsonPathExecContext *cxt, JsonPathItem *jsp,
 						found->list = list_make2(found->singleton, NULL);
 						found->singleton = NULL;
 					}
-					else if (!found->list)
-						found->list = list_make1(NULL); /* Since JsonValueList uses a NULL singleton as shortcut, need to manually insert null value into list */
-					else
+					else if (found->list)
 						found->list = lappend(found->list, NULL);
+					else
+						found->list = list_make1(NULL); /* Since JsonValueList uses a NULL singleton as shortcut, need to manually insert null value into list */
```
forestkeeper: It's not PG format comment
```
					else
                        /* Since JsonValueList uses a NULL singleton as shortcut, need to manually insert null value into list */
						found->list = list_make1(NULL);
```

src/backend/utils/adt/jsonpath_exec.c
```diff
@@ -543,7 +566,7 @@ tsql_openjson_with_columnize(Jsonb *jb, char *col_info)
 		token = strtok(NULL, " ");
 	}
 
-	if (strlen(col_type) >= 3) /* Get column size restriction, if it exists */
+	if (col_type && strlen(col_type) >= 3) /* Get column size restriction, if it exists */
```
forestkeeper: The format for comment should be : 
```
if (col_type && strlen(col_type) >= 3) 
{
        /* Get column size restriction, if it exists */
}
```

forestkeeper: Either inside if, or before if

src/backend/utils/adt/jsonpath_exec.c
Jakeowen1: Mixed case currently handled incorrectly, will modify code and add test cases
```
1> SELECT * FROM OPENJSON('{"obj":{}}') WITH(field int 'STRICT $.obj')
2> GO
syntax error at end of jsonpath input
1> SELECT * FROM OPENJSON('{"obj":{}}') WITH(field int 'StRiCT $.obj')
2> go
syntax error at end of jsonpath input
```

src/backend/utils/adt/jsonpath_exec.c
Jakeowen1: We only care about the first element of the list which we return as a Jsonb type

forestkeeper (approved):

PR #387, opened 2024-06-14
Fix memory leak in remove_queryEnv

time-change:
### Description

In remove_queryEnv we need to also free any tuples that have been created via `heap_copyTuple` in ENRs. Otherwise they will leak memory until the connection is closed. 

Clean test run here: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2669

I tested this by cherry-picking to GitFarm and replacing the binaries on a mammoth instance, which showed significant memory usage improvement on the same test that Kristian ran during validation.
 
### Issues Resolved

BABEL-5033
 
### Check List

### DIFF
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -98,6 +98,33 @@ create_queryEnv2(MemoryContext cxt, bool top_level)
 	return queryEnv;
 }
 
+/* Loop through structures in the ENR and make sure to free anything that needs to be freed. */
+static void free_ENR(EphemeralNamedRelation enr)
+{
+	for (int i = 0; i < ENR_CATTUP_END; i++)
+	{
+		List *uncommitted_cattups = enr->md.uncommitted_cattups[i];
+		List *cattups = enr->md.cattups[i];
+		ListCell *lc2, *lc3;
+
+		foreach(lc2, uncommitted_cattups)
+		{
+			ENRUncommittedTuple uncommitted_tup = (ENRUncommittedTuple) lfirst(lc2);
+
+			heap_freetuple(uncommitted_tup->tup);
+		}
+
+		foreach(lc3, cattups)
+		{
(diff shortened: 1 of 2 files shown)


src/backend/utils/misc/queryenvironment.c
```diff
@@ -102,13 +102,53 @@ create_queryEnv2(MemoryContext cxt, bool top_level)
...
 
 	/* We should never "free" top level query env as it's in stack memory. */
 	if (!currentQueryEnv || currentQueryEnv == topLevelQueryEnv)
 		return;
 
 	tmp = currentQueryEnv->parentEnv;
 	oldcxt = MemoryContextSwitchTo(currentQueryEnv->memctx);
+
+	/* Clean up structures in currentQueryEnv */
+	foreach(lc, currentQueryEnv->dropped_namedRelList)
```
lejaokri, line 142: 1) Do you need to do same deep free logic in `ENRCommitChanges()`?

2) I know this isn't directly related, but should we fix the memory leak in `pgstat_drop_relation` as part of this delivery? I think that's pretty safe to do.

3) Why not combine the two `for (int i = 0; i < ENR_CATTUP_END; i++)` into 1 for loop?

time-change: 1. ENRCommitChanges does this deep free already. It's only in remove_queryEnv where we are freeing a whole queryEnv. 
2+3 Will do

lejaokri: Thanks for handling the comments.
Can you please create a fake PR on the extension side so we run all BBF testcases?

time-change:
fake PR running tests here: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2669

lejaokri (approved):
issue seems fixed. see internal Jira for results.
BBF pipeline test is failing the same way with or without this new code.

xhfanhe (approved):

PR #394, opened 2024-06-25
Fix concurrency issues with ENR temp tables.

Sairakan:
### Description

With the temp OID buffer introduced in [5a69710](https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/commit/5a69710c6dd38b58e2860a1f5b89edb5f4611e77), it is possible for temp tables from different sessions to share the same OID. Bypass locking for ENR temp tables in order to prevent these sessions from interfering with each other unnecessarily - they should not need to wait on each other to perform operations on their own temp tables.

Also, prevent invalidation messages for local temp tables from being added to the shared invalidation (SI) queue. Otherwise, under high concurrency we may see errors such as `relation XXXXX deleted while still in use` due to one backend receiving an invalidation message from a different backend for an OID which it is trying to use for its own temp table, when that table should still be valid.
 
### Issues Resolved

BABEL-4913, BABEL-4892
 
### Check List


src/backend/utils/cache/inval.c
```diff
@@ -459,6 +460,7 @@ AddRelcacheInvalidationMessage(InvalidationMsgsGroup *group,
 	msg.rc.id = SHAREDINVALRELCACHE_ID;
 	msg.rc.dbId = dbId;
 	msg.rc.relId = relId;
+	msg.rc.local_only = get_ENR_withoid(currentQueryEnv, relId, ENR_TSQL_TEMP);
```
time-change: get_ENR_withoid isn't exactly a slow function, but perhaps we can avoid the extra performance overhead by only calling it if we know we are in temp table path? we don't need to do all this if BBF isn't enabled, if temp OID isn't enabled, etc.

src/include/storage/lock.h
```diff
@@ -178,11 +179,16 @@ typedef struct LOCKTAG
...
+ * For SET_LOCKTAG_RELATION, if it is an ENR temp table we also add MyProcPid into the locktag
+ * in order to disambiguate locks from different backends in case we enter into the shared lock space.
+ */
+
 /* ID info for a relation is DB OID + REL OID; DB OID = 0 if shared */
 #define SET_LOCKTAG_RELATION(locktag,dboid,reloid) \
 	((locktag).locktag_field1 = (dboid), \
 	 (locktag).locktag_field2 = (reloid), \
-	 (locktag).locktag_field3 = 0, \
+	 (locktag).locktag_field3 = (get_ENR_withoid(currentQueryEnv, reloid, ENR_TSQL_TEMP) ? MyProcPid : 0), \
```
time-change: any way we can move this to a separate macro like above? i imagine this macro is called a lot, so the less we need to call get_ENR_withoid, the better

src/include/storage/sinval.h
```diff
@@ -80,6 +80,14 @@ typedef struct
...
 	Oid			dbId;			/* database ID, or 0 if a shared relation */
 	Oid			relId;			/* relation ID, or 0 if whole relcache */
+	/* 
+	 * We need a way to mark whether this inval message is for a local ENR entry,
+	 * so that we can skip inserting it into the SI message queue.
+	 * Note that checking whether the relId is an ENR entry at the time of insertion
+	 * is insufficient, as during drops the ENR entry is deleted before we perform
+	 * SI queue insertion.
+	 */
+	bool		local_only;
```
time-change, line 90: can we do the same for catcache invalidations?

src/backend/utils/cache/inval.c
Sairakan: Hm, maybe we might want to consider enhancing the performance of this lookup in the first place. I redid the test where I did not drop the tables until the end, and it increased the execution time from 52 seconds to 3 minutes 26 seconds. So there is definitely a degradation in performance as the number of simultaneous temp tables increases.

Sairakan: After consulting with Alex, it seems that it is extremely unlikely for customers to have more than single digit temp tables active at a single time, so the performance scaling is not much of a concern.

src/include/storage/lock.h
Sairakan: As far as I can tell, all of the places that call this macro are where it is possible for the target relation to be an ENR table, so I don't think it would make any difference whether we check here or in the source function.

lejaokri: do we need TSQL check here?

Sairakan: To be honest, I am not 100% sure whether we need a sql_dialect check here. Do we intend for ENR temp tables/temp tables on RO to only work with TSQL dialect, or are we considering generalizing the implementation to work with non-babelfish instances as well? 

As is, I don't think that it is unsafe to not have a sql_dialect check here - it will simply add a slight overhead due to the function call, but the function itself is already programmed quite defensively.

src/include/storage/sinval.h
Sairakan: I don't see why not. However, as it is, there don't seem to be any issues related to catcache invalidations being sent to the SI queue, so it doesn't seem to be necessary.

src/include/storage/lock.h
time-change: yes, we want temp on RO to only work on TSQL tables. it's a separate discussion whether we eventually want to support a use case on native APG.

src/backend/utils/cache/inval.c
time-change: if the difference in performance is that drastic (3x worse in that case) then maybe we need to see if we can just have a fast path lookup function for this case.

src/include/storage/sinval.h
time-change: adding this local_only field to SI messages essentially addresses the cache inval change as well, so we should combine the two. if the intent is also have this SI fix as part of the lock changes then catcache invalidations need to be addressed too since temp tables certainly emit catcache invals.

src/backend/executor/execIndexing.c
```diff
@@ -243,7 +243,7 @@ ExecCloseIndices(ResultRelInfo *resultRelInfo)
 			continue;			/* shouldn't happen? */
 
 		/* Drop lock acquired by ExecOpenIndices */
-		index_close(indexDescs[i], RowExclusiveLock);
+		index_close(indexDescs[i], get_ENR_withoid(currentQueryEnv, indexDescs[i]->rd_id, ENR_TSQL_TEMP) ? AccessShareLock : RowExclusiveLock);
```
surendravishnoi: can we override the lock in index_close function itself? that way it is a separate line with less chance of merge conflict

src/backend/utils/cache/inval.c
surendravishnoi: can we skip adding entry and simply return? that way we do not need local_only flag. Also, let us have this check as a hook so that we can skip it for non-babelfish dbs

src/include/storage/sinval.h
Sairakan: I think that unless identifying temp-table specific catcache messages is trivial to implement, it is better suited to include in a future PR. As is, this PR fully resolves all of the issues found related to high concurrency workloads with the temp oid buffer enabled.

src/include/storage/lock.h
```diff
@@ -178,11 +179,16 @@ typedef struct LOCKTAG
  * rather than accessing the fields directly.  Note multiple eval of target!
  */
 
+/*
```
surendravishnoi: we need to acquire locks in shared mode a two places, can we do tag manipulation there only in some way? It is not ideal to call get ENR for each call here. Also, let us see if there are some existing hooks we can use here (in case)

src/backend/utils/cache/inval.c
Sairakan: The thing is that we *do* need local invalidation messages even for ENR tables. So it is necessary to allow the inval message to be created and saved, we just need to prevent it from being put in the shared inval queue.

src/backend/executor/execIndexing.c
Sairakan: I looked into this, but the problem is that index_close() only takes in the oid of the index, not the temp table - so as is, it cannot tell whether the input index is being used for an ENR table or not. 

I did experiment earlier with adding an extra field in the QueryEnvironment that would store *all* Oids being used in ENR, whether it is an index or other catalog. With that implementation it would be possible to put the check into index_close(), but it also requires quite a bit of extra added code.

src/include/storage/lock.h
Sairakan: hm, it turns out that this change may not actually be necessary. I ran a test where I had two sessions which both exceeded the fast path lock limit (so went into shared locks) without this change, and both sessions were able to freely create, insert, update, etc their own temp tables.

src/backend/parser/parse_relation.c
```diff
@@ -1602,6 +1602,7 @@ addRangeTableEntryForRelation(ParseState *pstate,
 	RangeTblEntry *rte = makeNode(RangeTblEntry);
 	RTEPermissionInfo *perminfo;
 	char	   *refname = alias ? alias->aliasname : RelationGetRelationName(rel);
+	lockmode = get_ENR_withoid(currentQueryEnv, rel->rd_id, ENR_TSQL_TEMP) ? AccessShareLock : lockmode;
```
time-change: this check should only be done in tsql dialect + if temp oid is enabled, right? lets avoid as much unnecessary overhead as we can

src/include/storage/sinval.h
time-change: then maybe we could move the cache inval changes out of this PR? i think it's best to not mix lock changes with an incomplete fix for cache inval in one PR, especially since the relcache portion is just a few lines of code.

src/backend/parser/parse_relation.c
Sairakan: I can move the dialect check within the function itself, since the function is called in a number of places anyways. There should be minimal overhead from the function call in that case.

src/backend/executor/execIndexing.c
Sairakan: scratch that - my previous answer doesn't apply at all. I'll move the check inside of index_close()

src/include/storage/sinval.h
Sairakan: The thing is that this change is required to fix the issues with `relation deleted while still in use` and `relation missing (x) attributes`. Unless we still don't plan on enabling temp oids in Uakari at all, I think that this change is important for overall stability, while the changes for cache inval don't seem as necessary.

src/backend/parser/parse_relation.c
Sairakan: Just ran some tests, with 100 threads each creating + dropping 1,000 temp tables.

1. temp oid disabled, default locking: 2:33.621066
2. temp oid disabled, reduced locking: 2:33.507345
3. temp oid enabled, reduced locking: 2:31.988261

So overall there seems to be little performance difference.

src/backend/access/index/indexam.c
```diff
@@ -183,6 +183,9 @@ index_close(Relation relation, LOCKMODE lockmode)
 {
 	LockRelId	relid = relation->rd_lockInfo.lockRelId;
 
+	if (get_ENR_withoid(currentQueryEnv, RelationGetRelid(relation), ENR_TSQL_TEMP))
```
time-change: minor nit, we could create a macro for this call because it is used so often across our codebase, the `get_ENR_withoid(currentQueryEnv, [relid], ENR_TSQL_TEMP)`. doesn't necessarily have to be in this change

src/include/storage/sinval.h
time-change: ok. i will make the catcache changes in a separate change (#411 ) since it is a bit more invasive.

time-change (approved):
only nit is concerning the macro; not a blocker for merge though.

time-change:
do we have a corresponding extension PR running our test suite?

Sairakan:
Yes, the extension PR running the test suite is here https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2686

src/backend/storage/ipc/sinvaladt.c
```diff
@@ -491,6 +491,12 @@ SIInsertDataEntries(const SharedInvalidationMessage *data, int n)
 		max = segP->maxMsgNum;
 		while (nthistime-- > 0)
 		{
+			if (data->id == SHAREDINVALRELCACHE_ID && data->rc.local_only)
```
time-change: minor nit: we can consider putting this in a hook

src/backend/storage/lmgr/lmgr.c
```diff
@@ -332,6 +333,9 @@ CheckRelationLockedByMe(Relation relation, LOCKMODE lockmode, bool orstronger)
 {
 	LOCKTAG		tag;
 
+	if (get_ENR_withoid(currentQueryEnv, RelationGetRelid(relation), ENR_TSQL_TEMP))
```
time-change: can we wrap `get_ENR_withoid(currentQueryEnv, [relid], ENR_TSQL_TEMP)` pattern in a macro, since it is used in many temp table related places. would improve readability

time-change (approved):
approved with minor suggestions

src/backend/storage/ipc/sinvaladt.c
```diff
@@ -491,6 +491,12 @@ SIInsertDataEntries(const SharedInvalidationMessage *data, int n)
 		max = segP->maxMsgNum;
 		while (nthistime-- > 0)
 		{
```
surendravishnoi, line 496: this must be under a hook

src/backend/storage/lmgr/lmgr.c
surendravishnoi: I would suggest to wrap it under a hook, will provide us double protection

src/include/storage/sinval.h
surendravishnoi, line 90: how are we making sure that this variable is false in all code paths. We will start with uninitialised value

src/backend/storage/lmgr/lock.c
```diff
@@ -845,6 +845,27 @@ LockAcquireExtended(const LOCKTAG *locktag,
 						lockMethodTable->lockModeNames[lockmode]),
 				 errhint("Only RowExclusiveLock or less can be acquired on database objects during recovery.")));
 
+	if (locktag->locktag_type == LOCKTAG_RELATION &&
```
surendravishnoi: Let us ad hooks for acquire, for release can use get_ENR_withoid hook

src/include/storage/sinval.h
Sairakan: The only code path which sets this value is inside of `AddRelcacheInvalidationMessage()` which is guaranteed to set the value to true or false. I have done a full search of the code base, and no other code assigns any values to `msg.rc`.

src/backend/utils/cache/inval.c
lejaokri: nit: we can skip calling `get_ENR_withoid()` if relid is invalid so no need to search for the entire array.

I noticed the callers of `RegisterRelcacheInvalidation()` have access to relation in a way to check if relation is a temp table. It might makes sense to consider changing the function signature in the future to pass an argument if table is temp so we could potentially skip the seq scan  if we're dealing with non-temp tables. But that is a bit invasive ofcourse.

My (nonblocking) concern is that we would also do the sequential search even for non-temporary tables if we do it this way. If a session has a lot of temp tables, then this would also degrade perf for all types of tables. This also wont be a no-op for non-babelfish instance either.

time-change (approved):

lejaokri (approved):

surendravishnoi (approved):

PR #395, opened 2024-06-26
Include ordering for constraints on babelfish tables

rishabhtanwar29:
### Description
Babelfish supports column ordering and NULLs ordering in table constraints but PG
does not. It was implemented through special grammar rule and few dialect T-SQL dialect
dependent logic in engine. Currently this is not handled in pg_dump so the information
about column/nulls ordering gets lost during dump. Although, this information is available
in the underlying index for a UNIQUE/PRIMARY KEY constraint but this information anyway
gets lost since the underlying index does not get dumped for a UNIQUE/PRIMARY KEY constraint,
instead it is dumped as a constraint.

To fix this issue, we will now dump the underlying index of a UNIQUE/PRIMARY KEY constraint
and make use of ALTER TABLE ADD CONSTRAINT USING snytax to add that as constraint.
Postgres disallows creation of an constraint using index like this for non default sort ordering as
native pg_dump does not handle such constraints. So we will need to bypass that check for
Babelfish dump/restore with this special handling in bbf_dump.

Task: BABEL-4888

Extension PR: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2688
 
### Check List

### DIFF
--- a/src/backend/parser/parse_utilcmd.c
+++ b/src/backend/parser/parse_utilcmd.c
@@ -61,6 +61,7 @@
 #include "rewrite/rewriteManip.h"
 #include "utils/acl.h"
 #include "utils/builtins.h"
+#include "utils/guc.h"
 #include "utils/lsyscache.h"
 #include "utils/partcache.h"
 #include "utils/rel.h"
@@ -2246,6 +2247,7 @@ transformIndexConstraint(Constraint *constraint, CreateStmtContext *cxt)
 		oidvector  *indclass;
 		Datum		indclassDatum;
 		int			i;
+		const char *bbf_dump_restore = GetConfigOption("babelfishpg_tsql.dump_restore", true, false);
 
 		/* Grammar should not allow this with explicit column list */
 		Assert(constraint->keys == NIL);
@@ -2384,10 +2386,11 @@ transformIndexConstraint(Constraint *constraint, CreateStmtContext *cxt)
 
 				defopclass = GetDefaultOpClass(attform->atttypid,
 											   index_rel->rd_rel->relam);
-				if (indclass->values[i] != defopclass ||
+				if ((indclass->values[i] != defopclass ||
 					attform->attcollation != index_rel->rd_indcollation[i] ||
 					attoptions != (Datum) 0 ||
-					index_rel->rd_indoption[i] != 0)
+					index_rel->rd_indoption[i] != 0) &&
+					(!bbf_dump_restore || strcmp(bbf_dump_restore, "on") != 0)) /* The dump/reload problem has been fixed for babelfish database */
 					ereport(ERROR,
 							(errcode(ERRCODE_WRONG_OBJECT_TYPE),
 							 errmsg("index \"%s\" column number %d does not have default sorting behavior", index_name, i + 1),
--- a/src/bin/pg_dump/dump_babel_utils.c
+++ b/src/bin/pg_dump/dump_babel_utils.c
@@ -224,6 +224,33 @@ dumpBabelGUCs(Archive *fout)
 }
 
 /*
+ * serverSupportsConstraintUsingIndex:
+ * Helper function to determine if remote server supports constraint
+ * using index. It is supported on version 15.8 higher and 16.4 or higher.
+ */
+static bool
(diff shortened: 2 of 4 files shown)


src/backend/parser/parse_utilcmd.c
```diff
@@ -2384,10 +2386,11 @@ transformIndexConstraint(Constraint *constraint, CreateStmtContext *cxt)
 
 				defopclass = GetDefaultOpClass(attform->atttypid,
 											   index_rel->rd_rel->relam);
-				if (indclass->values[i] != defopclass ||
+				if ((indclass->values[i] != defopclass ||
 					attform->attcollation != index_rel->rd_indcollation[i] ||
 					attoptions != (Datum) 0 ||
-					index_rel->rd_indoption[i] != 0)
+					index_rel->rd_indoption[i] != 0) &&
```
surendravishnoi, line 2392: Please make sure that we have test coverage for this code path in APG tests

src/bin/pg_dump/pg_dump.c
```diff
@@ -16550,7 +16552,7 @@ dumpIndex(Archive *fout, const IndxInfo *indxinfo)
 	 * emitted comment has to be shown as depending on the constraint, not the
 	 * index, in such cases.
 	 */
-	if (!is_constraint)
+	if (!is_constraint || bbfShouldDumpIndex(fout, indxinfo, &is_constraint))
```
surendravishnoi: instead of changing is_constraint inside function, set it to false in if block itself

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -1873,3 +1918,73 @@ babelfishDumpOpclassHelper(Archive *fout, const OpclassInfo *opcinfo, PQExpBuffe
...
+		appendPQExpBufferStr(q, " DEFERRABLE");
+		if (constrinfo->condeferred)
+			appendPQExpBufferStr(q, " INITIALLY DEFERRED");
+	}
+	appendPQExpBufferStr(q, ";\n");
+
+	resetPQExpBuffer(delq);
+	appendPQExpBuffer(delq, "ALTER %sTABLE ONLY %s ", foreign,
+					  fmtQualifiedDumpable(tbinfo));
+		appendPQExpBuffer(delq, "DROP CONSTRAINT %s;\n",
```
surendravishnoi: indentation

src/bin/pg_dump/dump_babel_utils.c
```diff
@@ -223,6 +223,33 @@ dumpBabelGUCs(Archive *fout)
 	destroyPQExpBuffer(qry);
 }
 
+/*
+ * serverSupportsConstraintUsingIndex:
+ * Helper function to determine if remote server supports constraint
+ * using index. It is supported on version 15.8 higher and 16.4 or higher.
+ */
+static bool
+serverSupportsConstraintUsingIndex(Archive *fout)
```
surendravishnoi: We should remove this check for old versions, it would be better to fail rather than going with old behavior where we loose information about constraint ordering.

src/bin/pg_dump/pg_dump.c
rishabhtanwar29: Sounds good, updated!

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: Fixed!

src/bin/pg_dump/dump_babel_utils.c
rishabhtanwar29: Discussed offline, removed the check in newer revision.

src/backend/parser/parse_utilcmd.c
rishabhtanwar29: 1. We have create_index.sql test to verify this behavior for normal PG.
2. The babelfishpg_tsql.dump_restore GUC won't even be available for normal PG server(with babelfish not installed) and we have proper NULL check for the GUC below which will make sure not to bypass the condition for normal PG test.

surendravishnoi (approved):

sumitj824 (approved):

PR #409, opened 2024-07-13
Support non-default collation at database level in Babelfish

ahmed-shameem:
### Description

Currently babelfish supports setting collation at server, column and expression levels. User can only set collation at database level, if it is same as server level.
This commit enables the users to set collation at database level using TSQL syntax:
```
CREATE DATABASE <db_name> COLLATE <collation_name>
```

This will set the collation of expressions, columns of tables of string datatypes as same as database level collation by default. If any user wants to have a different collation, then they can set it at column or expression level.
User can choose NOT to set collation at database level explicitly. In that case, the database collation will be as same as server level collation. The identifiers will be using server level collation itself.

This commit updates the collation of datatypes and parameters to facilitate the use of database level collation. We have made following two changes -- 
1. Changes in get_typcollation(...) and typeTypeCollation(...), Functions responsible to supply collations for different node of the parse tree. We have to update it to return CLUSTER_COLLATION_OID() for collatable Babelfish datatypes so that assign_collation_walker can set collation according to database or server level later.
2. Update handling of param collation through fixed_paramref_hook(...), function which transforms ParamRef to Param. This special handling for PG dialect is required to support checking of foreign key constraints which happens in PG dialect even if insert/update is being done through TSQL dialect. Refer to comment made on hook implementation for more details.

Extension side PR: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2739

Task: BABEL-4953
 
Authored by: Dipesh Dhameliya <dddhamel@amazon.com>
Co-authored by: Shameem Ahmed <shmeeh@amazon.com>


src/backend/parser/parse_param.c
```diff
@@ -115,7 +117,14 @@ fixed_paramref_hook(ParseState *pstate, ParamRef *pref)
...
 	param->paramtypmod = -1;
-	param->paramcollid = get_typcollation(param->paramtype);
+	if (handle_param_collation_hook)
+	{
+		param->paramcollid = handle_param_collation_hook(param);
+	}
+	else
+	{
+		param->paramcollid = get_typcollation(param->paramtype);
+	}
```
kuntalghosh, line 127: Why not rewrite fixed_paramref_hook in TSQL extension and use pre_parse_analyze_hook to replace with
```
pstate->p_paramref_hook = tsql_fixed_paramref_hook
```
Also, I noticed that get_typcollation already has the same check as set_param_collation. So, I'm not sure why we need this change.

src/backend/parser/parse_type.c
```diff
@@ -659,6 +659,15 @@ typeTypeCollation(Type typ)
...
 
 	typtup = (Form_pg_type) GETSTRUCT(typ);
+	if (OidIsValid(typtup->typcollation) &&
+		sql_dialect == SQL_DIALECT_TSQL && pg_strcasecmp(get_namespace_name(typtup->typnamespace), "sys") == 0)
+	{
+		/*
+		 * Always set CLUSTER_COLLATION_OID() for babelfish collatable types so that
+		 * we can set collation according to database or server level later.
+		 */
+		return CLUSTER_COLLATION_OID();
```
kuntalghosh: Let's introduce a hook here instead and move this code in the extension.

src/backend/utils/cache/lsyscache.c
```diff
@@ -3051,6 +3051,16 @@ get_typcollation(Oid typid)
...
+		if (IsNormalProcessingMode() && OidIsValid(typtup->typcollation) &&
+			sql_dialect == SQL_DIALECT_TSQL && pg_strcasecmp(get_namespace_name(typtup->typnamespace), "sys") == 0)
+		{
+			/*
+			 * Always set CLUSTER_COLLATION_OID() for babelfish collatable types so that
+			 * we can set collation according to database or server level later.
+			 */
+			result = CLUSTER_COLLATION_OID();
+		}
+
```
kuntalghosh: Let's just call typeTypeCollation by default.
1. Why do we need IsNormalProcessingMode() check here?
2. There're multiple callers of get_typcollation where we're again setting collation to CLUSTER_COLLATION_OID. Let's remove those codes that are no longer needed because of this change.
3. How does it affect inter-operability? I'm concerned that we're using different collations depending on the dialect. Won't that output wrong data in PG dialect?

kuntalghosh (changes requested):

src/backend/utils/cache/lsyscache.c
kuntalghosh: Also, what are the scenarios when a built-in datatype is created with non-default collation? The built-in datatypes are in our control. So, why not fix the collation when we create the same!!

src/backend/parser/parse_param.c
Deepesh125: This special handling is required for PG dialect while checking for constraints, eg., foreign key (FK) constraint. If we try to insert some values on T-SQL table then PG in turn will create query like `SELECT 1 FROM ONLY \"master_dbo\".\"t5\" x WHERE \"a\" OPERATOR(pg_catalog.=) $1 FOR KEY SHARE OF x` to evaluate FK which would be done in PG dialect. 

If we set data type default collation (which will always be same as `babelfishpg_tsql.server_collation_name`) then we may face collation conflict error is T-SQL table is defined under non-default database level collation (ie. `db collation` is not equal `babelfishpg_tsql.server_collation_name`).

This special handling will set `DEFAULT_COLLATION_OID` to param so that other collation (collation of left operand) will be picked up always. Note - collation of left operand will always be defined in all cases.

src/backend/parser/parse_type.c
Deepesh125: Done

src/backend/utils/cache/lsyscache.c
Deepesh125: > Let's just call typeTypeCollation by default.

Instead I am calling hook defined to handle default collation as there is special handling required for string Const node.

> Why do we need IsNormalProcessingMode() check here?

This part of code is also getting called during bootstrap so this check was added to safely handle that case. But since we are moving the handling to extension through hook, this check is no more required and hence removed it. 

> There're multiple callers of get_typcollation where we're again setting collation to CLUSTER_COLLATION_OID. 

Let's remove those codes that are no longer needed because of this change.
Yes, code cleanup is done with additional handling for some cases, ie., handling default collation of user defined data type. Default collation of UDF should always be same as database collation. check https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2739/commits/d0417a36e5917c2edc750531f7dcb7d0b7198665 and https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2739/commits/c50e653e127eb90a11505da69b0419093cdb3562


>How does it affect inter-operability? I'm concerned that we're using different collations depending on the dialect. Won't that output wrong data in PG dialect?

There will not be any impact on inter-operability as collation of collatable data is always defined. So even in PG dialect, appropriate collation will be picked up.

> Also, what are the scenarios when a built-in datatype is created with non-default collation? The built-in datatypes are in our control. So, why not fix the collation when we create the same!!

built-in datatypes are always defined with collation same as server_collation_name. But with database level collation being supported (which might be different from server_collation_name), we need this handling to assign database level collation whenever object collation is not explicitly stated.

kuntalghosh (approved):

forestkeeper (approved):

PR #410, opened 2024-07-18
Fix error when trying to drop columns from temp tables.

Sairakan:
### Description

Our original code for ENRgetSystableScan did not take into account the fact that pg_depend scans can use up to three keys. Fix the issue by making sure to use the third key if it is supplied.

Also fix the logic around performing the deletion so that we don't try to drop the entire ENR table when deleting columns.

Backport to PG15 as well.
 
### Issues Resolved

BABEL-4912
 
### Check List

### DIFF
--- a/src/backend/catalog/dependency.c
+++ b/src/backend/catalog/dependency.c
@@ -1391,7 +1391,12 @@ deleteOneObject(const ObjectAddress *object, Relation *depRel, int flags)
 	DeleteSecurityLabel(object);
 	DeleteInitPrivs(object);
 
-	// Delete from ENR - noop if not found from ENR
+	/*
+	 * If objectSubId != 0, then this is a column. There are no ENR entries
+	 * for individual columns, so skip ENRDropEntry in this case (or else we
+	 * will delete the entire table instead of just the column). Note that this
+	 * is a no-op if the objectId is not found from ENR.
+	 */
 	if (object->objectSubId == 0)
 		ENRDropEntry(object->objectId);
 
--- a/src/backend/catalog/dependency.c
+++ b/src/backend/catalog/dependency.c
@@ -887,6 +887,11 @@ findDependentObjects(const ObjectAddress *object,
 		Form_pg_depend foundDep = (Form_pg_depend) GETSTRUCT(tup);
 		int			subflags;
 
+		if (foundDep->refclassid == object->classId &&
+			foundDep->refobjid == object->objectId &&
+			foundDep->refobjsubid != object->objectSubId)
+			continue;
+
 		otherObject.classId = foundDep->classid;
 		otherObject.objectId = foundDep->objid;
 		otherObject.objectSubId = foundDep->objsubid;
@@ -1392,7 +1397,8 @@ deleteOneObject(const ObjectAddress *object, Relation *depRel, int flags)
 	DeleteInitPrivs(object);
 
 	// Delete from ENR - noop if not found from ENR
-	ENRDropEntry(object->objectId);
+	if (object->objectSubId == 0)
+		ENRDropEntry(object->objectId);
 
 	/*
 	 * CommandCounterIncrement here to ensure that preceding changes are all
--- a/src/backend/catalog/dependency.c
+++ b/src/backend/catalog/dependency.c
@@ -887,11 +887,6 @@ findDependentObjects(const ObjectAddress *object,
 		Form_pg_depend foundDep = (Form_pg_depend) GETSTRUCT(tup);
 		int			subflags;
 
-		if (foundDep->refclassid == object->classId &&
-			foundDep->refobjid == object->objectId &&
-			foundDep->refobjsubid != object->objectSubId)
-			continue;
-
 		otherObject.classId = foundDep->classid;
 		otherObject.objectId = foundDep->objid;
 		otherObject.objectSubId = foundDep->objsubid;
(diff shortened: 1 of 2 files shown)


src/backend/catalog/dependency.c
```diff
@@ -1392,7 +1392,8 @@ deleteOneObject(const ObjectAddress *object, Relation *depRel, int flags)
 	DeleteInitPrivs(object);
 
 	// Delete from ENR - noop if not found from ENR
-	ENRDropEntry(object->objectId);
+	if (object->objectSubId == 0)
```
time-change, line 1400: so then how does an ENR entry with `object->objectSubId != 0` get cleaned up?

src/backend/utils/misc/queryenvironment.c
```diff
@@ -435,25 +435,27 @@ bool ENRgetSystableScan(Relation rel, Oid indexId, int nkeys, ScanKey key, List
...
 				* So we cannot return right away if there is a match.
 				*/
 				ListCell   *lc;
 				foreach(lc, enr->md.cattups[ENR_CATTUP_DEPEND]) {
 					Form_pg_depend tup = (Form_pg_depend) GETSTRUCT((HeapTuple) lfirst(lc));
 					if (indexId == DependDependerIndexId &&
 						tup->classid == (Oid)v1 &&
-						tup->objid == (Oid)v2)
+						tup->objid == (Oid)v2 &&
+						(nkeys == 2 || tup->objsubid == (int32)v3))
```
time-change, line 449: is it safe to just check if tup->objsubid == v3? since v3 if not set is 0, and tup->objsubid will be 0 for non-column lookups

src/backend/catalog/dependency.c
Sairakan: There aren't any ENR entries for such objects - ENR entries are only created for the base table, which would have a subid of 0. This test case is to fix an erroneous case where we are trying to delete something that doesn't actually have an ENR entry at all.

src/backend/utils/misc/queryenvironment.c
Sairakan: That doesn't work - if the caller of this function only wants 2 keys, then they want all pg_depend entries that match the objid, no matter what the subid is - so we need to completely ignore the subid in that case.

src/backend/catalog/dependency.c
forestkeeper, line 1400: Should use InvalidOid instead of 0 in this case

src/backend/utils/misc/queryenvironment.c
forestkeeper, line 449: Why other have (Oid) to cast to Oid, while v3 is cast to (int32) , and v1,v2,v3 is really confusing in naming, should assign a more reasonable naming.

src/backend/catalog/dependency.c
Sairakan: objectSubId is not an Oid, it is an int32 (see the definition in pg_depend https://www.postgresql.org/docs/current/catalog-pg-depend.html or in the code for FormData_pg_depend). I don't think it would be appropriate to use InvalidOid when this is not an Oid value, this is an int value representing the column number represented in the pg_depend tuple (or 0 in the case that it is not used).

src/backend/utils/misc/queryenvironment.c
Sairakan: As above, this is cast to int32 in order to match the definition of the field in ForData_pg_depend. We use generic v1, v2, v3 naming as they are used to refer to the keys being passed into the function, which could be for any catalog table, not just pg_depend. Thus, the value they represent can be multiple different types depending on which catalog we are searching.

src/backend/catalog/dependency.c
Sairakan: In all other places in the code, this value is explicitly compared to 0, e.g. https://github.com/postgres/postgres/blob/master/src/backend/catalog/dependency.c#L557

src/backend/utils/misc/queryenvironment.c
```diff
@@ -435,25 +435,27 @@ bool ENRgetSystableScan(Relation rel, Oid indexId, int nkeys, ScanKey key, List
...
 					{
 						*tuplist = list_insert_nth(*tuplist, index++, lfirst(lc));
 						*tuplist_flags |= SYSSCAN_ENR_NEEDFREE;
 						found = true;
 					}
 					else if (indexId == DependReferenceIndexId &&
 						tup->refclassid == (Oid)v1 &&
-						tup->refobjid == (Oid)v2)
+						tup->refobjid == (Oid)v2 &&
+						(nkeys == 2 || tup->refobjsubid == (int32)v3))
```
lejaokri, line 458: I see. good catch!

src/backend/catalog/dependency.c
lejaokri: nit: please add a comment why we explicitly check for `objectSubId`. I cant seem to find the testcase for this code change

Sairakan: Sorry, I worded my comment awkwardly - the overall fix requires this change in order to work. Without it, we will get an error when trying to ALTER TABLE DROP COLUMN, as the table will erroneously be deleted when we only wanted to drop one column. So in a sense, all of the test cases for the feature in https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2753 exercise this code path.

`This test case is to fix an erroneous case` should have been `This check is to fix an erroneous case`.

I'll add a comment clarifying why we need this check.

lejaokri (approved):

PR #411, opened 2024-07-23
Disable Shared Catcache Invalidation Messages for Temp Tables

time-change:
### Description

An alternate change to #384 - here we disable SI messages on a more granular level. Most notably adding infrastructure around Catcache invalidations so we can track whether they involve a TSQL-style temp table or not. Note that relcache invalidations are addressed by Jason's PR #394 

Catcache invalidation messages should not be added to the Shared Invalidation Message queue. Catcache inval messages are comprised of three fields, `id` indicating the id of the cache that maps to the catalog, `dbId` indicating the OID of the database, and `hashValue` which is a hashed key for the catcache.

We don't need to send invalidation messages regarding catcache to the SI queue, we only need to process them locally. So when adding to the global SI queue, we add a check to throw out out any messages that are ENR only.

### Issues Resolved

BABEL-5012
 
### Check List

### DIFF
--- a/src/backend/storage/ipc/sinvaladt.c
+++ b/src/backend/storage/ipc/sinvaladt.c
@@ -26,6 +26,7 @@
 #include "storage/shmem.h"
 #include "storage/sinvaladt.h"
 #include "storage/spin.h"
+#include "utils/queryenvironment.h"
 
 /* hooks */
 pltsql_is_local_only_inval_msg_hook_type pltsql_is_local_only_inval_msg_hook = NULL;
--- a/src/backend/storage/ipc/sinvaladt.c
+++ b/src/backend/storage/ipc/sinvaladt.c
@@ -19,6 +19,7 @@
 
 #include "access/transam.h"
 #include "miscadmin.h"
+#include "parser/parser.h"
 #include "storage/backendid.h"
 #include "storage/ipc.h"
 #include "storage/proc.h"
--- a/src/backend/storage/ipc/sinvaladt.c
+++ b/src/backend/storage/ipc/sinvaladt.c
@@ -19,7 +19,6 @@
 
 #include "access/transam.h"
 #include "miscadmin.h"
-#include "parser/parser.h"
 #include "storage/backendid.h"
 #include "storage/ipc.h"
 #include "storage/proc.h"
@@ -27,7 +26,6 @@
 #include "storage/shmem.h"
 #include "storage/sinvaladt.h"
 #include "storage/spin.h"
-#include "utils/queryenvironment.h"
 
 /* hooks */
 pltsql_is_local_only_inval_msg_hook_type pltsql_is_local_only_inval_msg_hook = NULL;
--- a/src/backend/utils/cache/catcache.c
+++ b/src/backend/utils/cache/catcache.c
@@ -40,6 +40,7 @@
 #include "utils/rel.h"
 #include "utils/resowner_private.h"
 #include "utils/syscache.h"
+#include "utils/queryenvironment.h"
 
 
  /* #define CACHEDEBUG */	/* turns DEBUG elogs on */
@@ -2219,7 +2220,8 @@ void
 PrepareToInvalidateCacheTuple(Relation relation,
 							  HeapTuple tuple,
 							  HeapTuple newtuple,
-							  void (*function) (int, uint32, Oid))
+							  void (*function) (int, uint32, Oid), 
+							  bool is_enr)
 {
 	slist_iter	iter;
 	Oid			reloid;
@@ -2261,6 +2263,9 @@ PrepareToInvalidateCacheTuple(Relation relation,
 		dbid = ccp->cc_relisshared ? (Oid) 0 : MyDatabaseId;
 
 		(*function) (ccp->id, hashvalue, dbid);
+		/* If this is a temp table, store the hashvalue somewhere. That way we can remember to not add it to the SI queue at EOXact. */
+		if (is_enr)
+			SaveCatcacheMessage(ccp->id, hashvalue, dbid);
 
 		if (newtuple)
 		{
@@ -2269,7 +2274,11 @@ PrepareToInvalidateCacheTuple(Relation relation,
 			newhashvalue = CatalogCacheComputeTupleHashValue(ccp, ccp->cc_nkeys, newtuple);
 
 			if (newhashvalue != hashvalue)
+			{
 				(*function) (ccp->id, newhashvalue, dbid);
+				if (is_enr)
+					SaveCatcacheMessage(ccp->id, newhashvalue, dbid);
+			}
 		}
 	}
 }
--- a/src/backend/utils/cache/catcache.c
+++ b/src/backend/utils/cache/catcache.c
@@ -40,7 +40,6 @@
 #include "utils/rel.h"
 #include "utils/resowner_private.h"
 #include "utils/syscache.h"
-#include "utils/queryenvironment.h"
 
 
  /* #define CACHEDEBUG */	/* turns DEBUG elogs on */
@@ -2270,9 +2269,7 @@ PrepareToInvalidateCacheTuple(Relation relation,
 			newhashvalue = CatalogCacheComputeTupleHashValue(ccp, ccp->cc_nkeys, newtuple);
 
 			if (newhashvalue != hashvalue)
-			{
 				(*function) (ccp->id, newhashvalue, dbid);
-			}
 		}
 	}
 }
--- a/src/backend/utils/cache/catcache.c
+++ b/src/backend/utils/cache/catcache.c
@@ -2220,8 +2220,7 @@ void
 PrepareToInvalidateCacheTuple(Relation relation,
 							  HeapTuple tuple,
 							  HeapTuple newtuple,
-							  void (*function) (int, uint32, Oid), 
-							  bool is_enr)
+							  void (*function) (int, uint32, Oid))
 {
 	slist_iter	iter;
 	Oid			reloid;
@@ -2263,9 +2262,6 @@ PrepareToInvalidateCacheTuple(Relation relation,
 		dbid = ccp->cc_relisshared ? (Oid) 0 : MyDatabaseId;
 
 		(*function) (ccp->id, hashvalue, dbid);
-		/* If this is a temp table, store the hashvalue somewhere. That way we can remember to not add it to the SI queue at EOXact. */
-		if (is_enr)
-			SaveCatcacheMessage(ccp->id, hashvalue, dbid);
 
 		if (newtuple)
 		{
@@ -2276,8 +2272,6 @@ PrepareToInvalidateCacheTuple(Relation relation,
 			if (newhashvalue != hashvalue)
 			{
 				(*function) (ccp->id, newhashvalue, dbid);
-				if (is_enr)
-					SaveCatcacheMessage(ccp->id, newhashvalue, dbid);
 			}
 		}
 	}
--- a/src/backend/utils/cache/inval.c
+++ b/src/backend/utils/cache/inval.c
@@ -116,6 +116,7 @@
 #include "catalog/catalog.h"
 #include "catalog/pg_constraint.h"
 #include "miscadmin.h"
+#include "parser/parser.h"
 #include "storage/sinval.h"
 #include "storage/smgr.h"
 #include "utils/catcache.h"
@@ -403,6 +404,9 @@ AddCatcacheInvalidationMessage(InvalidationMsgsGroup *group,
 	msg.cc.id = (int8) id;
 	msg.cc.dbId = dbId;
 	msg.cc.hashValue = hashValue;
+
+	if (sql_dialect == SQL_DIALECT_TSQL)
+		msg.cc.local_only = SIMessageIsForTempTable(&msg);
 
 	/*
 	 * Define padding bytes in SharedInvalidationMessage structs to be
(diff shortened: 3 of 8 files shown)


src/backend/storage/ipc/sinvaladt.c
```diff
@@ -491,6 +493,12 @@ SIInsertDataEntries(const SharedInvalidationMessage *data, int n)
 		max = segP->maxMsgNum;
 		while (nthistime-- > 0)
 		{
+			if(sql_dialect == SQL_DIALECT_TSQL && (data->id > 0 && data->cc.local_only))
```
time-change: btw, i will move this to a hook to match the implementation in #394.

src/backend/utils/cache/catcache.c
```diff
@@ -2223,6 +2225,9 @@ PrepareToInvalidateCacheTuple(Relation relation,
 		dbid = ccp->cc_relisshared ? (Oid) 0 : MyDatabaseId;
 
 		(*function) (ccp->id, hashvalue, dbid);
```
time-change: note: maybe can add a new function here for local_only. register a different function in CacheInvalidateHeapTupleENR and then use the hook directly to check the message in SIInsertDataEntries, no need for `local_only` tag

src/backend/utils/cache/inval.c
```diff
@@ -1326,6 +1345,127 @@ CacheInvalidateHeapTuple(Relation relation,
 	RegisterRelcacheInvalidation(databaseId, relationId);
 }
 
+void
+CacheInvalidateENRHeapTuple(Relation relation,
```
Sairakan, line 1361: This looks like almost a 1:1 copy of CacheInvalidateHeapTuple - is it not possible to just modify CacheInvalideHeapTuple rather than duplicating so much of its code? If not, we should have a comment to explain what the differences are, and why they are necessary

time-change: It's not possible to modify CacheInvalidateHeapTuple because there is not enough information to determine whether the catalog in question is referencing an ENR catalog or not. Refactoring the function isn't desirable either because it would mean far more engine changes.

src/backend/utils/misc/queryenvironment.c
```diff
@@ -151,6 +153,9 @@ void remove_queryEnv() {
 	list_free(currentQueryEnv->dropped_namedRelList);
 	currentQueryEnv->dropped_namedRelList = NIL;
 
+	list_free(currentQueryEnv->savedCatcacheMessages);
+	currentQueryEnv->dropped_namedRelList = NIL;
```
surendravishnoi: should be savedCatcacheMessages

src/backend/utils/misc/queryenvironment.c
```diff
@@ -241,6 +246,83 @@ bool has_existing_enr_relations()
...
+ * SI queue, as catalog changes are all session-local.
+ * 
+ * See LocalExecuteInvalidationMessage
+ */
+bool SIMessageIsForTempTable(const SharedInvalidationMessage *msg)
+{
+	if (sql_dialect != SQL_DIALECT_TSQL || temp_oid_buffer_size == 0)
+		return false;
+
+	if (msg->id >= 0)
```
surendravishnoi, line 286: switch will simplify multiple conditions which return false

src/backend/utils/cache/inval.c
```diff
@@ -1326,6 +1351,133 @@ CacheInvalidateHeapTuple(Relation relation,
...
+	 * If we're not prepared to queue invalidation messages for this
+	 * subtransaction level, get ready now.
+	 */
+	PrepareInvalidationState();
+
+	/*
+	 * First let the catcache do its thing
+	 */
+	tupleRelId = RelationGetRelid(relation);
+	if (RelationInvalidatesSnapshotsOnly(tupleRelId))
```
surendravishnoi, line 1398: we should simplify this function, shared relation should not apply to temp tables. Also, we can calculate database id at one place in simple manner.

src/backend/utils/cache/inval.c
```diff
@@ -1326,6 +1351,133 @@ CacheInvalidateHeapTuple(Relation relation,
 	RegisterRelcacheInvalidation(databaseId, relationId);
 }
 
+/*
+ * This is nearly identical to CacheInvalidateHeapTuple but specifically called from
```
surendravishnoi, line 1355: we should put these functions in extension wherever possible.

src/backend/utils/cache/inval.c
```diff
@@ -552,6 +552,30 @@ RegisterCatcacheInvalidation(int cacheId,
...
+ * the SI queue later.
+ * 
+ * This is necessary because there isn't enough information in a 
+ * Catcache SI message itself to determine whether it's applying to ENR.
+ * 
+ * Saved messages are cleared at the end of the current transaction
+ * (see ClearSavedCatcacheMessages).
+ */
+static void
+RegisterENRCatcacheInvalidation(int cacheId,
```
surendravishnoi, line 570: how do we validate that we are locally invalidating cache wherever needed? we should have test for each use case which applies to temp tables catalog tables

src/backend/utils/misc/queryenvironment.c
```diff
@@ -241,6 +246,83 @@ bool has_existing_enr_relations()
 	return false;
 }
 
+/* This only stores catcache inval messages meant for ENR tables in the current transaction. */
+void SaveCatcacheMessage(int cacheId,
+							 uint32 hashValue,
+							 Oid dbId)
+{
+	SharedInvalCatcacheMsg *msg = (SharedInvalCatcacheMsg *) palloc0(sizeof(SharedInvalCatcacheMsg));
```
surendravishnoi, line 254: what if this list grows very large? also can there be case where hash value is not unique across entries?

src/backend/utils/misc/queryenvironment.c
time-change: this cannot be a switch, because msg->id is not a constant: 

```
queryenvironment.c:309:3: error: case label does not reduce to an integer constant
```

maybe due to SI messages being `typedef union`?

src/backend/utils/cache/inval.c
time-change: pg_depend and pg_shdepend are in ENR, so this codepath can still be called invoked. unless you're saying we can skip RegisterSnapshotInvalidation entirely?

src/backend/utils/misc/queryenvironment.c
time-change: 1. The list growing large would be no different than the size of the inval message queue growing very large. I'm not aware of any cap on number of invalidation messages in one transaction.
2. We can't alter the hash value here because it must match the hash that is expected by the existing SI code in `SIInsertDataEntries`. So if there is a collision in PG, then there would be one here. 

the hash function itself is `CatalogCacheComputeHashValue`, which doesn't provide comments on whether the algorithm might produce any collisions

src/backend/utils/cache/inval.c
time-change: several of the functions needed here are static to `inval.c`. they can't easily be moved to hooks in extension without refactoring more engine code along the way

src/backend/utils/cache/inval.c
time-change: added tests in extension

src/backend/utils/cache/inval.c
surendravishnoi: for example, does IsSharedRelation apply to temp tables, if not databaseId will be MyDatabaseId and no need for multiple condition checks

src/backend/utils/cache/inval.c
surendravishnoi: if there are 1 or 2 functions, we can add wrappers and move it to extension. Otherwise, it is fine in current form

src/backend/utils/misc/queryenvironment.c
surendravishnoi: PG inval queue is not used for lookup. Here, we do lookup whenever a new entry is added. So, we need to think if this can be an issue in average case as we are doing sequential loopkup vs using hash map etc.

src/backend/utils/cache/inval.c
time-change: yes, it can. the relation here is the catalog relation. `SharedDependRelationId` is in ENR, so it can apply in this case.

src/backend/utils/misc/queryenvironment.c
time-change: i think it is not an issue. the number of catalog entries that can be in one transaction should basically be limited by the number of temp tables created in that transaction. 

if we were to switch to a hash table, we then would need to take the overhead of initializing and emptying the hash table for each transaction, which could get expensive, even for small transactions with only one temp table create. 

we can also check flame graph during validation, since we have added many new temp table features this time around.

src/backend/utils/misc/queryenvironment.c
```diff
@@ -241,6 +246,83 @@ bool has_existing_enr_relations()
...
+	currentQueryEnv->savedCatcacheMessages = lappend(currentQueryEnv->savedCatcacheMessages, msg);
+}
+
+/* Clear any saved catcache messages at end of xact. */
+void ClearSavedCatcacheMessages()
+{
+	if (!currentQueryEnv)
+		return;
+
+	list_free_deep(currentQueryEnv->savedCatcacheMessages);
```
surendravishnoi, line 268: let us add a NULL check for savedCatcacheMessages to be extra safe

src/backend/utils/misc/queryenvironment.c
```diff
@@ -86,6 +87,7 @@ create_queryEnv2(MemoryContext cxt, bool top_level)
 		queryEnv = topLevelQueryEnv;
 		queryEnv->namedRelList = NIL;
 		queryEnv->dropped_namedRelList = NIL;
+		queryEnv->savedCatcacheMessages = NIL;
```
surendravishnoi, line 90: how are we making sure that savedCatcacheMessages is NULL initialized in all cases? do we crate queryEnv with all fields set to NULL?

src/backend/utils/cache/inval.c
```diff
@@ -1063,6 +1087,7 @@ AtEOXact_Inval(bool isCommit)
 
 	/* Need not free anything explicitly */
 	transInvalInfo = NULL;
+	ClearSavedCatcacheMessages();
```
surendravishnoi, line 1090: can we do this via hook? Just to be extra safe so that in case we have some issue with savedCatcacheMessages memory allocation, it does not leak outside babelfish

src/backend/utils/misc/queryenvironment.c
```diff
@@ -151,6 +153,9 @@ void remove_queryEnv() {
 	list_free(currentQueryEnv->dropped_namedRelList);
 	currentQueryEnv->dropped_namedRelList = NIL;
 
+	list_free(currentQueryEnv->savedCatcacheMessages);
```
surendravishnoi, line 156: let us put all babelfish related cleanup in a hook

src/backend/utils/misc/queryenvironment.c
```diff
@@ -241,6 +246,83 @@ bool has_existing_enr_relations()
...
+/*
+ * SIMessageIsForTempTable
+ * 
+ * Determine whether the msg sent is for a temp table. 
+ * TSQL style temp tables do not need to add messages to the
+ * SI queue, as catalog changes are all session-local.
+ * 
+ * See LocalExecuteInvalidationMessage
+ */
+bool SIMessageIsForTempTable(const SharedInvalidationMessage *msg)
```
surendravishnoi, line 281: can this be a hook as well?

src/backend/utils/misc/queryenvironment.c
```diff
@@ -57,6 +57,7 @@ struct QueryEnvironment
 {
 	List	   *namedRelList;
 	List	   *dropped_namedRelList;
+	List 	   *savedCatcacheMessages;
```
surendravishnoi, line 60: in general, do we have clean separation of ENR code specific to babelfish and ENR code for PG?

src/backend/utils/misc/queryenvironment.c
time-change: queryEnv is either created with palloc0 through `create_queryEnv`, which sets all fields to null, or here in `create_queryEnv2`, where we specify it explicitly.

src/backend/utils/misc/queryenvironment.c
time-change: re this and above comment about separating ENR babelfish and PG logic - let's do this in a separate PR. i think it would bloat this change too much. there are a few things here we can clean up at once, like moving many of the BBF specific functions to hooks, cleaning up the hook file for all of the temp table functions, normalizing naming, and as you mention, making the queryEnv initialization and cleanup into BBF specific hooks. 

likely for the same reasons mentioned previously, it can't be in a hook. but we can put it in a separate function, and if necessary gate it behind some sort of GUC or dialect check.

src/backend/utils/cache/inval.c
time-change: this is not completely straightforward to add as a hook. because queryEnvironment is defined in `queryEnvironment.c`, we can't really reference internal members of the struct outside `queryenvironment.c`. it can work if we expose it to the header file, but that also means altering PG code and making the struct non private.

surendravishnoi (approved):
We should raise a separate PR to take care of ENR cod separation between BBF and PG. Also, add code in hooks wherever possible.

PR #433, opened 2024-08-29
Re-enable "Fix concurrency issues with ENR temp tables."

Sairakan:
### Description

https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/commit/5140f79a8ebbf37e0eea6527532f77cf7fa4d0c7 had a bug which caused a crash when aborting nested transactions with table variables due to missing a case when a lock was being acquired during table creation. Fix that case by adding a proper check for ENR-ness, and skip the lock acquisition in that case.
 
### Issues Resolved

BABEL-4913, BABEL-4892
 
### Check List

### DIFF
--- a/src/backend/catalog/heap.c
+++ b/src/backend/catalog/heap.c
@@ -1318,7 +1318,8 @@ heap_create_with_catalog(const char *relname,
 	 * it doesn't hurt to hold AccessExclusiveLock.  Do it here so callers
 	 * can't accidentally vary in their lock mode or acquisition timing.
 	 */
-	LockRelationOid(relid, AccessExclusiveLock);
+	if (!is_enr)
+		LockRelationOid(relid, AccessExclusiveLock);
 
 	/*
 	 * Determine the relation's initial permissions.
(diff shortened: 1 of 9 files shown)


src/backend/catalog/heap.c
```diff
@@ -1318,7 +1318,8 @@ heap_create_with_catalog(const char *relname,
 	 * it doesn't hurt to hold AccessExclusiveLock.  Do it here so callers
 	 * can't accidentally vary in their lock mode or acquisition timing.
 	 */
-	LockRelationOid(relid, AccessExclusiveLock);
+	if (!is_enr)
+		LockRelationOid(relid, AccessExclusiveLock);
```
Sairakan, line 1322: This is the case that was missed in the original PR

time-change (approved):

src/backend/catalog/heap.c
lejaokri: Thanks! quick question, did the original PR have all tests passing before we merged? How come we didnt hit the issue before merge?

Sairakan: Yeah, the original PR had all tests passing. I'm not sure why the crash didn't appear at that time.

lejaokri: approving since Jason confirmed the crash was hit during table_variable_xact_nested test, which confirmed passes with this fix, and crashes without it, hence we have a coverage with the new fix.

lejaokri (approved):

PR #448, opened 2024-09-25
Support View usage for PIVOT operator (Deparsed psql-text) BABEL3

RIC06X:
### Description

Support create/select/drop view on stmt with pivot operator. 

We switched to deparsed psql-text approach for VIEW/JOIN/CTE, all pivot metadata will be passed to bbf_pivot function arguments, so we removed all funcCall/funcExpr context field related code.

Extension PR:

### Issues Resolved

Task: BABEL-4673
 
### Check List

### DIFF
--- a/src/backend/utils/adt/ruleutils.c
+++ b/src/backend/utils/adt/ruleutils.c
@@ -11897,7 +11897,7 @@ generate_relation_name(Oid relid, List *namespaces)
 	if (!need_qual)
 		need_qual = !RelationIsVisible(relid);
 
-	if (need_qual)
+	if (need_qual || sql_dialect == SQL_DIALECT_TSQL)
 		nspname = get_namespace_name_or_temp(reltup->relnamespace);
 	else
 		nspname = NULL;
--- a/src/backend/executor/execSRF.c
+++ b/src/backend/executor/execSRF.c
@@ -33,9 +33,6 @@
 #include "utils/memutils.h"
 #include "utils/typcache.h"
 
-/* Hook to set TSQL pivot data to fcinfo */
-pass_pivot_data_to_fcinfo_hook_type pass_pivot_data_to_fcinfo_hook = NULL;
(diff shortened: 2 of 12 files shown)


src/backend/utils/adt/ruleutils.c
```diff
@@ -11897,7 +11897,7 @@ generate_relation_name(Oid relid, List *namespaces)
 	if (!need_qual)
 		need_qual = !RelationIsVisible(relid);
 
-	if (need_qual)
+	if (need_qual || sql_dialect == SQL_DIALECT_TSQL)
```
xhfanhe, line 11900: Why do we need to change this?

xhfanhe: Yanjie's reply:

This line is to fix https://jira.rds.a2z.com/browse/BABEL-5284:
During pg_get_querydef (pg deparse function), PG will call `generate_relation_name(Oid relid, List *namespaces)` to generate rel name. PG will skip table schema name when the calling env is using the same schema name. PG uses this function only during pg_dump, and pg_get_ruledef (when check view definition). So for pivot view, when the current schema name is equal to (search path includes current schema name) the relation's schema name in the view definition,  PG will skip the schema name in the view definition, and this will result table not found issue.
The fix is also straightforward, we can add a sql_dialect check, when it is tsql, we always produce fully qualified name. I tested in my local environment and it worked great.
```
	/* Otherwise, qualify the name if not visible in search path */
	if (!need_qual)
		need_qual = !RelationIsVisible(relid);
```
```
	if (need_qual || sql_dialect == SQL_DIALECT_TSQL)
		nspname = get_namespace_name_or_temp(reltup->relnamespace);
	else
		nspname = NULL;
```
fix is in src/backend/utils/adt/ruleutils.c:12110

forestkeeper (approved):
LGTM

PR #449, opened 2024-09-25
Fix crash when trying to alter temp table column type.

Sairakan:
### Description
In cases where a temp table column is being altered so that only the size changes (e.g. char(20) -> char(5)), the code path needs to drop the pg_depend entry from the normal catalog table (to be later recreated using the new target type). However, for ENR temp tables all catalog tuples are stored in currentQueryEnv and not actually in the catalogs themselves. Thus, the code would not be able to find the tuple for deletion, leading to an assertion error. Fix the issue by using ENRDropTuple() instead of CatalogTupleDelete() so that the code drops the tuple in the correct place.

Backport of https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/442

Note that there are additional changes compared to the original commit from https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/442 to allow this change to work with pg15, as there are commits which are present in pg16 which have not been backported to pg15 which are required for this change to work properly.
 
### Issues Resolved

BABEL-5273
 
### Check List

### DIFF
--- a/src/backend/commands/cluster.c
+++ b/src/backend/commands/cluster.c
@@ -734,8 +734,14 @@ make_new_heap(Oid OIDOldHeap, Oid NewTableSpace, Oid NewAccessMethod,
 	 * a shared rel.  However, we do make the new heap mapped if the source is
 	 * mapped.  This simplifies swap_relation_files, and is absolutely
 	 * necessary for rebuilding pg_class, for reasons explained there.
+	 *
+	 * We also must ensure that these temp tables are properly named in TSQL
+	 * so that the metadata is properly cleaned up after in this function.
 	 */
-	snprintf(NewHeapName, sizeof(NewHeapName), "pg_temp_%u", OIDOldHeap);
+	if (sql_dialect == SQL_DIALECT_TSQL && relpersistence == RELPERSISTENCE_TEMP && get_ENR_withoid(currentQueryEnv, OIDOldHeap, ENR_TSQL_TEMP))
+		snprintf(NewHeapName, sizeof(NewHeapName), "#pg_temp_%u", OIDOldHeap);
+	else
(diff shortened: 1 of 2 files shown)


src/backend/commands/cluster.c
```diff
@@ -734,8 +734,14 @@ make_new_heap(Oid OIDOldHeap, Oid NewTableSpace, Oid NewAccessMethod,
 	 * a shared rel.  However, we do make the new heap mapped if the source is
 	 * mapped.  This simplifies swap_relation_files, and is absolutely
 	 * necessary for rebuilding pg_class, for reasons explained there.
+	 *
+	 * We also must ensure that these temp tables are properly named in TSQL
+	 * so that the metadata is properly cleaned up after in this function.
 	 */
-	snprintf(NewHeapName, sizeof(NewHeapName), "pg_temp_%u", OIDOldHeap);
+	if (sql_dialect == SQL_DIALECT_TSQL && relpersistence == RELPERSISTENCE_TEMP && get_ENR_withoid(currentQueryEnv, OIDOldHeap, ENR_TSQL_TEMP))
```
time-change, line 741: can we be sure this change on its own has no side effects? Since it was packaged as part of a bigger change in rel16, and Rel16 also has other commits that don't exist on Rel15.

Sairakan: As far as I can tell, there are no side effects of this change. All JDBC tests pass, and it still has the intended benefit of allowing ALTER COLUMN <type> queries to go through. In terms of static analysis, I think this looks like a safe change to make - all we are doing is using a slightly different name for the generated temp table, which is still valid for both temp and permanent table naming conventions anyways.

time-change (approved):

2jungkook (approved):

PR #450, opened 2024-09-25
 Manually delete physical files for ENR entries when leaving scope.

Sairakan:
### Description

Second attempt at https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/2930 (which was reverted). This time, we completely ignore the pending deletes list, and instead manually trigger file deletion for each of the ENR entries we are dropping during ENRDropTempTables(). This ensures that we don't need to mess with the pendingDeletes list at all, while still properly deleting the physical storage for ENR entries when they are no longer in scope (such as at the end of a stored procedure call, explicit transaction, or batch).
 
### Issues Resolved

BABEL-4751
 
### Check List

### DIFF
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -42,6 +42,7 @@
 #include "catalog/pg_shdepend.h"
 #include "catalog/pg_index_d.h"
 #include "parser/parser.h"      /* only needed for GUC variables */
+#include "storage/smgr.h"
 #include "utils/inval.h"
 #include "utils/guc.h"
 #include "utils/syscache.h"
@@ -1549,9 +1550,12 @@ bool useTempOidBufferForOid(Oid relId)
 void
 ENRDropTempTables(QueryEnvironment *queryEnv)
 {
-	ListCell   *lc = NULL;
-	ObjectAddress object;
-	ObjectAddresses *objects;
+	ListCell		*lc = NULL;
+	ObjectAddress	object;
+	ObjectAddresses	*objects;
+	int				nrels = 0,
+					maxrels = 0;
+	SMgrRelation	*srels = NULL;
 
 	if (!queryEnv)
 		return;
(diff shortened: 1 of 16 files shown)


src/backend/utils/misc/queryenvironment.c
```diff
@@ -1572,6 +1584,10 @@ ENRDropTempTables(QueryEnvironment *queryEnv)
 		object.objectSubId = 0;
 		object.objectId = enr->md.reliddesc;
 		add_exact_object_address(&object, objects);
+		/*
+		 * Also register the physical file to be deleted at the end of scope.
+		 */
+		ENRDelete(enr->md.reliddesc);
```
surendravishnoi: Why not delete the file right here rather than adding it to delete list and later calling smgrDoPendingDeletes. We can write a new function to delete a file in SMGR

Sairakan: Let me look into this. The original reasoning behind doing it this way is that I am not sure how much of the storage deletion needs to be copied, vs just immediately calling smgr_close() on the relation

Sairakan: I've implemented this suggestion - now the files are immediately cleaned up during this function, which has the added benefit of not needing to modify any other function signatures or anything else in the code.

src/backend/utils/misc/queryenvironment.c
```diff
@@ -1572,6 +1578,45 @@ ENRDropTempTables(QueryEnvironment *queryEnv)
...
 		add_exact_object_address(&object, objects);
+		/*
+		 * Delete the physical storage for the relation.
+		 * See: smgrDoPendingDeletes()
+		 */
+		rel = relation_open(enr->md.reliddesc, AccessExclusiveLock);
+		srel = smgropen(rel->rd_locator, rel->rd_backend);
+
+		/* allocate the initial array, or extend it, if needed */
+		if (maxrels == 0)
```
surendravishnoi, line 1589: why are we doing this? we have single relationship, we should just pass pointer to that relation with count as 1 to smgrdounlinkall

Sairakan: We do not necessarily only have one relation - there can be multiple table variables and temp tables defined in the same scope, which would all be cleaned up at once by this function.

surendravishnoi (approved):

lejaokri (approved):

src/backend/utils/misc/queryenvironment.c
Sairakan: Correction: currently table variables are being cleaned up prior to this function during pltsql_clean_table_variables(), which manually executes a DROP TABLE <tblname> statement to drop the table variable. Since that function only registers the physical file for deletion at transaction rather than immediately, we probably should remove that function and let this function do both table variable and temp table cleanup, but that can be done in a later commit.

TLDR: currently only temp tables in scope are being cleaned up here, and not table variables, which are handled separately.

PR #451, opened 2024-09-25
Clean up temp table naming, introduce new functions for readability

time-change:
### Description

BABEL-5128

- Reorganize `queryenvironment.c` to have better separation of different families of functions (ENR catalog search/insertion, cleanup, invalidation messages, rollback handling). 
- Added static function declarations at top of queryenvironment.c for better organization. 
- Refactored some function names to have consistent naming
- Introduced new functions to clean up commonly used temp table checks across engine code

Note: `tsql_setup_queryEnv` and `tsql_cleanup_queryEnv` cannot be moved to hooks, since the internals of the QueryEnvironment struct are not visible outside of `queryenvironment.c`. 
 
### Check List


src/backend/utils/misc/queryenvironment.c
```diff
@@ -384,6 +285,16 @@ get_ENR_withoid(QueryEnvironment *queryEnv, Oid id, EphemeralNameRelationType ty
 	return NULL;
 }
 
+/*
+ * This is a wrapper function to get_ENR_withoid with the commonly 
+ * provided arguments of currentQueryEnv and ENR_TSQL_TEMP.
+ */
+EphemeralNamedRelation
+OidBelongsToENRTempTable(Oid id)
+{
```
lejaokri, line 295: Do we really need this? seems like it's a one line function so we can just call `get_ENR_withoid` directly.
ALso, the function name suggests it's supposed to return true/false but it's returning a `EphemeralNamedRelation`

src/backend/access/heap/heapam_visibility.c
```diff
@@ -468,7 +469,7 @@ HeapTupleSatisfiesUpdate(Relation relation, HeapTuple htup, CommandId curcid,
 	HeapTupleHeader tuple = htup->t_data;
 
 	/* See HeapTupleSatisfiesVisibility why */
-	if (sql_dialect == SQL_DIALECT_TSQL && RelationIsBBFTableVariable(relation))
+	if (IsTsqlTableVariable(relation))
```
lejaokri, line 472: nice - thanks for doing this.

src/backend/utils/misc/queryenvironment.c
```diff
@@ -1524,113 +1402,282 @@ void ENRRollbackSubtransaction(SubTransactionId subid, QueryEnvironment *queryEn
...
 
-		if (ret) {
-			HeapTuple htup;
-			MemoryContext oldcxt;
+/* Simple wrapper for Table Variable checks */
+bool IsTsqlTableVariable(Relation rel)
+{
+	return sql_dialect == SQL_DIALECT_TSQL 
+		&& rel
+		&& RelationIsBBFTableVariable(rel);
```
lejaokri: you can inline `RelationIsBBFTableVariable` here so we will always use `IsTsqlTableVariable` and get rid of `RelationIsBBFTableVariable`

src/backend/utils/misc/queryenvironment.c
```diff
@@ -1524,113 +1402,282 @@ void ENRRollbackSubtransaction(SubTransactionId subid, QueryEnvironment *queryEn
...
-		EphemeralNamedRelation enr = (EphemeralNamedRelation) lfirst(lc);
+		/*
+		* We skip invalidating IndexRelationId on drop (ie committed table, index created in transaction) 
+		* because it will be taken care of when the index itself is dropped - otherwise we risk
+		* throwing an error because the entry is already wiped away.
+		*/
+		
+		Form_pg_index idx_form = (Form_pg_index) GETSTRUCT(uncommitted_tup->tup);
+		EphemeralNamedRelation tmp_enr = get_ENR_withoid(queryEnv, idx_form->indrelid, ENR_TSQL_TEMP);
+		if (tmp_enr)
```
lejaokri: nit - we dont use `tmp_enr` anywhere else, you can just call `get_ENR_withoid()` directly.

src/backend/utils/misc/queryenvironment.c
time-change: there were a few places in the code that it wasn't immediately apparent we could replace `RelationIsBBFTableVariable` but I'll look into it a bit more

src/backend/utils/misc/queryenvironment.c
time-change: My thinking was just to add this for readability, since `get_ENR_withoid` had some arguments that never change. For instance, it's almost always called with `ENR_TSQL_TEMP` as the third argument. Maybe we can just rewrite the base function instead.

lejaokri: I see your point. It makes sense to rename `OidBelongsToENRTempTable` to something more descriptive then because I initially thought the name would return true/false. Something like `GetENRTempTableWithOid(Oid oid)` or something better (im not good in namings) :)

Sairakan: +1 on renaming this function to something more accurate to its behavior.

time-change: that name LGTM! refactored.

src/backend/utils/misc/queryenvironment.c
time-change: done

src/backend/access/heap/heapam_visibility.c
```diff
@@ -468,7 +469,7 @@ HeapTupleSatisfiesUpdate(Relation relation, HeapTuple htup, CommandId curcid,
 	HeapTupleHeader tuple = htup->t_data;
 
 	/* See HeapTupleSatisfiesVisibility why */
-	if (sql_dialect == SQL_DIALECT_TSQL && RelationIsBBFTableVariable(relation))
```
surendravishnoi, line 471: also see which header we can remove (parser.h?)

src/backend/utils/misc/queryenvironment.c
```diff
@@ -67,6 +67,15 @@ struct QueryEnvironment topLevelQueryEnvData;
 struct QueryEnvironment *topLevelQueryEnv = &topLevelQueryEnvData;
 struct QueryEnvironment *currentQueryEnv = NULL;
 
+static void tsql_setup_queryEnv(QueryEnvironment *queryEnv);
```
surendravishnoi, line 70: will it make sense to put BBF changes in a separate file? if not possible due to static functions etc we can use wrappers.
Another possibility is to move most of the code to BBF extension

src/backend/utils/misc/queryenvironment.c
```diff
@@ -86,9 +95,7 @@ create_queryEnv2(MemoryContext cxt, bool top_level)
 
 	if (top_level) {
 		queryEnv = topLevelQueryEnv;
-		queryEnv->namedRelList = NIL;
-		queryEnv->dropped_namedRelList = NIL;
-		queryEnv->savedCatcacheMessages = NIL;
+		tsql_setup_queryEnv(queryEnv);
```
surendravishnoi, line 98: Hooks can help us reduce regression chances. We should never call a TSQL function in APG instance. I think it is worth having hooks so that we never execute these code path outside BBF env.

timchang514: I generally agree that we shouldn't be calling tsql functions in APG code. The problem with this one in particular though, is that the members of QueryEnvironment are not visible outside the file, so we would have to make the queryenv struct public to be able to move that into a hook. An alternative would be making `tsql_setup_queryEnv` a public function that we invoke from a hook in the extension, but this to me seems counterintuitive to have a public function for private struct invoked through a hook. 

Other types of checks like sql_dialect might not be sufficient either because we can enter psql dialect in Babelfish code. 

At least for this case, I think we can add a comment saying to not introduce any code here that may have side effects given that the function could be called from APG as well.

src/backend/utils/misc/queryenvironment.c
time-change: the main issue with moving BBF changes to a separate file is that the QueryEnvironment struct is meant to be private to the `queryenvironment.c` file. Moving everything to the extension would be possible if we move the QueryEnvironment struct into `queryenvironment.h`, making it publicly visible, but it has its own risks if community decides to refactor that code in some way.

lejaokri:
I dont have anymore comments. approving.

src/backend/utils/misc/queryenvironment.c
lejaokri: agreed to Surendra here btw.

lejaokri (approved):

forestkeeper (approved):

src/backend/utils/misc/queryenvironment.c
```diff
@@ -773,193 +684,62 @@ bool ENRgetSystableScan(Relation rel, Oid indexId, int nkeys, ScanKey key, List
...
-						if (tup->oid == entry->objid)
-							return tmp_enr;
-					}
-				}
-				break;
+/*
+ * Add tuple to an ENR. It assumes that an ENR entry has been created with
+ * the relation name and relation oid.
+ */
+bool ENRAddTuple(Relation rel, HeapTuple tup)
```
forestkeeper, line 692: I have the same thought with Surendra, can we add more hook impl for functions like this one ?

PR #470, opened 2024-11-12
Display unmapped schema names in error messages

chxwang:
### Description

Previously, Babelfish displays the mapped schema names in error messages. This commit imports a hook function remove_db_name_in_schema and uses it to restore thr unmapped name. Original unmapped name will be displayed with this commit.
 
### Issues Resolved

[BABEL-2961](https://jira.rds.a2z.com/browse/BABEL-2961)
 
### Check List

### DIFF
--- a/src/backend/parser/parse_func.c
+++ b/src/backend/parser/parse_func.c
@@ -2059,11 +2059,11 @@ funcname_signature_string(const char *funcname, int nargs,
 	int			numposargs;
 	ListCell   *lc;
 	int			i;
+	const char *ptr;
 
 	initStringInfo(&argbuf);
 
 	/* Check if there is schema in funcname */
-	char * ptr = funcname;
 	while (ptr) {
 		if (*ptr == '.') {
             break;
--- a/src/backend/parser/parse_func.c
+++ b/src/backend/parser/parse_func.c
@@ -2063,14 +2063,6 @@ funcname_signature_string(const char *funcname, int nargs,
 
 	initStringInfo(&argbuf);
 
-	/* Check if there is schema in funcname */
-	while (ptr) {
-		if (*ptr == '.') {
-            break;
-        }
-        ptr++;
-	}
-
 	if (sql_dialect == SQL_DIALECT_TSQL && remove_db_name_in_schema_hook && ptr) {
 		funcname = (*remove_db_name_in_schema_hook)(funcname);
 	}
--- a/src/backend/parser/parse_func.c
+++ b/src/backend/parser/parse_func.c
@@ -2062,6 +2062,10 @@ funcname_signature_string(const char *funcname, int nargs,
 
 	initStringInfo(&argbuf);
 
+	if (sql_dialect == SQL_DIALECT_TSQL && remove_db_name_in_schema_hook) {
+		funcname = (*remove_db_name_in_schema_hook)(funcname);
+	}
+
 	appendStringInfo(&argbuf, "%s(", funcname);
 
 	numposargs = nargs - list_length(argnames);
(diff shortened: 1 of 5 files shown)


src/backend/parser/parse_func.c
```diff
@@ -2062,6 +2062,10 @@ funcname_signature_string(const char *funcname, int nargs,
 
 	initStringInfo(&argbuf);
 
+	if (sql_dialect == SQL_DIALECT_TSQL && remove_db_name_in_schema_hook) {
+		funcname = (*remove_db_name_in_schema_hook)(funcname);
```
xhfanhe: This is dealing with funcname, is schema name mapping needed here? 
I'm guessing it's something like `db_schema.func()` -> `schema.func()`, but is there always a schema involved?

chxwang: No, thanks for pointing this out. I'll add a check to determine if there is schema included in funcname.

src/backend/parser/parse_func.c
```diff
@@ -2059,9 +2059,22 @@ funcname_signature_string(const char *funcname, int nargs,
...
 
 	initStringInfo(&argbuf);
 
+	/* Check if there is schema in funcname */
+	while (ptr) {
+		if (*ptr == '.') {
+            break;
+        }
+        ptr++;
+	}
```
xhfanhe: To minimize change in engine code, please move this logic into hook func. 
Also, I searched in BBF code base and found a function `split_object_name` that splits multiple-part name into an array. There should be some examples of using it in BBF repo.
In test cases, please also consider delimiters, upper/lower/mixed cases.

src/backend/parser/parse_relation.c
```diff
@@ -1453,10 +1453,15 @@ parserOpenTable(ParseState *pstate, const RangeVar *relation, int lockmode)
...
+		{
+			const char * schema_name = relation->schemaname;
+			if (sql_dialect == SQL_DIALECT_TSQL && remove_db_name_in_schema_hook)
+				schema_name = (*remove_db_name_in_schema_hook)(schema_name);
 			ereport(ERROR,
 					(errcode(ERRCODE_UNDEFINED_TABLE),
 					 errmsg("relation \"%s.%s\" does not exist",
-							relation->schemaname, relation->relname)));
+							schema_name, relation->relname)));
+		}
```
RIC06X, line 1468: Here I think it is better to have a separate if stmt with erreport,  it is not recommended to to directly modify PG code.
You can do something like this: 
```
if (TSQL and hook function is defined)
  call hook function (within hook function, you raise the error)
else
 original ereport.
```
Also for the schema_name, can you simply use relation->schemaname?

src/backend/parser/parse_func.c
xhfanhe: Is this checked in new code?

src/backend/parser/parse_relation.c
time-change: +1

RIC06X:
LGTM

src/backend/parser/parse_func.c
xhfanhe: Would you please check if `funcname` can possibly don’t contain schema name? This is the only place that doesn’t explicitly state that there must be a schema name.
- If it’s guaranteed to have schema name, then your current implementation is okay, may be better to mention that it’s caller’s responsibility to make sure that input object_name contains a schema name
- If it’s possible to have one-part funcname without schema, then you may want to add special handling for this scenario

xhfanhe (approved):
LGTM

PR #474, opened 2024-11-13
Fix drop database issue after MVU to PG17

sumitj824:
### Description

After performing MVU to PG17, we weren't able to drop the existing database because a new dependency was getting added between the `dbo` system role and physical babelfish database. This dependency is getting introduced when running the TSQL upgrade script babelfishpg_tsql--3.4.0--4.0.0.sql, which grants permissions to the bbf_role_admin role. The new dependency is expected and is a result of a community change https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/commit/514a3de8b0e63aaa9e126c65b5b68bea20fbdde3. However, even though there is a function RemoveRoleFromInitPriv() to clean up these dependencies when executing DROP OWNED BY commands, it is not working correctly for system roles. This is because the function assumes the grantor of rights is always the same as the owner, which is not the case for system roles, where the grantor should be sysadmin.

This commit made changes to use sysadmin as grantor when generating ACLs for physical babelfish database while removing role from initial privileges.
 

Signed-off-by: Sumit Jaiswal <sumiji@amazon.com>

Related Task: BABEL-5410
 
### Check List

### DIFF
--- a/src/backend/catalog/aclchk.c
+++ b/src/backend/catalog/aclchk.c
@@ -5008,13 +5008,15 @@ RemoveRoleFromInitPriv(Oid roleid, Oid classid, Oid objid, int32 objsubid)
 	if (old_acl != NULL)
 	{
 		Oid sysadminOid;
+		const char *babelfish_db_name;
 
 		/*
 		 * For babelfish database, grantor will be sysadmin instead of object owner.
 		 */
 		if (bbf_get_sysadmin_oid_hook &&
 			classid == DatabaseRelationId &&
-			objid == get_database_oid(GetConfigOption("babelfishpg_tsql.database_name", true, false), true) &&
+			(babelfish_db_name = GetConfigOption("babelfishpg_tsql.database_name", true, false)) &&
+			objid == get_database_oid(babelfish_db_name, true) &&
 			is_member_of_role(GetUserId(), sysadminOid = (*bbf_get_sysadmin_oid_hook)()))
 		{
 			grantorId = sysadminOid;
--- a/src/backend/catalog/aclchk.c
+++ b/src/backend/catalog/aclchk.c
@@ -5010,10 +5010,11 @@ RemoveRoleFromInitPriv(Oid roleid, Oid classid, Oid objid, int32 objsubid)
 		Oid sysadminOid;
 
 		/*
-		 * For TSQL system roles, grantor will be sysadmin instead of owner.
+		 * For babelfish database, grantor will be sysadmin instead of object owner.
 		 */
 		if (bbf_get_sysadmin_oid_hook &&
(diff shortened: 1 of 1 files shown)


src/backend/catalog/aclchk.c
```diff
@@ -5005,14 +5006,30 @@ RemoveRoleFromInitPriv(Oid roleid, Oid classid, Oid objid, int32 objsubid)
...
+
+		/*
+		 * For TSQL system roles, grantor will be sysadmin instead of owner.
+		 */
+		if (bbf_get_sysadmin_oid_hook &&
+			classid == DatabaseRelationId &&
+			is_member_of_role(GetUserId(), sysadminOid = (*bbf_get_sysadmin_oid_hook)()))
+		{
+			grantorId = sysadminOid;
+		}
```
HarshLunagariya: 1. Will Grantor always be sysadmin in case of Babelfish? At the time of inserting the init_privs, how are we deciding the grantor? If we see possibility that grantor being diff. from sysadmin then we should utilise `select_best_grantor()`
2. Will it by triggered for other physical databases? Ideally it shouldn't. If it is then, I would suggest add check for babelfish physical db oid comparison if possible.
3. We should check whether it is babelfish role or not as well.

HarshLunagariya (changes requested):

src/backend/catalog/aclchk.c
sumitj824: Discussed it offline and added check for babelfish database.

src/backend/catalog/aclchk.c
```diff
@@ -5005,14 +5006,31 @@ RemoveRoleFromInitPriv(Oid roleid, Oid classid, Oid objid, int32 objsubid)
...
 	if (old_acl != NULL)
+	{
+		Oid sysadminOid;
+
+		/*
+		 * For babelfish database, grantor will be sysadmin instead of object owner.
+		 */
+		if (bbf_get_sysadmin_oid_hook &&
+			classid == DatabaseRelationId &&
+			objid == get_database_oid(GetConfigOption("babelfishpg_tsql.database_name", true, false), true) &&
```
HarshLunagariya: What if `GetConfigOption("babelfishpg_tsql.database_name", true, false)` returns NULL?
Can you please include some manual testing from PG endpoint in the description which makes sure that it is not interfering when hitting this code path from different physical database?

HarshLunagariya (changes requested):

src/backend/catalog/aclchk.c
sumitj824: added check for that as well.

HarshLunagariya (approved):

shardgupta (approved):

PR #487, opened 2024-11-27
Retain result type after UNION as base type in T-SQL

hopebo:
### Description

When a base type such as `money` is UNIONed with another integer type, the result type will be `fixeddecimal` instead of `money`, which is inconsistent with SQL Server's behaviour. And could also cause some characters not displaying well due to inappropriate result type.
 
### Issues Resolved

The root cause of the issue is Postgres will consider the base type of domain type as the result type. In Babelfish, `money` is implemented as a domain type of `fixeddecimal`. Actually, `money` is a base type in T-SQL, thus Babelfish should retain `money` type after UNION.

After a final type has determined by the optimizer, check if it's a base type in T-SQL. If it is, try to use the domain type as result type if the domain type has higher precision than other types.

BABEL-5242
 
### Check List

### DIFF
--- a/src/backend/parser/parse_coerce.c
+++ b/src/backend/parser/parse_coerce.c
@@ -35,6 +35,7 @@
 
 find_coercion_pathway_hook_type find_coercion_pathway_hook = NULL;
 determine_datatype_precedence_hook_type determine_datatype_precedence_hook = NULL;
+is_tsql_base_datatype_hook_type is_tsql_base_datatype_hook = NULL;
 coerce_string_literal_hook_type coerce_string_literal_hook = NULL;
 validate_implicit_conversion_from_string_literal_hook_type validate_implicit_conversion_from_string_literal_hook = NULL;
 select_common_type_hook_type select_common_type_hook = NULL;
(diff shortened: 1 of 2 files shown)


src/backend/parser/parse_coerce.c
```diff
@@ -1507,32 +1508,80 @@ select_common_type(ParseState *pstate, List *exprs, const char *context,
...
-	 * probably end up needing a runtime coercion from UNKNOWN to something
-	 * else, and we usually won't have it.  We need to coerce the unknown
-	 * literals while they are still literals, so a decision has to be made
-	 * now.
-	 */
-	if (ptype == UNKNOWNOID && sql_dialect == SQL_DIALECT_TSQL)
+
+	if (sql_dialect == SQL_DIALECT_TSQL)
 	{
-		bool all_nullconst = true;
```
forestkeeper: Is it possible to put all those logic into single hook and move to babelfish code only ?

hopebo: I've optimized the codes and nest to the original loop. And it works together with the following codes outside of the loop. So it's not easy to make them in a single hook.

src/backend/parser/parse_coerce.c
```diff
@@ -1504,9 +1505,38 @@ select_common_type(ParseState *pstate, List *exprs, const char *context,
...
+			 * the result type, it could loss precision.
+			 * Here we don't need to update other variables since they are the
+			 * same.
+			 */
+			if (is_tsql_base_datatype_hook &&
+				(*is_tsql_base_datatype_hook)(exprType(nexpr)) &&
+				determine_datatype_precedence_hook &&
+				determine_datatype_precedence_hook(exprType(nexpr),
+												   exprType(pexpr)))
+				pexpr = nexpr;
```
Deepesh125, line 1524: why arent we updating ptype? because next iteration may need to make some decision

Deepesh125: second, why cant we fit this logic into select_common_type_hook?

Deepesh125 (changes requested):

src/backend/parser/parse_coerce.c
hopebo: Since in this branch, `ntype` is equal to `ptype`. So as `pcategory` and `pispreferred`. We change the `expr` for the information of domain type. This logic is combined with the following changes.

hopebo: >second, why cant we fit this logic into select_common_type_hook?

It is because a general fix for this issue should happens after the main process in `select_common_type`, in which the common type has been decided. The `select_common_type_hook` does a early return. And if we want to do it in this hook, we need to copy most codes in `select_common_type` to that hook.

Deepesh125 (approved):

PR #488, opened 2024-11-28
ExecProcessReturning is unexpectedly called twice for psql  request causing overflow error

hopebo:
### Description

In community postgresql, ExecProcessReturning is called once after the
main process of ExecInsert. But due to some merging errors related to
Babelfish and T-SQL, this function is wrongly called twice for psql
requests. At the first call, the xmax is not initialized causing
unexpected error raising.

The first intention of these pieces of codes is to call it in the
first place only for T-SQL(CR-52326143). And then got wrongly
cherry-picked in CR-57250041. Finally, the condition check gets
removed in CR-65026879. We should keep the first intention of the
codes, which also makes it consistent with the behaviour in
ExecUpdate.

Task: BABEL-5343
 
### Issues Resolved

Unexpected error raised for psql requests

```
CREATE TABLE table1(id SERIAL, col1 VARCHAR UNIQUE);
postgres=# INSERT INTO table1 VALUES (default, 'alma')  RETURNING xmax::text::int;
ERROR:  value "4294967295" is out of range for type integer
```
 
### Tests

All BabelExtension tests have passed in [https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3189](https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3189), except one upgrade test failed due to HTTP error.

Postgres regression tests in parallel schedule passed.

### Check List

### DIFF
--- a/src/test/regress/sql/babel_5343_returning.sql
+++ b/src/test/regress/sql/babel_5343_returning.sql
@@ -1,0 +1,5 @@
+CREATE TABLE toverflow(id SERIAL, col1 VARCHAR);
+INSERT INTO toverflow VALUES (default, 'hope') RETURNING xmax::text::int;
+INSERT INTO toverflow VALUES (default, 'jack') RETURNING xmax;
+SELECT * FROM toverflow ORDER BY id;
+DROP TABLE toverflow;
--- a/src/backend/executor/nodeModifyTable.c
+++ b/src/backend/executor/nodeModifyTable.c
@@ -823,7 +823,7 @@ ExecInsert(ModifyTableContext *context,
 	}
 
 	/* Process RETURNING if present */
-	if (resultRelInfo->ri_projectReturning)
+	if (resultRelInfo->ri_projectReturning && sql_dialect == SQL_DIALECT_TSQL)
 		result = ExecProcessReturning(resultRelInfo, slot, planSlot);
 
 	/* INSTEAD OF ROW INSERT Triggers */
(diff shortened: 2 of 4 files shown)


src/test/regress/sql/babel_5343_returning.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE toverflow(id SERIAL, col1 VARCHAR);
+INSERT INTO toverflow VALUES (default, 'hope') RETURNING xmax::text::int;
+INSERT INTO toverflow VALUES (default, 'jack') RETURNING xmax;
+SELECT * FROM toverflow;
+DROP TABLE toverflow;
```
lejaokri: not sure if it's worth creating a new test file just for this. Do we have any existing test file that alredy does INSERT RETURNING, if not, can probably add in insert testcase instead. 

Also, any chance we can test this code in BBF? I know we cannot really query for `xmax` etc in BBF, but wondering if there is a way to validate this in BBF

lejaokri: As discussed, since preference is to create a new test file, `babel_5343_returning` may not make sense so much in the community. perhaps call it `insert_returning.sql`?

hopebo: The function call for BBF should've been covered in the initial test cases. Also, there is no `RETURNING` or `xmax` supported for BBF.

Actually, `insert_returning.sql` sounds like the general test cases for the `INSERT ... RETURNING` feature. But this test case is only needed due to our inappropriate handles on the codes. I prefer a specific name. After we cherry-pick, it's more consistent with the naming of similar cases in the internal repository. It doesn't matter too much, since it's not the official community repository.

2jungkook (approved):

PR #500, opened 2024-12-13
Add sanity checks for ENR entries.

time-change:
### Description

Tests are running in https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3245

This change adds a number of sanity checks when ENR entries are added. Previously, there were no checks here which meant that we could run into cases we had not considered where an ENR entry could depend on a non-ENR entry, and vice-versa. When adding these tests, I discovered that there are still 3 on-disk catalogs that can be referenced by ENR entries:

- pg_collation
- pg_authid
- pg_opclass

These catalogs can't be trivially added to ENR, so I will create some followup tasks on properly handling these cases if possible.

Note that there are still some potential limitations. For example, the issue fixed in BABEL-4868 (see https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3122) would not have been caught by this change due to the way that dependency entries are created for functions as column defaults.

A way to truly guarantee that all corner cases could be caught is to implement a sort of check in Postgres' dependency code. This would be a pretty sizable change and have potential performance impact, since we would need to a) add infrastructure to allow lookups to all tuples in ENR by OID, and b) search through this structure every time we try to add a dependency. I've opted to not include that for now due to the reasons mentioned, but if there is interest in that we can have a separate discussion on it.
 
### Issues Resolved

BABEL-4776
 
### Check List

### DIFF
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -805,11 +805,12 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
 					{
 						Form_pg_class classForm = (Form_pg_class) GETSTRUCT(tup);
 
-						Assert(!classForm->relisshared);
-						Assert(classForm->relpersistence == RELPERSISTENCE_TEMP);
-						Assert(!classForm->relhastriggers);
-						Assert(!classForm->relrowsecurity);
-						Assert(!classForm->relispartition);
+						if ((classForm->relisshared) ||
+							(classForm->relpersistence != RELPERSISTENCE_TEMP) ||
+							(classForm->relhastriggers) ||
+							(classForm->relrowsecurity) ||
+							(classForm->relispartition))
+							elog(ERROR, "Invalid pg_class ENR entry.");
 					}
 				}
 				break;
@@ -832,8 +833,13 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
 								 * When adding entries to pg_depend, do an additional sanity check
 								 * to verify we aren't creating links to non-ENR catalogs.
 								 */
-								if (op == ENR_OP_ADD && (!IsENRRelationOid(tf1->classid, true) || !IsENRRelationOid(tf1->refclassid, true)))
-									elog(ERROR, "Unexpected catalog OID %d referenced in ENR.", tf1->classid);
+								if (op == ENR_OP_ADD)
+								{
+									if (!IsENRRelationOid(tf1->classid, true))
+										elog(ERROR, "Unexpected catalog OID %d referenced in ENR.", tf1->classid);
+									else if (!IsENRRelationOid(tf1->refclassid, true))
+										elog(ERROR, "Unexpected catalog OID %d referenced in ENR.", tf1->refclassid);
+								}
 
 								break;
 							}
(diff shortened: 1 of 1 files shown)


src/backend/utils/misc/queryenvironment.c
```diff
@@ -326,6 +351,37 @@ ENRMetadataGetTupDesc(EphemeralNamedRelationMetadata enrmd)
...
 }
 
+/* Simple check to see if the provided OID is the OID of an ENR'd catalog. */
+static bool IsENRRelationOid(Oid reloid, bool extended)
+{
+	/* This should always be a catalog relation we're checking. */
+	if (!IsCatalogRelationOid(reloid))
+		return false;
+
+	for (int i = 0; i < 11; i++)
```
lejaokri: I suggest defining a constant than hard coding 11. I probably wouldn't remember to update this if I had to add a new catalog in ENR. 

nit:
The name "IsENRRelationOid" is a bit inaccurate, I think. If I have a temp table in ENR, I would consider it an ENR relation. but this function checks for catalog oids only right? Perhaps call it something like "`IsCataloigOidENR()`" - idk if that's any better, im bad at namings :)

src/backend/utils/misc/queryenvironment.c
```diff
@@ -748,6 +794,23 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
...
+					 */
+					if ((op == ENR_OP_ADD || op == ENR_OP_DROP) && HeapTupleIsValid(tup))
+					{
+						Form_pg_class classForm = (Form_pg_class) GETSTRUCT(tup);
+
+						Assert(!classForm->relisshared);
+						Assert(classForm->relpersistence == RELPERSISTENCE_TEMP);
+						Assert(!classForm->relhastriggers);
+						Assert(!classForm->relrowsecurity);
+						Assert(!classForm->relispartition);
```
lejaokri: Are Asserts enabled in prod? If not, im curious how we would catch the issues in prod?

lejaokri (changes requested):

src/backend/utils/misc/queryenvironment.c
```diff
@@ -764,6 +830,19 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
 								tf1->objid == tf2->objid &&
 								tf1->objsubid == tf2->objsubid) {
 								lc = curlc;
+
+								/*
+								 * When adding entries to pg_depend, do an additional sanity check
+								 * to verify we aren't creating links to non-ENR catalogs.
+								 */
+								if (op == ENR_OP_ADD)
+								{
```
surendravishnoi, line 839: this adds value but will not detect cases where there is dependency between temp and main objects

src/backend/utils/misc/queryenvironment.c
```diff
@@ -748,6 +796,24 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
...
+					 * All of the below asserts should hold true for TSQL temp tables.
+					 *
+					 * This helps ensure that we don't have any dependencies pointing to
+					 * non-ENR catalogs.
+					 */
+					if ((op == ENR_OP_ADD || op == ENR_OP_DROP) && HeapTupleIsValid(tup))
+					{
+						Form_pg_class classForm = (Form_pg_class) GETSTRUCT(tup);
+
+						if ((classForm->relisshared) ||
```
surendravishnoi, line 810: what value these conditions add? for example temp tables cannot have row level security attributes. But what we really want to check is that any part of an ENR table should not be stored in PG catalog. It means that I cannot have an entry in PG catalog to store row level security information. Either I store that in ENR or whole table goes into PG catalog. But are these checks right candidates to validate above assumptions?

src/backend/utils/misc/queryenvironment.c
time-change: see above comment.

src/backend/utils/misc/queryenvironment.c
time-change: It's at least some amount of validation on the pg_class entry. You're correct that this doesn't check whether there is an entry for an ENR table in the PG catalog for row level security information. But that row is still tied to a pg_class entry with the flag set, so this will still catch that case and throw an error.

As I mentioned in the description of the PR: 

> Note that there are still some potential limitations. For example, the issue fixed in BABEL-4868 (see https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3122) would not have been caught by this change due to the way that dependency entries are created for functions as column defaults.
> 
> A way to truly guarantee that all corner cases could be caught is to implement a sort of check in Postgres' dependency code. This would be a pretty sizable change and have potential performance impact, since we would need to a) add infrastructure to allow lookups to all tuples in ENR by OID, and b) search through this structure every time we try to add a dependency. I've opted to not include that for now due to the reasons mentioned, but if there is interest in that we can have a separate discussion on it.

time-change: I will start a followup discussion re: the incompatible catalogs as well as the problems with checking for ENR entries in pg catalogs. I wouldn't consider either of those blockers for this change.

lejaokri (approved):
I dont have any more comments

surendravishnoi (approved):

PR #506, opened 2024-12-19
Add sanity checks for ENR entries.

time-change:
Cherry-pick of #3245 to 5_X branches. 

Description
Tests are running in https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3245

This change adds a number of sanity checks when ENR entries are added. Previously, there were no checks here which meant that we could run into cases we had not considered where an ENR entry could depend on a non-ENR entry, and vice-versa. When adding these tests, I discovered that there are still 3 on-disk catalogs that can be referenced by ENR entries:

pg_collation
pg_authid
pg_opclass
These catalogs can't be trivially added to ENR, so I will create some followup tasks on properly handling these cases if possible.

Note that there are still some potential limitations. For example, the issue fixed in BABEL-4868 (see https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3122) would not have been caught by this change due to the way that dependency entries are created for functions as column defaults.

A way to truly guarantee that all corner cases could be caught is to implement a sort of check in Postgres' dependency code. This would be a pretty sizable change and have potential performance impact, since we would need to a) add infrastructure to allow lookups to all tuples in ENR by OID, and b) search through this structure every time we try to add a dependency. I've opted to not include that for now due to the reasons mentioned, but if there is interest in that we can have a separate discussion on it.

Issues Resolved
BABEL-4776

### DIFF
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -32,6 +32,9 @@
 #include "catalog/catalog.h"
 #include "catalog/dependency.h"
 #include "catalog/indexing.h"
+#include "catalog/pg_authid.h"
+#include "catalog/pg_collation.h"
+#include "catalog/pg_opclass.h"
 #include "catalog/pg_constraint.h"
 #include "catalog/pg_statistic.h"
 #include "catalog/pg_statistic_ext.h"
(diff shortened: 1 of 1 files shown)


xhfanhe (approved):

xhfanhe:
16x change: https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/500

PR #507, opened 2024-12-20
Performance Improvement of create login in BBF

anju15bharti:
### Description

Extension PR : https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3291

In Major version 16 Babelfish introduced a new role bbf_admin_role which is a member of all the roles which created implicitly from TSQL. This increases the Load on any Grant/Revoke operations that happen since in BBF to "select the role through which it gets permission to current user to grant", is inefficient and loops over members of members to find the all members of the role even though the required role is find.
This commit optimizes the selection of role with is_admin_option true for babelfish by escaping the search for any TSQL roles conditionally.
Without the changes the time taken to create a login increases drastically and linearly:

Flame graph : 
kernel

The average time consumed by create login stmt(with 125 dbs and 125 logins present) : 12836 ms
with these changes we see the time taken to create login increasing linearly but not as drastically as it was before the changes:

Flame graph : 
perf-kernel-new-hook

The average time consumed by create login stmt(with 125 dbs and 125 logins present) : 49 ms

### Issues Resolved

BABEL-5349
 
### Check List

### DIFF
--- a/src/backend/utils/adt/acl.c
+++ b/src/backend/utils/adt/acl.c
@@ -5076,7 +5076,18 @@ roles_is_member_of(Oid roleid, enum RoleRecurseType type,
 			 */
 			if (otherid == admin_of && form->admin_option &&
 				OidIsValid(admin_of) && !OidIsValid(*admin_role))
-				*admin_role = memberid;
+				{
+					*admin_role = memberid;
+
+					/* 
+					 * Need not to iterate through all the members 
+					 * if admin_role is found.
+					 */
+					if (sql_dialect == SQL_DIALECT_TSQL)
+					{
+						return NULL;
+					}
+				}
 
 			/* If we're supposed to ignore non-heritable grants, do so. */
 			if (type == ROLERECURSE_PRIVS && !form->inherit_option)
--- a/src/backend/utils/adt/acl.c
+++ b/src/backend/utils/adt/acl.c
@@ -134,7 +134,7 @@ bbf_get_sysadmin_oid_hook_type bbf_get_sysadmin_oid_hook = NULL;
 get_bbf_admin_oid_hook_type get_bbf_admin_oid_hook = NULL;
 pltsql_get_object_owner_hook_type pltsql_get_object_owner_hook = NULL;
 is_bbf_db_ddladmin_operation_hook_type is_bbf_db_ddladmin_operation_hook = NULL;
-has_bbf_role_direct_membership_with_admin_true_hook_type has_bbf_role_direct_membership_with_admin_true_hook = NULL;
+bbf_check_member_has_direct_priv_to_grant_role_hook_type bbf_check_member_has_direct_priv_to_grant_role_hook = NULL;
 
 
 /*
@@ -5279,17 +5279,10 @@ is_admin_of_role(Oid member, Oid role)
 	if (member == role)
 		return false;
 
-	/*
-	 * Check if the given member is bbf_role_admin.
-         * If the member has the privilege to grant a given role through direct membership,
-         * returns the member with admin option set to true.
-	 */
-	if (sql_dialect == SQL_DIALECT_TSQL 
-		&& get_bbf_admin_oid_hook && member == (*get_bbf_admin_oid_hook)()
-			&& (has_bbf_role_direct_membership_with_admin_true_hook) 
-				&& (*has_bbf_role_direct_membership_with_admin_true_hook)(role))
+	if ((bbf_check_member_has_direct_priv_to_grant_role_hook) 
+		&& (*bbf_check_member_has_direct_priv_to_grant_role_hook)(member, role))
 	{
-		return member;
+		return true;
 	}
 
 	(void) roles_is_member_of(member, ROLERECURSE_MEMBERS, role, &admin_role);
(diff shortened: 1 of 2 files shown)


src/backend/utils/adt/acl.c
```diff
@@ -5076,7 +5076,18 @@ roles_is_member_of(Oid roleid, enum RoleRecurseType type,
...
+					*admin_role = memberid;
+
+					/* 
+					 * Need not to iterate through all the members 
+					 * if admin_role is found.
+					 */
+					if (sql_dialect == SQL_DIALECT_TSQL)
+					{
+						return NULL;
+					}
```
HarshLunagariya: I suggest to restrict this logic to some special roles with significant membership like 'sysadmin' and 'bbf_role_admin'.

Let's get more opinions from perf. experts @surendravishnoi @KushaalShroff

HarshLunagariya (changes requested):

src/backend/utils/adt/acl.c
```diff
@@ -5076,7 +5076,19 @@ roles_is_member_of(Oid roleid, enum RoleRecurseType type,
...
+					/* 
+					 * Need not to iterate through all the members 
+					 * if roleid is bbf_role_admin and admin_role is found.
+					 */
+					if (sql_dialect == SQL_DIALECT_TSQL &&
+						get_bbf_admin_oid_hook  && roleid == (*get_bbf_admin_oid_hook)())
+					{
+						return NULL;
+					}
+				}
```
tanscorpio7: What if we put a hook in `is_admin_of_role()` instead.
If member to be checked is bbf_role_admin then do a direct check for admin membership using syscache, if direct admin membership does not exists fall back to `roles_is_member_of()`

tanscorpio7 (changes requested):

src/backend/utils/adt/acl.c
anju15bharti: If direct membership doesn't exist then it will follow the previous path which again will iterate through all the members even if it will get the required role which we want to avoid.
Also by adding the hook in role_is_member ensures both the cases:
1. If member to be checked is bbf_role_admin then it is doing direct check for admin membership using syscache and then returning the required role.
2. if direct admin membership does not exists then it is iterating through all the membership and once got that required role returning from there itself without further iteration which we do in role_is_member.

tanscorpio7: for babelfish roles direct member ship will always exists.

tanscorpio7: Also the current changes breaks how `roles_is_member_of` is defined to work.
`roles_is_member_of` should always returns all the members of the given role with given member ship type recursively.

anju15bharti: `roles_is_member_of ` will continue to work the same except for admin option case when member to be checked is bbf_role_admin.  I we add hook in `is_admin_of_role ` we need to additionally handle checking membership through syscache which we are anyway doing it in `roles_is_member_of ` in our case without any additional handling.

In case of admin option although we were storing  all the members of the given role but we were actually not using it in that case, we were using the inplace variable value not the list.

tanscorpio7: `roles_is_member_of` is defined to return both, all the members of the given role & optionally the admin.
This can easily break if some one starts using this function for both fetching the list of roles and admin of a given role.

tanscorpio7: Plus `roles_is_member_of` is a very hot code path called by many other functions, how do we logically explain this change will not break the other functions in future ?

anju15bharti: Added hook in `is_admin_of_role ` to check direct membership there itself.

src/backend/utils/adt/acl.c
```diff
@@ -5278,6 +5279,19 @@ is_admin_of_role(Oid member, Oid role)
 	if (member == role)
 		return false;
 
+	/*
+	 * Check if the given member is bbf_role_admin.
+         * If the member has the privilege to grant a given role through direct membership,
+         * returns the member with admin option set to true.
+	 */
+	if (sql_dialect == SQL_DIALECT_TSQL 
+		&& get_bbf_admin_oid_hook && member == (*get_bbf_admin_oid_hook)()
```
tanscorpio7: we can put comments and all the other conditions inside the hook function. This helps during merge from community.

src/backend/utils/adt/acl.c
```diff
@@ -5278,6 +5279,19 @@ is_admin_of_role(Oid member, Oid role)
...
+			&& (has_bbf_role_direct_membership_with_admin_true_hook) 
+				&& (*has_bbf_role_direct_membership_with_admin_true_hook)(role))
+	{
+		return member;
```
tanscorpio7: return true ?

tanscorpio7 (changes requested):

src/backend/utils/adt/acl.c
anju15bharti: what do you mean by true?

tanscorpio7: return type for function is bool

tanscorpio7 (approved):

HarshLunagariya (approved):

KushaalShroff (approved):

surendravishnoi (approved):

PR #517, opened 2025-01-13
[BABEL] Fix object resolution inside pltsql routines

tanscorpio7:
### Description

Remove search path handling for pltsql functions in fmgr.c.
This will now be handled in extension code.

Also we were implicitly appending sys schema to search path
when dialect was TSQL but now we ensure search path will 
always contain sys for TDS connections.
So to help performance, make implicit appending of sys to
search path logic work only for non TDS connections to 
babelfish_db.

### Issues Resolved

[BABEL-5448]

### Sign Off

### Check List

### DIFF
--- a/src/backend/catalog/namespace.c
+++ b/src/backend/catalog/namespace.c
@@ -3969,7 +3969,7 @@ GetSearchPathMatcher(MemoryContext context)
 			result->addTemp = true;
 		else
 		{
-			if (!is_bbf_tds_connection_hook || is_bbf_tds_connection_hook(false))
+			if (!is_bbf_tds_connection_hook || is_bbf_tds_connection_hook() || sql_dialect != SQL_DIALECT_TSQL)
 				Assert(linitial_oid(schemas) == PG_CATALOG_NAMESPACE);
 			result->addCatalog = true;
 		}
@@ -4036,7 +4036,7 @@ SearchPathMatchesCurrentEnvironment(SearchPathMatcher *path)
 	/* If path->addCatalog, next item should be pg_catalog. */
 	if (path->addCatalog)
 	{
-		if (is_bbf_tds_connection_hook && !is_bbf_tds_connection_hook(false))
+		if (is_bbf_tds_connection_hook && !is_bbf_tds_connection_hook() && sql_dialect == SQL_DIALECT_TSQL)
 		{
 			if (lc && lfirst_oid(lc) == get_namespace_oid(SYS_NAMESPACE_NAME, true))
 				lc = lnext(activeSearchPath, lc);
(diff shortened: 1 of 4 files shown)


src/backend/catalog/namespace.c
```diff
@@ -4872,6 +4866,14 @@ assign_search_path(const char *newval, void *extra)
...
 assign_sql_dialect(int newval, void *extra)
 {
+	/*
+	 * We only append sys to search path implictly for non tds connection
+	 * to babelfish database so no need to invalidate search path or its
+	 * cache for tds connections
+	 */
+	if (is_bbf_tds_connection_hook && is_bbf_tds_connection_hook())
+		return;
+
```
tanscorpio7, line 4877: We implicitly append sys to search path when dialect was tsql and sys was not already in search path. This is bad since everytime sql_dialect changes we have to recompute the search path. From PG17 onwards its more problematic since we also invalidate the search path hast table on every dialect change.

But now we are ensuring search path is always correct for TDS connections in extensions code, so we do not need this for TDS connections.

surendravishnoi: why don't we complete remove it? is there a use case to set it for PG connections with TSQL dialect?

tanscorpio7: For PG connections when users execute pltsql procedures, they will implicitly get sys in their search path.
To maintain this behaviour, I have still kept this change for non tds connections.

surendravishnoi: if it is pltsql procedure, wouldn't your new code take care of setting search path correctly?

tanscorpio7: The new changes are only triggered for ptsql routines and TDS connections. We are not handling the postgres endpoint behaviour for pltsql routines in this PR. There are couple of reasons for not handling it in this PR:
1. We need to define how search path would work in postgres endpoint for babelfish_db, which needs more thought.
2. Switching db context is not possible for postgres endpoint because when you are resetting the db_context, the saved db name will be "" (empty), so we need to think of a different way to reset db_context for postgres endpoint.

surendravishnoi: this needs discussion

surendravishnoi (dismissed):

src/backend/catalog/namespace.c
```diff
@@ -3968,8 +3969,8 @@ GetSearchPathMatcher(MemoryContext context)
 			result->addTemp = true;
 		else
 		{
-			/* sys comes before pg_catalog when sql_dialect is tsql */
-			if (sql_dialect != SQL_DIALECT_TSQL)
+			if (!is_bbf_tds_connection_hook || is_bbf_tds_connection_hook() || sql_dialect != SQL_DIALECT_TSQL)
```
surendravishnoi: is this wrong copy paste?

tanscorpio7: Yes, removed the redundant if condition

surendravishnoi (approved):

PR #522, opened 2025-01-30
Fix trigger failure with FK reference on table variables/temp tables

roshan0708:
### Description

Resolve issue where triggers referencing foreign keys fail with "relation does not exist" error when dealing with table variables or temporary tables. This fix ensures proper handling of search paths and query revalidation in the context of triggers.

Additionally, optimize dialect setting by directly using assign_sql_dialect function instead of set_config_option in the `RUN_AS_PSQL` macro, reducing overhead in dialect switching operations.

Extension PR: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3440
 
### Issues Resolved

Task: BABEL-5270
 
### Check List


src/backend/utils/adt/ri_triggers.c
```diff
@@ -1208,11 +1213,11 @@ ri_set(TriggerData *trigdata, bool is_set_null, int tgkind)
 	/*
 	 * We have a plan now. Run it to update the existing references.
 	 */
-	ri_PerformCheck(riinfo, &qkey, qplan,
+	RUN_AS_PSQL(ri_PerformCheck(riinfo, &qkey, qplan,
```
tanscorpio7, line 1210: Do we also need to do this at line 1605. `spi_result = SPI_execute_snapshot(qplan,` ?

tanscorpio7 (changes requested):

src/backend/utils/adt/ri_triggers.c
```diff
@@ -430,11 +435,11 @@ RI_FKey_check(TriggerData *trigdata)
 	 * referenced side.  The reason is that our snapshot must be fresh in
 	 * order for the hack in find_inheritance_children() to work.
 	 */
-	ri_PerformCheck(riinfo, &qkey, qplan,
+	RUN_AS_PSQL(ri_PerformCheck(riinfo, &qkey, qplan,
```
surendravishnoi, line 432: instead of changing all the calls, can we change inside the function at lowest level where it is required?

.github/workflows/main.yml
```diff
@@ -15,6 +15,7 @@ jobs:
         run: |
           sudo apt update --fix-missing -y
           sudo apt install -y libipc-run-perl
+          sudo apt-get install -y libreadline-dev
```
surendravishnoi: why this change?

src/backend/utils/adt/ri_triggers.c
```diff
@@ -95,29 +96,33 @@
...
+	bool reset_dialect = (sql_dialect == SQL_DIALECT_TSQL);                                         		\
+	int sql_dialect_old = sql_dialect;																		\
+	void *newextra = NULL;																					\
 	if (reset_dialect)																						\
-		set_config_option("babelfishpg_tsql.sql_dialect", "postgres",										\
-			GUC_CONTEXT_CONFIG,						\
-			PGC_S_SESSION, GUC_ACTION_SAVE, true, 0, false);												\
-																											\
+	{																										\
+		sql_dialect = SQL_DIALECT_PG;																		\
```
surendravishnoi, line 103: Please confirm with Tanzeel that it is safe to do so. Also, do we have perf data that it is an issue?

tanscorpio7: I actually prefer it this way because this is how we change sql_dialect in fmgr.c.
set_config() would have been okay here if we are creating a new nest level but that will only add to the performance hit.

.github/workflows/main.yml
roshan0708: Since we've now configured engine with `--without-readline`, @tanscorpio7 do we still need this?

src/backend/utils/adt/ri_triggers.c
```diff
@@ -749,11 +754,11 @@ ri_restrict(TriggerData *trigdata, bool is_no_action)
 	/*
 	 * We have a plan now. Run it to check for existing references.
 	 */
-	ri_PerformCheck(riinfo, &qkey, qplan,
+	RUN_AS_PSQL(ri_PerformCheck(riinfo, &qkey, qplan,
```
Deepesh125, line 751: which test case triggers this?

src/backend/utils/adt/ri_triggers.c
```diff
@@ -976,11 +981,11 @@ RI_FKey_cascade_upd(PG_FUNCTION_ARGS)
 	/*
 	 * We have a plan now. Run it to update the existing references.
 	 */
-	ri_PerformCheck(riinfo, &qkey, qplan,
+	RUN_AS_PSQL(ri_PerformCheck(riinfo, &qkey, qplan,
```
Deepesh125, line 978: can you please point me the test case for this?

src/backend/utils/adt/ri_triggers.c
Deepesh125: plz point out test case for this

src/backend/utils/adt/ri_triggers.c
```diff
@@ -855,11 +860,11 @@ RI_FKey_cascade_del(PG_FUNCTION_ARGS)
 	 * We have a plan now. Build up the arguments from the key values in the
 	 * deleted PK tuple and delete the referencing rows
 	 */
-	ri_PerformCheck(riinfo, &qkey, qplan,
+	RUN_AS_PSQL(ri_PerformCheck(riinfo, &qkey, qplan,
```
Deepesh125, line 857: Test case?

src/backend/utils/adt/ri_triggers.c
Deepesh125: and yes, we need to handle this for alter table add fk case.

Deepesh125 (changes requested):

.github/workflows/main.yml
tanscorpio7: No, we do not need to install readline now.

src/backend/utils/adt/ri_triggers.c
tanscorpio7: Since we also support partitions now, we may also need this change in `RI_PartitionRemove_Check`.
Let us review the entire file for any more places where this change could be needed now.

roshan0708: @tanscorpio7 this function won't do anything right now since we haven't supported `SWITCH PARTITION` yet. Added it anyway in the latest commit so it's ready to go when we get around to supporting it in `ALTER TABLE`. 👍

src/backend/utils/adt/ri_triggers.c
```diff
@@ -95,29 +96,33 @@
...
 		if (reset_dialect)																					\
-			set_config_option("babelfishpg_tsql.sql_dialect", "tsql",										\
-							GUC_CONTEXT_CONFIG,		\
-							PGC_S_SESSION, GUC_ACTION_SAVE, true, 0, false);								\
+		{																									\
+			sql_dialect = sql_dialect_old;																	\
+			assign_sql_dialect(sql_dialect, newextra);														\
+		}																									\
 		PG_RE_THROW();																						\
 	}																										\
```
tanscorpio7: can be converted to PG_FINALLY();

.github/workflows/main.yml
```diff
@@ -15,6 +15,7 @@ jobs:
         run: |
           sudo apt update --fix-missing -y
           sudo apt install -y libipc-run-perl
+          sudo apt-get install -y
```
tanscorpio7: We can uncommit this line.

tanscorpio7 (approved):

.github/workflows/main.yml
roshan0708: This is to resolve this - https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/pull/522#discussion_r1937370671

tanscorpio7 (approved):

src/backend/utils/adt/ri_triggers.c
```diff
@@ -95,29 +96,27 @@
...
 		qplan;																								\
 	}																										\
-	PG_CATCH();																								\
+	PG_FINALLY();																							\
 	{																										\
 		if (reset_dialect)																					\
-			set_config_option("babelfishpg_tsql.sql_dialect", "tsql",										\
-							GUC_CONTEXT_CONFIG,		\
-							PGC_S_SESSION, GUC_ACTION_SAVE, true, 0, false);								\
-		PG_RE_THROW();																						\
```
surendravishnoi, line 114: why are we not re-throwing the error?

src/backend/utils/adt/ri_triggers.c
```diff
@@ -419,8 +418,8 @@ RI_FKey_check(TriggerData *trigdata)
 		appendStringInfoString(&querybuf, " FOR KEY SHARE OF x");
 
 		/* Prepare and save the plan */
-		RUN_AS_PSQL(qplan = ri_PlanCheck(querybuf.data, riinfo->nkeys, queryoids,
```
surendravishnoi: though this is not ideal, let us not change existing code and apply checks at lower level, that might create unintended regressions in existing code

src/backend/utils/adt/ri_triggers.c
```diff
@@ -1597,11 +1596,11 @@ RI_Initial_Check(Trigger *trigger, Relation fk_rel, Relation pk_rel)
 	 * register the snapshot, because SPI_execute_snapshot will see to it. We
 	 * need at most one tuple returned, so pass limit = 1.
 	 */
-	spi_result = SPI_execute_snapshot(qplan,
+	RUN_AS_PSQL(spi_result = SPI_execute_snapshot(qplan,
```
surendravishnoi, line 1599: I think this is no better than original one. Let us keep original one only. I thought we want do it at one place in caller but that does not seems like the case unless we do something like this
func()
{
    SET_PSQL_DIALECT();
    function body
    RESET_PSQL_DIALECT();
}

#define SET_PSQL_DIALECT 
 set dialect;
PG_TRY();
 {
        

#define RESET_PSQL_DIALECT
}
PG_CATCH()
{
     reset dialect;
}

which is again is not ideal.

src/backend/utils/adt/ri_triggers.c
```diff
@@ -550,8 +549,8 @@ ri_Check_Pk_Match(Relation pk_rel, Relation fk_rel,
 		appendStringInfoString(&querybuf, " FOR KEY SHARE OF x");
 
 		/* Prepare and save the plan */
-		RUN_AS_PSQL(qplan = ri_PlanCheck(querybuf.data, riinfo->nkeys, queryoids,
```
surendravishnoi: safest would be this

if (dialect == TSQL)
   RUN_AS_PSQL()
else
   current code

roshan0708: We already have this check inside `RUN_AS_PSQL` macro. So we don't need to explicitly check this.

src/backend/utils/adt/ri_triggers.c
roshan0708: When using `PG_FINALLY()`, explicit calls to `PG_RE_THROW()` are unnecessary. The `PG_END_TRY()` macro automatically handles error re-throwing if an exception occurred during execution.

src/backend/utils/adt/ri_triggers.c
```diff
@@ -95,31 +96,45 @@
...
-		set_config_option("babelfishpg_tsql.sql_dialect", "tsql",											\
-						GUC_CONTEXT_CONFIG,			\
-						PGC_S_SESSION, GUC_ACTION_SAVE, true, 0, false);									\
 } while (0)
 
+/*
+ * macros for executing the given code block directly if not in TSQL dialect,
+ * or wrapped with RUN_AS_PSQL if in TSQL dialect.
+ */
+#define DIALECT_WRAPPER(code)																				\
```
surendravishnoi: why we need separate macro, let us modify RUN_AS_PSQL to do the same

surendravishnoi (approved):

Deepesh125 (approved):

PR #529, opened 2025-02-05
Added BBF hook to adjust numeric result to correct scale

rohit01010:
### Description
For Babelfish, the expected behaviour should be for every arithmetic computation which results in numeric datatype, the result should be adjusted(truncated/rounded) to the correct scale. To get this behaviour added hooks at following places to adjust the result value.
- evaluate_expr() - All the arithmetic computations that result in constant value, happens here. So added a hook to adjust the result after every such arithmetic computation.
- ExecInterpExpr() - All the arithmetic computations other than mentioned above, happens here. So added a hook to adjust the result after every such arithmetic computation.
- finalize_aggregate() - For aggregate functions, the final result is computed here, so added a hook to adjust the result for aggregate functions.

Also, 
- for some nodes such as `T_OpExpr, T_DistinctExpr, T_CoerceViaIO` the result_typmod was directly hardcoded as `-1` in PostgreSQL, Added hook `pltsql_exprTypmod_hook` to compute the typmod of the expression and passed those instead of `-1` to avoid losing typmod information.
- added `pltsql_exprTypmod_hook` to compute expression typmod for expressions whose typmod is not computed by `exprTypmod`.
- During initialisation of resulttype in function `ExecInitResultTypeTL` added hook `ExecUpdateResultTypeTL_hook` to set correct typmod of all attributes of result tuple.

Extension PR: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3455 

### Issues Resolved
BABEL-5467
 
### Check List


src/backend/optimizer/util/clauses.c
```diff
@@ -2742,12 +2747,15 @@ eval_const_expressions_mutator(Node *node,
 					set_opfuncid((OpExpr *) expr);	/* rely on struct
 													 * equivalence */
 
+					if (sql_dialect == SQL_DIALECT_TSQL && resolve_numeric_typmod_from_exp_hook && expr->opresulttype == NUMERICOID)
```
Yvinayak07: Can we use get  getBaseType(expr->opresulttype) == NUMERICOID. 
This can cover decimal and UDT's.

src/backend/nodes/nodeFuncs.c
```diff
@@ -539,6 +540,11 @@ exprTypmod(const Node *expr)
 		default:
 			break;
 	}
+
+	if (sql_dialect == SQL_DIALECT_TSQL && resolve_numeric_typmod_from_exp_hook){
```
Yvinayak07: NIT: `{` should be in next line.

rohit01010: Updated

src/backend/optimizer/util/clauses.c
rohit01010: Updated

src/backend/nodes/nodeFuncs.c
```diff
@@ -539,6 +540,10 @@ exprTypmod(const Node *expr)
 		default:
 			break;
 	}
+
+	if (sql_dialect == SQL_DIALECT_TSQL && resolve_numeric_typmod_from_exp_hook && getBaseType(exprType(expr)) == NUMERICOID)
```
shah-nirmit: Suggestion: Should we make naming more generic also the if conditions? This only seems to target numeric
In future if we need support to resolve typmod for more expressions , we could reuse this hook

src/backend/executor/execExprInterp.c
```diff
@@ -762,6 +766,10 @@ ExecInterpExpr(ExprState *state, ExprContext *econtext, bool *isnull)
 			}
 			fcinfo->isnull = false;
 			d = op->d.func.fn_addr(fcinfo);
+
+			if (sql_dialect == SQL_DIALECT_TSQL && bbf_trunc_numeric_result_hook)
```
shah-nirmit: Question: For every FuncExpr we will enter this hook, extra instructions for every expression evaluation, seems to me it will affect performance

src/backend/nodes/nodeFuncs.c
rohit01010: Updated hook to be more generic so that we can extend it later for other type of expressions

src/include/executor/executor.h
```diff
@@ -102,6 +102,12 @@ extern PGDLLEXPORT TriggerRecuresiveCheck_hook_type TriggerRecuresiveCheck_hook;
 typedef bool (*bbfViewHasInsteadofTrigger_hook_type) (Relation view, CmdType event);
 extern PGDLLIMPORT bbfViewHasInsteadofTrigger_hook_type bbfViewHasInsteadofTrigger_hook;
 
+typedef Datum (*pltsql_trunc_numeric_result_hook_type) (Plan *plan, Node *fn_expr, Datum result, Oid result_type, int32 result_typmod);
```
Deepesh125: hook name should be generic, pltsql does not need to be added

src/backend/executor/nodeAgg.c
```diff
@@ -1117,6 +1118,8 @@ finalize_aggregate(AggState *aggstate,
 			Datum		result;
 
 			result = FunctionCallInvoke(fcinfo);
+            if (sql_dialect == SQL_DIALECT_TSQL && pltsql_trunc_numeric_result_hook && getBaseType(peragg->aggref->aggtype) == NUMERICOID)
+                result = pltsql_trunc_numeric_result_hook(aggstate->ss.ps.plan, (Node *) peragg->aggref, result, peragg->aggref->aggtype, -1);
 			*resultIsNull = fcinfo->isnull;
```
Deepesh125, line 1120: we should call this hook after this NULL check.

Deepesh125 (changes requested):

src/include/nodes/nodeFuncs.h
```diff
@@ -14,6 +14,7 @@
 #define NODEFUNCS_H
 
 #include "nodes/parsenodes.h"
+#include "nodes/plannodes.h"
```
Deepesh125, line 17: why?

src/backend/optimizer/util/clauses.c
```diff
@@ -5128,6 +5140,9 @@ evaluate_expr(Expr *expr, Oid result_type, int32 result_typmod,
 	/* Release all the junk we just created */
 	FreeExecutorState(estate);
 
+	if (sql_dialect == SQL_DIALECT_TSQL && pltsql_trunc_numeric_result_hook && getBaseType(result_type) == NUMERICOID)
```
Deepesh125: should only be doing this if const is not null?

src/backend/optimizer/util/clauses.c
```diff
@@ -2631,19 +2631,23 @@ eval_const_expressions_mutator(Node *node,
...
 				OpExpr	   *newexpr;
+				int32       result_typmod = -1;
 
 				/*
 				 * Need to get OID of underlying function.  Okay to scribble
 				 * on input to this extent.
 				 */
 				set_opfuncid(expr);
 
+				if (sql_dialect == SQL_DIALECT_TSQL && pltsql_exprTypmod_hook)
```
Deepesh125: please explain why are we handling only handful of expr type? what about other node types?

Deepesh125 (changes requested):

src/backend/optimizer/util/clauses.c
rohit01010: we need to set the correct typmod only for those nodes whose typmod is set as `-1`. For other nodes, the typmod is already set correctly.

src/include/nodes/nodeFuncs.h
rohit01010: First argument of `pltsql_exprTypmod_hook_type` is of type Plan, and Plan struct is defined in plannodes.h

src/include/executor/executor.h
rohit01010: Updated

src/backend/executor/nodeAgg.c
rohit01010: Added NULL check before calling the hook

src/backend/optimizer/util/clauses.c
rohit01010: Added null check

shah-nirmit (approved):

src/backend/executor/execExprInterp.c
```diff
@@ -762,9 +767,17 @@ ExecInterpExpr(ExprState *state, ExprContext *econtext, bool *isnull)
 			}
 			fcinfo->isnull = false;
 			d = op->d.func.fn_addr(fcinfo);
+
 			*op->resvalue = d;
 			*op->resnull = fcinfo->isnull;
 
+			if (sql_dialect == SQL_DIALECT_TSQL && trunc_numeric_result_hook && !(*op->resnull))
```
Deepesh125: what is difference b/w `EEOP_FUNCEXPR` and `EEOP_FUNCEXPR_STRICT`, do we need similar truncation for other case?

src/backend/executor/execTuples.c
```diff
@@ -69,6 +69,7 @@
 #include "utils/expandeddatum.h"
 #include "utils/lsyscache.h"
 #include "utils/typcache.h"
+#include "parser/parser.h"
```
Deepesh125: parser files included in executor?

src/backend/executor/execTuples.c
```diff
@@ -86,6 +87,8 @@ const TupleTableSlotOps TTSOpsHeapTuple;
 const TupleTableSlotOps TTSOpsMinimalTuple;
 const TupleTableSlotOps TTSOpsBufferHeapTuple;
 
+/* pltsql hook to initialize result type */
+pltsql_ExecUpdateResultTypeTL_hook_type pltsql_ExecUpdateResultTypeTL_hook = NULL;
```
Deepesh125: hook name should be generic

src/backend/executor/execExprInterp.c
```diff
@@ -762,9 +767,17 @@ ExecInterpExpr(ExprState *state, ExprContext *econtext, bool *isnull)
 			}
 			fcinfo->isnull = false;
 			d = op->d.func.fn_addr(fcinfo);
+
```
Deepesh125: plz remove this

Deepesh125 (changes requested):

src/backend/executor/execExprInterp.c
```diff
@@ -762,9 +767,17 @@ ExecInterpExpr(ExprState *state, ExprContext *econtext, bool *isnull)
...
 			*op->resvalue = d;
 			*op->resnull = fcinfo->isnull;
 
+			if (sql_dialect == SQL_DIALECT_TSQL && trunc_numeric_result_hook && !(*op->resnull))
+			{
+				func_result_type = get_func_rettype(fcinfo->flinfo->fn_oid);
```
surendravishnoi: just add a hook here and put rest of logic including dialect check inside the hook code, this will reduce code here and no need to include new headers. Applies to other places as well

src/backend/executor/execTuples.c
rohit01010: removed

src/backend/executor/execTuples.c
rohit01010: updated

src/backend/executor/execExprInterp.c
rohit01010: removed

src/backend/executor/execExprInterp.c
rohit01010: Moved checks around hook inside hooks, hence also removed added header files as they are not required now.

src/backend/executor/nodeAgg.c
```diff
@@ -1118,6 +1118,15 @@ finalize_aggregate(AggState *aggstate,
 
 			result = FunctionCallInvoke(fcinfo);
 			*resultIsNull = fcinfo->isnull;
+			if (trunc_numeric_result_hook)
+			{
+			    result = trunc_numeric_result_hook(aggstate->ss.ps.plan,
```
surendravishnoi: Please make sure that either we have validated that we can traverse pointers here or put check like if(a) if (a->b) if (a->b->c) inside hook

surendravishnoi (dismissed):

src/backend/executor/nodeAgg.c
rohit01010: Added sanity checks on such traverse pointers

src/backend/executor/execExprInterp.c
rohit01010: Functions which are not marked as strict will be executed under `EEOP_FUNCEXPR` and functions that are marked strict will be executed under `EEOP_FUNCEXPR_STRICT`. We need similar truncation for other case as well. Added.

surendravishnoi (dismissed):

Deepesh125 (dismissed):

KushaalShroff (dismissed):

surendravishnoi (approved):

tanscorpio7:
I am curious why we cannot do this in post planner hook. That is walk the tree and update the typmod.

shah-nirmit:
> I am curious why we cannot do this in post planner hook. That is walk the tree and update the typmod.

I am guessing this may be due to `simplify_function` which happens during planning...

PR #539, opened 2025-02-24
Fix Improper Plan Invalidation for Fkey Integrity check queries

shah-nirmit:
### Description
Introduce a hook , that checks whether the hashed plan for Referential integrity queries is valid or not , as in case for babelfish there can be change in search path due to for example table variables etc , if the search path doesn't match we discard the plan . And replan this query.
 
### Issues Resolved
BABEL-5270

Extension PR https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3528
Signed-off-by: Nirmit Shah <nirmisha@amazon.com>

### Check List

### DIFF
--- a/src/backend/commands/trigger.c
+++ b/src/backend/commands/trigger.c
@@ -5293,7 +5293,7 @@ MakeTransitionCaptureState(TriggerDesc *trigdesc, Oid relid, CmdType cmdType)
 	 * delete statment as well, so we'll init both old and new transition tables
 	 * for Tsql dialect
 	 * */
-	if (check_pltsql_support_tsql_transactions_hook && (*check_pltsql_support_tsql_transactions_hook)() && (need_old_del || need_new_ins))
+	if (sql_dialect == SQL_DIALECT_TSQL && (need_old_del || need_new_ins))
 	{
 		need_new_ins = true;
 		need_old_del = true;
--- a/src/backend/commands/trigger.c
+++ b/src/backend/commands/trigger.c
@@ -5293,7 +5293,7 @@ MakeTransitionCaptureState(TriggerDesc *trigdesc, Oid relid, CmdType cmdType)
 	 * delete statment as well, so we'll init both old and new transition tables
 	 * for Tsql dialect
 	 * */
-	if (sql_dialect == SQL_DIALECT_TSQL && (need_old_del || need_new_ins))
+	if (check_pltsql_support_tsql_transactions_hook && (*check_pltsql_support_tsql_transactions_hook)() && (need_old_del || need_new_ins))
 	{
 		need_new_ins = true;
 		need_old_del = true;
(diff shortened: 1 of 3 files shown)


tanscorpio7:
Changes make sense to me but can we ensure there is no other place which could have this issue.
Very vague suggestion but can we think of / look at the code path where the internal foreign key queries could reach and we have added dialect checks.

tanscorpio7:
For example I can see lot of dialect checks in instead of triggers in nodeModifyTable.c. Can we test those with foreign keys.

src/backend/commands/trigger.c
```diff
@@ -5293,7 +5293,7 @@ MakeTransitionCaptureState(TriggerDesc *trigdesc, Oid relid, CmdType cmdType)
 	 * delete statment as well, so we'll init both old and new transition tables
 	 * for Tsql dialect
 	 * */
-	if (sql_dialect == SQL_DIALECT_TSQL && (need_old_del || need_new_ins))
+	if (check_pltsql_support_tsql_transactions_hook && (*check_pltsql_support_tsql_transactions_hook)() && (need_old_del || need_new_ins))
```
surendravishnoi: why are we using transaction check? there is different check to find wether it is TSQL trigger or not

shah-nirmit: Are you referring to this? https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/181b49df4a594be8aac13e968e4e6aaa4bc7c8b2/src/backend/commands/trigger.c#L7388 
I was thinking there can be a chain of triggers , and if we encounter any PG related trigger we could limit that using this hook?

tanscorpio7: I think we have the trigger OID available here. What if we change the check to
```
if (TSQL Trigger)
    then do ...
```

shah-nirmit: We have list of triggers available here in TriggerDesc.

tanscorpio7: At least 1 trigger is a t-sql trigger in the list should be a good check. Also that is the original intention behind this code. T-sql triggers should always have access to all transition tables.

tanscorpio7: @surendravishnoi Would that be a valid check ?

tanscorpio7: I will try to summarize the offline discussion with Surendra
Setting psql dialect here is not a great option because then t-sql triggers could be fired from a postgres query. In the past there were some inter interoperability issues with t-sql triggers, which is why they were blocked.

Other approaches we can investigate
1. In ri_triggers can we check if current active search path changed and is not equal to the cached _SPI_plan's search path. (invalidate plan before even reaching `ri_PerformCheck`.
2. Can we add a dialect field in plan structure which will tell us what was the dialect when this plan was generated. We will replan using this dialect always.
3. Can we convert the foreign key queries into a t-sql query when the dialect is TSQL ?

tanscorpio7 (approved):

shardgupta (approved):

PR #560, opened 2025-04-01
Avoid checking permission of Babelfish temp tables on parallel worker

Deepesh125:
### Description

Consider following facts,

1. Babelfish temp tables are implemented using ENR which is not shared between different backends. So if Parallel worker
tries to check permissions on Babelfish then it will fail.

2. Any user should be able to access Babelfish temp tables under given session. 

3. Postgres by default does not allow parallel operations on temp tables. Attempt to do so will result in run time error.

Due to above facts, we should avoid permission check on temp tables within parallel workers while ensuring that leader
does required permission check on other tables. This commits achieves this behaviour by introducing following three
hooks, 

1. ParallelQueryMain_hook -- Hook that allows other extensions to pass on additional details from Leader node to
parallel worker. For example, Babelfish extension can pass details of Babelfish temp table defined under current
session with Parallel workers.

2. ExecInitParallelPlan_hook -- Hook that allows Parallel worker to gather additional details passed by Leader node.
For example, Babelfish extension can collect the details of Babelfish temp table shared by Leader node so that it
can avoid permission checks.

3. ExecCheckOneRelPerms_hook -- Hook that allows extension control permission checking on given relation/table.
For example, Babelfish can use it to avoid permission check on temp tables under parallel worker.
 
### Issues Resolved

BABEL-5703
 
### Check List

### DIFF
--- a/src/backend/executor/execMain.c
+++ b/src/backend/executor/execMain.c
@@ -62,8 +62,6 @@
 #include "utils/snapmgr.h"
 
 
-Bitmapset *tempRelids = NULL;
-
 /* Hooks for plugins to get control in ExecutorStart/Run/Finish/End */
 ExecutorStart_hook_type ExecutorStart_hook = NULL;
 ExecutorRun_hook_type ExecutorRun_hook = NULL;
--- a/src/backend/executor/execMain.c
+++ b/src/backend/executor/execMain.c
@@ -656,6 +656,7 @@ ExecCheckOneRelPerms(RTEPermissionInfo *perminfo)
 	 * operations are not allowed for temp table in Postgres. Babelfish
 	 * can skip permission check for such use cases under parallel worker
 	 * using this hook.
+	 * Note - This hook must not be used outside of Babelfish parallel worker
 	 */
 	if (IsBabelfishParallelWorker() &&
 		ExecCheckOneRelPerms_hook &&
--- a/src/backend/executor/execMain.c
+++ b/src/backend/executor/execMain.c
@@ -657,7 +657,8 @@ ExecCheckOneRelPerms(RTEPermissionInfo *perminfo)
 	 * are not allowed for temp table in Postgres. So give extension a
 	 * chance to skip permission check for such use cases.
 	 */
-	if (ExecCheckOneRelPerms_hook &&
+	if (IsBabelfishParallelWorker() &&
+		ExecCheckOneRelPerms_hook &&
 		(*ExecCheckOneRelPerms_hook)(perminfo))
 		return true;
 
--- a/src/backend/executor/execMain.c
+++ b/src/backend/executor/execMain.c
@@ -658,8 +658,8 @@ ExecCheckOneRelPerms(RTEPermissionInfo *perminfo)
 	 * using this hook.
 	 * Note - This hook must not be used outside of Babelfish parallel worker
 	 */
-	if (IsBabelfishParallelWorker() &&
-		ExecCheckOneRelPerms_hook &&
+	if (ExecCheckOneRelPerms_hook &&
+		IsBabelfishParallelWorker() &&
 		(*ExecCheckOneRelPerms_hook)(perminfo))
 		return true;
 
--- a/src/backend/executor/execMain.c
+++ b/src/backend/executor/execMain.c
@@ -72,6 +72,8 @@ ExecutorEnd_hook_type ExecutorEnd_hook = NULL;
 
 TriggerRecuresiveCheck_hook_type TriggerRecuresiveCheck_hook = NULL;
 check_rowcount_hook_type check_rowcount_hook = NULL;
+skip_ExecutorCheckPerms_hook_type skip_ExecutorCheckPerms_hook = NULL;
+
 /* Hook for plugin to get control in ExecCheckPermissions() */
 ExecutorCheckPerms_hook_type ExecutorCheckPerms_hook = NULL;
 
@@ -618,7 +620,13 @@ ExecCheckPermissions(List *rangeTable, List *rteperminfos,
 
 		Assert(OidIsValid(perminfo->relid));
 
-		if (IsBabelfishParallelWorker() && bms_is_member(perminfo->relid, tempRelids))
+		/* 
+		 * Babelfish specific logic - skip permission check for temp table
+		 * under parallel worker
+		 */
+		if (IsBabelfishParallelWorker() &&
+			skip_ExecutorCheckPerms_hook &&
+			(*skip_ExecutorCheckPerms_hook)(perminfo->relid))
 			continue;
 
 		result = ExecCheckOneRelPerms(perminfo);
--- a/src/backend/executor/execMain.c
+++ b/src/backend/executor/execMain.c
@@ -70,7 +70,7 @@ ExecutorEnd_hook_type ExecutorEnd_hook = NULL;
 
 TriggerRecuresiveCheck_hook_type TriggerRecuresiveCheck_hook = NULL;
 check_rowcount_hook_type check_rowcount_hook = NULL;
-skip_ExecutorCheckPerms_hook_type skip_ExecutorCheckPerms_hook = NULL;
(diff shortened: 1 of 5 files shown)


src/backend/executor/execMain.c
```diff
@@ -70,6 +70,8 @@ ExecutorEnd_hook_type ExecutorEnd_hook = NULL;
 
 TriggerRecuresiveCheck_hook_type TriggerRecuresiveCheck_hook = NULL;
 check_rowcount_hook_type check_rowcount_hook = NULL;
+skip_ExecutorCheckPerms_hook_type skip_ExecutorCheckPerms_hook = NULL;
+
```
kuntalghosh: Not a good naming choice for a hook. It should be something like,
```
skip_permission_check = false;

pre_ExecutorCheckPerms_hook(..., &skip_permission_check);

if (skip_permission_check == true)
return;
```

src/backend/executor/execMain.c
```diff
@@ -615,6 +616,16 @@ ExecCheckPermissions(List *rangeTable, List *rteperminfos,
 		RTEPermissionInfo *perminfo = lfirst_node(RTEPermissionInfo, l);
 
 		Assert(OidIsValid(perminfo->relid));
+
+		/* 
+		 * Babelfish specific logic - skip permission check for temp table
+		 * under parallel worker
+		 */
```
kuntalghosh: The comment should give the full context.

src/backend/executor/execMain.c
```diff
@@ -636,7 +647,7 @@ ExecCheckPermissions(List *rangeTable, List *rteperminfos,
  * ExecCheckOneRelPerms
  *		Check access permissions for a single relation.
  */
-static bool
+bool
```
kuntalghosh: Define an extern wrapper instaed.

src/backend/executor/execParallel.c
```diff
@@ -1433,6 +1448,11 @@ ParallelQueryMain(dsm_segment *seg, shm_toc *toc)
 	area_space = shm_toc_lookup(toc, PARALLEL_KEY_DSA, false);
 	area = dsa_attach_in_place(area_space, seg);
 
+	if (IsBabelfishParallelWorker() && ParallelQueryMain_hook)
+	{
+		(*ParallelQueryMain_hook)(toc);
+	}
+
```
kuntalghosh: Why is this needed? Please add a comment.

src/include/access/parallel.h
```diff
@@ -81,15 +81,11 @@ extern void ParallelWorkerMain(Datum main_arg);
 /* Below helpers are added to support parallel workers in Babelfish context */
 extern bool IsBabelfishParallelWorker(void);
 
-/* Key for BabelfishFixedParallelState */
-#define BABELFISH_PARALLEL_KEY_FIXED        UINT64CONST(0xBBF0000000000001)
-
```
kuntalghosh, line 86: This should be in engine to avoid future conflicts.

kuntalghosh (changes requested):

tanscorpio7:
Just curious how parallel workers fetches other details for temp tables relations after permission check ?

It must need more information about the relation while executing the node with temp table.

src/backend/executor/execMain.c
2jungkook: +1. WHY COMMENTS section in https://antirez.com/news/124 can be helpful.

src/include/executor/execParallel.h
```diff
@@ -48,4 +48,10 @@ extern void ExecParallelReinitialize(PlanState *planstate,
 
 extern void ParallelQueryMain(dsm_segment *seg, shm_toc *toc);
 
+typedef void (*ParallelQueryMain_hook_type)(shm_toc *toc);
+extern PGDLLIMPORT ParallelQueryMain_hook_type ParallelQueryMain_hook;
+
+typedef void (*ExecInitParallelPlan_hook_type)(EState *estate, ParallelContext *pcxt, bool estimate);
```
2jungkook, line 59: 1. What's the meaning of the third argument "estimate"?
2. Why do we need to call this function twice in the engine with `estimate = true/false`?

Please add a comment that can answer those two questions. It should be understandable by readers without looking at Babelfish implementation.

Deepesh125:
> Just curious how parallel workers fetches other details for temp tables relations after permission check ?
> 
> It must need more information about the relation while executing the node with temp table.

Postgres code is written in a way that it will throw an error if any kind of access made to temp table under parallel operations. If we are concerned with operation metadata then it would be already calculated by leader node. Parallel worker just init the plan and pass it to executor run. Do you have any specific example that we should try?

src/backend/executor/execMain.c
Deepesh125: Will add

src/backend/executor/execMain.c
Deepesh125: Instead resolved it by placing hook at more granular level ie, insied ExecCheckOneRelPerms

src/include/executor/execParallel.h
Deepesh125: Added explaination. We first need to estimate the memory requirement before we can actually send data to parallel worker

src/include/access/parallel.h
Deepesh125: As discussed offline, These macros are defined within C file on engine side so its difficult for now

src/backend/executor/execParallel.c
Deepesh125: Moved this check IsBabelfishParallelWorker inside hook implementation to let it be generic

src/backend/executor/execMain.c
Deepesh125: done

src/backend/executor/execMain.c
```diff
@@ -648,6 +650,18 @@ ExecCheckOneRelPerms(RTEPermissionInfo *perminfo)
 	requiredPerms = perminfo->requiredPerms;
 	Assert(requiredPerms != 0);
 
+	/* 
+	 * Some extension may want to avoid permission check for some special
```
surendravishnoi: Let us make comment specific to babelfish, this check no longer applies to anything outside BBF

src/backend/executor/execMain.c
```diff
@@ -648,6 +650,18 @@ ExecCheckOneRelPerms(RTEPermissionInfo *perminfo)
...
+	 * cases. For example, Babelfish temp table is implemented using ENR
+	 * which is not shared with parallel worker and parallel operations
+	 * are not allowed for temp table in Postgres. So give extension a
+	 * chance to skip permission check for such use cases.
+	 */
+	if (IsBabelfishParallelWorker() &&
+		ExecCheckOneRelPerms_hook &&
```
surendravishnoi: can we put extension hook check first to be more safe

Deepesh125: what difference will it make?

kuntalghosh: Shouldn't the parallel worker check be in the extension?

src/backend/executor/execParallel.c
```diff
@@ -683,6 +686,12 @@ ExecInitParallelPlan(PlanState *planstate, EState *estate,
 						   mul_size(PARALLEL_TUPLE_QUEUE_SIZE, pcxt->nworkers));
 	shm_toc_estimate_keys(&pcxt->estimator, 1);
 
+	/* Let extension estimate a dynamic shared memory needed to communicate additional context */
+	if (ExecInitParallelPlan_hook)
+	{
+		(*ExecInitParallelPlan_hook)(estate, pcxt, true);
+	}
```
kuntalghosh: Single line conditional code doesn't need to be surrounded with brackets.

surendravishnoi (approved):

src/backend/executor/execParallel.c
Deepesh125: done

kuntalghosh (approved):

PR #582, opened 2025-06-02
Support empty string handling for date and time datatypes

manisha-deshpande:
Add TSQL-compatible handling of empty string ('') for date and time types, defaulting to '1900-01-01' for date and time, similar to datetime, datetime2, datetimeoffset, smalldatetime.

Task: BABEL-3433

### Description

In SQL Server, when an empty string ('') is cast to `date` or `time` datatype, it should be return a default value - '1900-01-01' for date and '00:00:00.0000000' for time. Currently, Babelfish raises a syntax error for these cases. This change improves TSQL compatibility by matching SQL Server's behavior for empty string handling in date and time datatypes, while maintaining existing behavior for non-empty inputs.

```
1> create table bug(dt date null, tm time(3) null);
2> insert into bug values ('', '');
go
invalid input syntax for type date: ""
invalid input syntax for type time: ""
```
 
### Changes:
- Modified `date_in()` and `time_in()` functions in PostgreSQL engine to check for empty input string before ParseDateTime() and if in _**TSQL dialect**_, prepare date and time default values adjusting their time struct and converting to respective return type Datums (by calling `date2j` and `tm2time` respectively). 
  - This approach mirrors the empty string handling logic in the babelfish_common package for **datetime**, **datetime2**, **datetimeoffset**, and **smalldatetime** datatypes through the use of `initializeToDefaultDatetime()`.
- Added comprehensive test coverage in the Babelfish JDBC test suite to verify empty string handling across different scenarios including direct casts, table insert operation, type conversions, and formatting functions.

### Issues Resolved

BABEL-3433: Support for empty input string handling in date and time datatypes
 
### Check List


src/backend/utils/adt/date.c
```diff
@@ -127,6 +128,19 @@ date_in(PG_FUNCTION_ARGS)
 	char		workbuf[MAXDATELEN + 1];
 	DateTimeErrorExtra extra;
 
+	/*
+	 * Set input to default '1900-01-01' if empty string encountered
+	 */
+	if ((*str == '\0') && (sql_dialect == SQL_DIALECT_TSQL))
```
tanscorpio7: We always put dialect check / hook check as the first thing in the condition.

tanscorpio7 (changes requested):

src/backend/utils/adt/date.c
RIC06X: +1

src/backend/utils/adt/date.c
```diff
@@ -1392,6 +1406,17 @@ time_in(PG_FUNCTION_ARGS)
 	int			ftype[MAXDATEFIELDS];
 	DateTimeErrorExtra extra;
 
+	/*
+	 * Set input to default '00:00:00.0000000' if empty string encountered
+	 */
+	if ((*str == '\0') && (sql_dialect == SQL_DIALECT_TSQL))
```
RIC06X: also please have the dialect check first here.

src/backend/utils/adt/date.c
manisha-deshpande: Flipped the order. Tests passed locally. It is required to rerun the extensions-side PR tests?

src/backend/utils/adt/date.c
```diff
@@ -127,6 +128,19 @@ date_in(PG_FUNCTION_ARGS)
 	char		workbuf[MAXDATELEN + 1];
 	DateTimeErrorExtra extra;
 
+	/*
+	 * Set input to default '1900-01-01' if empty string encountered
+	 */
+	if (sql_dialect == SQL_DIALECT_TSQL && (*str == '\0'))
```
Anikait143: This condition would not effectively check for empty string consisting of whitespaces. For eg:
```
select CAST('    ' as date)
```
We may consider trimming these spaces first.

src/backend/utils/adt/date.c
Anikait143, line 1: We should also consider fixing this for CASTs from string datatypes to Date/Time.

src/backend/utils/adt/date.c
manisha-deshpande: Good point. Considered this, but left it for now to remain consistent with the other date types (datetime2, datetime, datetimeoffset) that don't check or trim blank strings and return a syntax error:
```
select CAST('   ' as datetimeoffset)
invalid input syntax for type timestamp with time zone: "   "
```
It might be good to expand the scope from just date/time datatypes and add the whitespace trimming logic for the other date types as well. Shall I proceed with those changes in this pr itself?

src/backend/utils/adt/date.c
manisha-deshpande: Agreed. Can you help me understand why the sql_dialect check fails for `CAST(CAST('' AS VARCHAR) AS DATE/TIME)`
- sql_dialect=1 when `date_in` is invoked implicitly
- sql_dialect=0 when `varchar2date` invokes `date_in`

tanscorpio7: dialect is changed to PG when executing non-tsql functions. In this case varchar2date.

I think we will need to fix this in two places instead of date_in()
1. `tsql_coerce_string_literal_hook` for direct constants casted to datetime types.
2. In cast functions while casting from other datatypes

Anikait143: Yes, it's Better to have these changes in extension itself as tanzeel mentioned, we can have handlings at two seperate places for string literals and casts respectively.

src/backend/utils/adt/date.c
Anikait143: Can we check the effort to fix other date/time datatypes as well. I think we should fix all of them if the effort is reasonable.

manisha-deshpande:
Fix moved to extension side: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3807

PR #585, opened 2025-06-05
Support for weak binding views to allow dropping of underlying objects

R4hul04:
### Description

This PR introduces improvements to Babelfish's view handling, focusing on weak binding views. The changes enable more SQL Server-compatible behavior for view dependencies.

This PR introduces the following changes:
1. Added hook in DROP codepath to allow dropping underlying objects (tables, functions, views) referenced by weak binding views without errors, while maintaining the view structure
2. Implemented a view repair mechanism hook that automatically fixes broken views when their referenced objects are recreated, eliminating the need for manual view recreation

Implementation Details
1. Added a new flags `BBF_VIEW_DEF_FLAG_IS_BROKEN` to track views with missing dependencies and `BBF_VIEW_DEF_FLAG_IS_WEAK_VIEW` to identify view's schemabinding property in `babelfish_view_def` catalog
2. Implemented hooks in object drop operations to detect and mark dependent views as broken
4. Added preQueryRewrtie hook for view repair mechanism that activates when a broken view is accessed during the rewriting
5. Wrapper function for DefineVirtualRelation
 
### Issues Resolved

BABEL-1660

Signed-off-by: Rahul Parande [rparande@amazon.com](mailto:rparande@amazon.com)
 
### Check List

### DIFF
--- a/src/backend/commands/dropcmds.c
+++ b/src/backend/commands/dropcmds.c
@@ -136,7 +136,7 @@ RemoveObjects(DropStmt *stmt)
 
 		if (object_access_hook && (stmt->removeType == OBJECT_FUNCTION) && sql_dialect == SQL_DIALECT_TSQL)
 		{
-			InvokeObjectDropHook(ProcedureRelationId,address.objectId,0);
+			InvokeObjectDropHook(ProcedureRelationId,address.objectId,-1);
 		}
 	}
 
--- a/src/backend/commands/dropcmds.c
+++ b/src/backend/commands/dropcmds.c
@@ -136,16 +136,9 @@ RemoveObjects(DropStmt *stmt)
 
 		/* Check for strong views and handle weak views */
 		if ((stmt->removeType == OBJECT_FUNCTION) && view_dependency_hook)
-		{
-			Relation depRel = table_open(DependRelationId, AccessShareLock);
-			
-			if (!(*view_dependency_hook)(&address, depRel, NULL))
-			{
-				table_close(depRel, AccessShareLock);
+		{			
+			if (!(*view_dependency_hook)(&address, NULL, NULL))
 				continue;
-			}
-			
-			table_close(depRel, AccessShareLock);
 		}
 	}
 
--- a/src/backend/commands/dropcmds.c
+++ b/src/backend/commands/dropcmds.c
@@ -19,9 +19,11 @@
 #include "catalog/dependency.h"
 #include "catalog/namespace.h"
 #include "catalog/objectaddress.h"
+#include "catalog/pg_depend_d.h"
 #include "catalog/pg_namespace.h"
 #include "catalog/pg_proc.h"
 #include "commands/defrem.h"
+#include "commands/tablecmds.h"
 #include "miscadmin.h"
 #include "parser/parser.h"
 #include "parser/parse_type.h"
@@ -131,6 +133,20 @@ RemoveObjects(DropStmt *stmt)
 			table_close(relation, NoLock);
 
 		add_exact_object_address(&address, objects);
+
+		/* Check for strong views and handle weak views */
+		if ((stmt->removeType == OBJECT_FUNCTION) && view_dependency_hook)
+		{
+			Relation depRel = table_open(DependRelationId, AccessShareLock);
+			
+			if (!(*view_dependency_hook)(&address, depRel, NULL))
+			{
+				table_close(depRel, AccessShareLock);
+				continue;
+			}
+			
+			table_close(depRel, AccessShareLock);
+		}
 	}
 
 	/* Here we really delete them. */
--- a/src/backend/commands/dropcmds.c
+++ b/src/backend/commands/dropcmds.c
@@ -18,6 +18,7 @@
 #include "access/xact.h"
 #include "catalog/dependency.h"
 #include "catalog/namespace.h"
+#include "catalog/objectaccess.h"
 #include "catalog/objectaddress.h"
 #include "catalog/pg_namespace.h"
 #include "catalog/pg_proc.h"
@@ -133,11 +134,9 @@ RemoveObjects(DropStmt *stmt)
 
 		add_exact_object_address(&address, objects);
 
-		/* Check for strong views and handle weak views */
-		if ((stmt->removeType == OBJECT_FUNCTION) && view_dependency_hook)
-		{			
-			if (!(*view_dependency_hook)(&address, NULL, NULL))
-				continue;
+		if (object_access_hook && (stmt->removeType == OBJECT_FUNCTION))
+		{
+			InvokeObjectDropHook(ProcedureRelationId,address.objectId,0);
 		}
 	}
 
--- a/src/backend/commands/dropcmds.c
+++ b/src/backend/commands/dropcmds.c
@@ -19,7 +19,6 @@
 #include "catalog/dependency.h"
 #include "catalog/namespace.h"
 #include "catalog/objectaddress.h"
-#include "catalog/pg_depend_d.h"
 #include "catalog/pg_namespace.h"
 #include "catalog/pg_proc.h"
 #include "commands/defrem.h"
--- a/src/backend/commands/dropcmds.c
+++ b/src/backend/commands/dropcmds.c
@@ -134,7 +134,7 @@ RemoveObjects(DropStmt *stmt)
 
 		add_exact_object_address(&address, objects);
 
-		if (object_access_hook && (stmt->removeType == OBJECT_FUNCTION))
+		if (object_access_hook && (stmt->removeType == OBJECT_FUNCTION) && sql_dialect == SQL_DIALECT_TSQL)
 		{
 			InvokeObjectDropHook(ProcedureRelationId,address.objectId,0);
 		}
--- a/src/backend/commands/dropcmds.c
+++ b/src/backend/commands/dropcmds.c
@@ -136,7 +136,7 @@ RemoveObjects(DropStmt *stmt)
 
 		if (object_access_hook && (stmt->removeType == OBJECT_FUNCTION) && sql_dialect == SQL_DIALECT_TSQL)
 		{
-			InvokeObjectDropHook(ProcedureRelationId,address.objectId,-1);
+			InvokeObjectDropHook(ProcedureRelationId, address.objectId, -1);
 		}
 	}
 
--- a/src/backend/commands/tablecmds.c
+++ b/src/backend/commands/tablecmds.c
@@ -138,7 +138,6 @@ InvokePreDropColumnHook_type InvokePreDropColumnHook = NULL;
 check_extended_attoptions_hook_type check_extended_attoptions_hook = NULL;
 find_attr_by_name_from_column_def_list_hook_type
 	find_attr_by_name_from_column_def_list_hook = NULL;
-view_dependency_hook_type view_dependency_hook = NULL;
 
 /*
  * State information for ALTER TABLE
--- a/src/backend/commands/tablecmds.c
+++ b/src/backend/commands/tablecmds.c
@@ -138,6 +138,7 @@ InvokePreDropColumnHook_type InvokePreDropColumnHook = NULL;
 check_extended_attoptions_hook_type check_extended_attoptions_hook = NULL;
 find_attr_by_name_from_column_def_list_hook_type
 	find_attr_by_name_from_column_def_list_hook = NULL;
+view_dependency_hook_type view_dependency_hook = NULL;
 
 /*
  * State information for ALTER TABLE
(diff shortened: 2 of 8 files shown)


src/backend/commands/dropcmds.c
```diff
@@ -131,6 +133,20 @@ RemoveObjects(DropStmt *stmt)
 			table_close(relation, NoLock);
 
 		add_exact_object_address(&address, objects);
+
+		/* Check for strong views and handle weak views */
```
forestkeeper: Why not move the open table into the hook implement ?

src/bin/pg_dump/pg_dump.c
```diff
@@ -15822,6 +15822,50 @@ dumpTable(Archive *fout, const TableInfo *tbinfo)
...
+/*
+ * Create a dummy AS clause for a view.  This is used when the real view
+ * definition has to be postponed because of circular dependencies.
+ * We must duplicate the view's external properties -- column names and types
+ * (including collation) -- so that it works for subsequent references.
+ *
+ * This returns a new buffer which must be freed by the caller.
+ */
+static PQExpBuffer
+createDummyViewAsClause(Archive *fout, const TableInfo *tbinfo)
```
forestkeeper: Why not reuse the createDummyViewAsClause def from community ?

src/bin/pg_dump/pg_dump.c
```diff
@@ -15833,6 +15877,7 @@ createViewAsClause(Archive *fout, const TableInfo *tbinfo)
 {
 	PQExpBuffer query = createPQExpBuffer();
 	PQExpBuffer result = createPQExpBuffer();
+	PQExpBuffer dummyResult = createPQExpBuffer();
```
forestkeeper: Dummyresult doesn't need to be created as createPQExpBuffer.

src/bin/pg_dump/pg_dump.c
```diff
@@ -15869,50 +15927,6 @@ createViewAsClause(Archive *fout, const TableInfo *tbinfo)
 	return result;
 }
 
-/*
- * Create a dummy AS clause for a view.  This is used when the real view
```
forestkeeper: Why move the whole function definition ?

src/backend/commands/dropcmds.c
```diff
@@ -19,9 +19,11 @@
 #include "catalog/dependency.h"
 #include "catalog/namespace.h"
 #include "catalog/objectaddress.h"
+#include "catalog/pg_depend_d.h"
```
forestkeeper: Why adding this .h ?

src/bin/pg_dump/pg_dump.c
R4hul04: Added `createDummyViewAsClause` inside `createViewAsClause`. Hence we needed to define createDummyViewAsClause before createViewAsClause.

src/bin/pg_dump/pg_dump.c
R4hul04: We are reusing the same function. We are only moving it before `createViewAsClause`, since we need  `createDummyViewAsClause` defined before the function call in `createViewAsClause`

src/bin/pg_dump/pg_dump.c
R4hul04: Removed `dummyResult` variable instead using the existing variable `result`

src/backend/commands/dropcmds.c
R4hul04: Removed pg_depend_d.h

src/bin/pg_dump/pg_dump.c
forestkeeper: We should not move this function's place in community code, since after moving, the follow up new commits diff from community will be lost for this function.

src/bin/pg_dump/pg_dump.c
forestkeeper: We should find a alternative way rather than moving this function.

src/backend/rewrite/rewriteHandler.c
```diff
@@ -4431,6 +4431,14 @@ QueryRewrite(Query *parsetree)
...
 
 	/*
+     * Step 0 (Babelfish extension)
+     *
+     * If this is a view with broken rules, try to repair it
+     * using the definition from babelfish_view_def
+     */
+	if (view_repair_hook)
+		(*view_repair_hook)(parsetree);
+	/*
```
tanscorpio7: Should we use a more generic name. pre_query_rewrite_hook

src/backend/commands/tablecmds.c
```diff
@@ -1676,6 +1677,13 @@ RemoveRelations(DropStmt *drop)
 		{
 			InvokeObjectDropHook(RelationRelationId,relOid,0);
 		}
+
+		/* Check for strong views and handle weak views */
+		if ((drop->removeType == OBJECT_TABLE || drop->removeType == OBJECT_VIEW) && view_dependency_hook)
+		{			
+			if (!(*view_dependency_hook)(&obj, NULL, NULL))
+				continue;
+		}
```
tanscorpio7: Let us move all comments and conditions inside the hook. Also would naming it `pre_drop_object_hook` make sense ?

tanscorpio7 (changes requested):

src/bin/pg_dump/pg_dump.c
R4hul04: Agreed. 
Added the function declaration of `createDummyViewAsClause` at the top instead of moving the whole function body.

src/bin/pg_dump/pg_dump.c
R4hul04: Added the function declaration of `createDummyViewAsClause` at the top instead of moving the whole function body.

src/backend/commands/tablecmds.c
```diff
@@ -1676,6 +1677,13 @@ RemoveRelations(DropStmt *drop)
 		{
 			InvokeObjectDropHook(RelationRelationId,relOid,0);
 		}
+
+		/* Check for strong views and handle weak views */
+		if ((drop->removeType == OBJECT_TABLE || drop->removeType == OBJECT_VIEW) && view_dependency_hook)
```
forestkeeper: Can we move this part also into the hook : (drop->removeType == OBJECT_TABLE || drop->removeType == OBJECT_VIEW) ?

src/backend/commands/dropcmds.c
```diff
@@ -131,6 +132,13 @@ RemoveObjects(DropStmt *stmt)
 			table_close(relation, NoLock);
 
 		add_exact_object_address(&address, objects);
+
+		/* Check for strong views and handle weak views */
+		if ((stmt->removeType == OBJECT_FUNCTION) && view_dependency_hook)
+		{			
+			if (!(*view_dependency_hook)(&address, NULL, NULL))
+				continue;
+		}
```
kuntalghosh: Let's make the hook generic and add all TSQL related check within the hook function. We're not worried about hook overhead in a DDL execution. Also, there's already a hook for this task - InvokeObjectDropHook. Let's plugin this hook here.

src/backend/commands/tablecmds.c
kuntalghosh: Let's use InvokeObjectDropHook which already there in the function.

src/backend/commands/view.c
```diff
@@ -381,6 +384,9 @@ DefineView(ViewStmt *stmt, const char *queryString,
 
 	viewParse = parse_analyze_fixedparams(rawstmt, queryString, NULL, 0, NULL);
 
+	if (check_view_dependencies_hook)
+			(*check_view_dependencies_hook)(viewParse);
+
```
kuntalghosh: The function parse_analyze_fixedparams already has a hook inside it - post_parse_analyze_hook(). Let's use that.

src/backend/rewrite/rewriteHandler.c
kuntalghosh: +1
Let's use the name `pre_QueryRewrite_hook`.

src/bin/pg_dump/pg_dump.c
```diff
@@ -15856,8 +15856,19 @@ createViewAsClause(Archive *fout, const TableInfo *tbinfo)
...
+		 * Instead of failing, create a dummy SELECT that preserves the structure
+		 */
+		PQclear(res);
+		destroyPQExpBuffer(query);
+		
+		/* Use createDummyViewAsClause to generate a compatible structure */
+		result = createDummyViewAsClause(fout, tbinfo);
+		
+		return result;
+	}
```
kuntalghosh: Whenever we modify in pg_dump, we make sure that this is a Babelfish Database. Check the functions with `bbf` prefix. Ideally, the logic should be,
```
if (bbfShouldDumpViews(Archive *fout, const TableInfo *tbinfo))
{
    createTSQLViewAsClause()
    return;
}
```
Within `createTSQLViewAsClause`, we should do all the handlings.

kuntalghosh (dismissed):

src/bin/pg_dump/pg_dump.c
R4hul04: Do we need to create our own `createTSQLViewAsClause` function in `dump_babel_utils.c` which will be called specifically for babelfish database. This function will be called from `createViewAsClause` in `pg_dump.c` ?

src/backend/rewrite/rewriteHandler.c
R4hul04: Renamed the hook to `pre_QueryRewrite_hook`

src/backend/commands/tablecmds.c
R4hul04: Moved implementation to `InvokeObjectDropHook` instead.

src/backend/commands/tablecmds.c
R4hul04: Moved implementation to use `InvokeObjectDropHook` instead.

src/backend/commands/dropcmds.c
R4hul04: Used `InvokeObjectDropHook` instead of `view_dependency_hook`.

src/backend/commands/view.c
R4hul04: Updated `post_parse_analyze_hook`() with the required logic.

src/bin/pg_dump/pg_dump.c
tanscorpio7: I thought we will insert a dummy view query into pg_rewrite at the time of DROP which will avoid the need to alter dump restore. The dummy query can be prepared according to the columns.
We should avoid keeping empty string as rewrite rule since engine code assumes that is an impossible case.

R4hul04: Reverted back to original code and inserted dummy_query in pg_rewrite during DROP, which takes care of pg_dump for broken views

forestkeeper (dismissed):

src/bin/pg_dump/pg_dump.c
```diff
@@ -345,7 +345,6 @@ static TableInfo *getRootTableInfo(const TableInfo *tbinfo);
 static bool forcePartitionRootLoad(const TableInfo *tbinfo);
 static void read_dump_filters(const char *filename, DumpOptions *dopt);
 
-
```
tanscorpio7: Let us make sure we do not commit any unnecessary changes, they will end up causing merge conflicts during community merge.

src/include/commands/view.h
```diff
@@ -27,4 +27,8 @@ extern PGDLLEXPORT store_view_definition_hook_type	store_view_definition_hook;
 
 typedef void (*inherit_view_constraints_from_table_hook_type) (ColumnDef  *col, Oid tableOid, AttrNumber colId);
 extern PGDLLEXPORT inherit_view_constraints_from_table_hook_type inherit_view_constraints_from_table_hook;
+
+extern ObjectAddress DefineVirtualRelation(RangeVar *relation, List *tlist, bool replace,
+					  						List *options, Query *viewParse);
+
```
tanscorpio7: We do not directly mark the function as extern instead. Instead declare a new function in view.h/.c and exposed it through that. For example see `SPI_get_depth()`

src/backend/rewrite/rewriteHandler.c
```diff
@@ -4431,6 +4431,14 @@ QueryRewrite(Query *parsetree)
 	Assert(parsetree->canSetTag);
 
 	/*
+     * Step 0 (Babelfish extension)
+     *
+     * If this is a view with broken rules, try to repair it
+     * using the definition from babelfish_view_def
+     */
```
tanscorpio7: Let's remove the comments from here.
And an indentation fix `hook)(pa` -> `hook) (pa`

src/backend/rewrite/rewriteHandler.c
```diff
@@ -4403,7 +4404,6 @@ RewriteQuery(Query *parsetree, List *rewrite_events, int orig_rt_length)
 	return rewritten;
 }
 
-
```
tanscorpio7: Not needed change.

src/backend/commands/dropcmds.c
```diff
@@ -131,6 +133,11 @@ RemoveObjects(DropStmt *stmt)
 			table_close(relation, NoLock);
 
 		add_exact_object_address(&address, objects);
+
+		if (object_access_hook && (stmt->removeType == OBJECT_FUNCTION) && sql_dialect == SQL_DIALECT_TSQL)
+		{
+			InvokeObjectDropHook(ProcedureRelationId,address.objectId,-1);
```
tanscorpio7: Indentation: spaces between args.

tanscorpio7 (changes requested):

src/include/commands/view.h
R4hul04: Added wrapper function `bbf_define_virtual_relation`

tanscorpio7 (approved):

forestkeeper (approved):

PR #588, opened 2025-06-25
Prevent Babelfish database via Psql without call sys.remove_babelfish() procedure 

kelepcera:
### Description

This change introduces a safeguard in the PostgreSQL engine to prevent accidental or improper deletion of the Babelfish-managed database (babelfish_db). Currently, if the database is dropped without first calling sys.remove_babelfish(), it leads to an inconsistent system state where Babelfish cannot be re-initialized due to lingering metadata (e.g., reserved roles, configurations).

To prevent this, the patch adds a check inside the database deletion logic to determine whether the target database is a Babelfish instance. If so, it validates certain preconditions—such as whether the sys.remove_babelfish() cleanup function was called—and only then allows the deletion. Even privileged users like root are restricted if the conditions are not met.

This helps ensure that Babelfish is uninstalled properly and avoids corrupted states in environments like RDS or custom Babelfish deployments
 
### Issues Resolved

- Prevents improper deletion of Babelfish-managed databases
- Mitigates reinitialization errors due to existing reserved roles like sysadmin
- Aligns database deletion process with Babelfish's expected lifecycle

### DIFF
--- a/src/backend/commands/dbcommands.c
+++ b/src/backend/commands/dbcommands.c
@@ -139,6 +139,8 @@ static void CreateDirAndVersionFile(char *dbpath, Oid dbid, Oid tsid,
 static void CreateDatabaseUsingFileCopy(Oid src_dboid, Oid dst_dboid,
 										Oid src_tsid, Oid dst_tsid);
 static void recovery_create_dbdir(char *path, bool only_tblspc);
+static bool is_babelfish_ownership_enabled(ArrayType *array);
+static bool has_babelfish_restriction(Oid target_db_id);
 
 /*
  * Create a new database using the WAL_LOG strategy.
@@ -1627,6 +1629,109 @@ createdb_failure_callback(int code, Datum arg)
 	remove_dbtablespaces(fparms->dest_dboid);
 }
 
+/*
+ * This function controls if babelfishpg_tsql.enable_ownership_structure
+ * was set. If so it will return true otherwise return false.
+ */
+static bool
+is_babelfish_ownership_enabled(ArrayType *options_array)
+{
+	int option_index;
+	static const char *target_option_name = "babelfishpg_tsql.enable_ownership_structure";
+	int target_option_name_len = strlen(target_option_name);
+	bool ownership_enabled = false;
+
+	for (option_index = 1; option_index <= ARR_DIMS(options_array)[0]; option_index++)
+	{
+		Datum option_datum;
+		bool is_option_null = false;
+		char *option_text = NULL;
+		char *parsed_option_name = NULL;
+		char *parsed_option_value = NULL;
+
+		option_datum = array_ref(options_array, 1, &option_index,
+								 -1 /* varlenarray */,
+								 -1 /* TEXT's typlen */,
+								 false /* TEXT's typbyval */,
+								 TYPALIGN_INT /* TEXT's typalign */,
+								 &is_option_null);
+
+		if (is_option_null)
+			continue;
+
+		option_text = TextDatumGetCString(option_datum);
+
+		ParseLongOption(option_text, &parsed_option_name, &parsed_option_value);
+		if (strncmp(parsed_option_name, target_option_name, target_option_name_len) == 0 &&
+			strncmp(parsed_option_value, "true", 4) == 0)
+		{
+			pfree(parsed_option_name);
+			pfree(parsed_option_value);
+			pfree(option_text);
+			ownership_enabled = true;
+			break;
+		}
+
+		pfree(parsed_option_name);
+		pfree(parsed_option_value);
+		pfree(option_text);
+	}
+
+	return ownership_enabled;
+}
+
+/*
+* This function will check if database to be removed is Babelfish exclusive one.
+* If so it will check whether some preconditions were fullfilled or not.
+*/
+static bool
+has_babelfish_restriction(Oid target_db_id) 
+{
+	Relation relsetting;
+	ScanKeyData scankey[2];
+	SysScanDesc scan;
+	HeapTuple tuple;
+	bool has_bbf_md;
+
+	has_bbf_md = false;
+
+	relsetting = table_open(DbRoleSettingRelationId, AccessShareLock);
+
+	ScanKeyInit(&scankey[0],
+				Anum_pg_db_role_setting_setdatabase,
+				BTEqualStrategyNumber, F_OIDEQ,
+				ObjectIdGetDatum(target_db_id));
+	ScanKeyInit(&scankey[1],
+				Anum_pg_db_role_setting_setrole,
+				BTEqualStrategyNumber, F_OIDEQ,
+				ObjectIdGetDatum(InvalidOid));
+
+	scan = systable_beginscan(relsetting, DbRoleSettingDatidRolidIndexId,
+								true, NULL, 2, scankey);
+	tuple = systable_getnext(scan);
+	if (HeapTupleIsValid(tuple))
+	{
+		bool isnull;
+		Datum datum;
+		ArrayType *configs = NULL;
+
+		datum = heap_getattr(tuple, Anum_pg_db_role_setting_setconfig,
+								RelationGetDescr(relsetting), &isnull);
+		if (!isnull)
+			configs = DatumGetArrayTypeP(datum);
+
+		if (configs && is_babelfish_ownership_enabled(configs))
+			has_bbf_md = true;
+	}
+	systable_endscan(scan);
+	table_close(relsetting, AccessShareLock);
+
+	if (has_bbf_md)
+		return true;
+
+	return false;
+}
+
 
 /*
  * DROP DATABASE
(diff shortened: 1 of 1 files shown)


kelepcera:
This also issue report at Babelfish Extension side created by me. You can reach out the details of this bug via the link below.

https://github.com/babelfish-for-postgresql/babelfish_extensions/issues/3875

tanscorpio7:
Please check `tdsutils_ProcessUtility` function. There is already `handle_dropdb` restriction, we can just move it above the superuser bypass check.

kelepcera:
> Please check `tdsutils_ProcessUtility` function. There is already `handle_dropdb` restriction, we can just move it above the superuser bypass check.

You're absolutely right that tdsutils_ProcessUtility includes a handle_dropdb() check. However, the root of the problem lies in the fact that this function is only invoked for TDS-based clients (e.g., sqlcmd, SSMS)—not for generic SQL clients like psql.

Here’s what happens:

When connecting via psql and executing:
`DROP DATABASE babelfish_db;`

…the PostgreSQL engine bypasses tdsutils_ProcessUtility() and routes the command directly to the default PostgreSQL ProcessUtility() handler.

This occurs because the command is recognized as standard SQL, not T-SQL, and thus Babelfish's custom dropdb hook is never triggered.
From the code, we can observe that:
Inside tdsutils_ProcessUtility(), there’s a dialect check that routes the execution based on whether the statement comes from the TDS protocol.

f it's not T-SQL, Babelfish does not apply its custom logic and defers to PostgreSQL’s native command execution path.

Here is a code path illustration (see screenshot):
image


So, even if we move handle_dropdb() above the superuser check, it still won’t help in the psql case, because tdsutils_ProcessUtility() itself is never called.

That’s why my fix adds a safeguard directly to the PostgreSQL engine-level dropdb() logic to catch this edge case regardless of client type.

You can reach its details from my issue link: https://github.com/babelfish-for-postgresql/babelfish_extensions/issues/3875

tanscorpio7:
`tdsutils_ProcessUtility` is babelfishpg_tds extension hook and we add babelfishpg_tds to shared_preload_libraries. So all its hook will be loaded on database restart. So I think even psql connections will go through this hook.

kelepcera:
> `tdsutils_ProcessUtility` is babelfishpg_tds extension hook and we add babelfishpg_tds to shared_preload_libraries. So all its hook will be loaded on database restart. So I think even psql connections will go through this hook.

You're right that tdsutils_ProcessUtility is registered via shared_preload_libraries, and I initially assumed this meant that it would intercept all DROP DATABASE statements—including those issued through psql.

Following your suggestion, I modified the code as below and moved the handle_dropdb() call above the superuser check:

image

To verify that the hook was working, I also added breakpoints inside handle_dropdb().

However, when I attempted to drop the Babelfish database via psql, the execution never reached handle_dropdb() at all. Instead, the default PostgreSQL dropdb() function was triggered.

At first I thought maybe I made a mistake in testing, so I double-checked:

Recompiled and restarted PostgreSQL

Verified that babelfishpg_tds was in shared_preload_libraries

Confirmed the breakpoint was active and code was instrumented properly

Then, I tried the same test via SSMS — and in that case, the handle_dropdb() function was correctly invoked, and the deletion was blocked as expected.

So at this point, it seems like tdsutils_ProcessUtility() is only invoked for TDS-based clients, and not for standard SQL clients like psql.

This seems consistent with the Babelfish dialect/protocol routing logic, which bypasses the Babelfish TDS layer for standard SQL input.

If I’m missing something here, happy to dig deeper — but based on what I’m observing, protecting DROP DATABASE across all clients requires a check in the core engine.

This is also the line which I changed to load babelfish library in "postgresql.conf" file. 
image


This is the output of SSMS. If I tried with psql I could drop it.
image

tanscorpio7:
I am actually confident about `tdsutils_ProcessUtility` hitting from psql client because this is what we use for restricting DROP ROLE from psql endpoint on babelfish managed roles.

I would suggest figuring out where `tdsutils_ProcessUtility` exits for DROP DATBASE command from postgres endpoint.

src/backend/commands/dbcommands.c
```diff
@@ -1627,6 +1629,109 @@ createdb_failure_callback(int code, Datum arg)
...
+		if (is_option_null)
+			continue;
+
+		option_text = TextDatumGetCString(option_datum);
+
+		ParseLongOption(option_text, &parsed_option_name, &parsed_option_value);
+		if (strncmp(parsed_option_name, target_option_name, target_option_name_len) == 0 &&
+			strncmp(parsed_option_value, "true", 4) == 0)
+		{
+			pfree(parsed_option_name);
```
shalinilohia50, line 1668: These pfree are duplicated.

shalinilohia50:
> I am actually confident about `tdsutils_ProcessUtility` hitting from psql client because this is what we use for restricting DROP ROLE from psql endpoint on babelfish managed roles.
> 
> I would suggest figuring out where `tdsutils_ProcessUtility` exits for DROP DATABASE command from postgres endpoint.

+1

kelepcera:
> I am actually confident about `tdsutils_ProcessUtility` hitting from psql client because this is what we use for restricting DROP ROLE from psql endpoint on babelfish managed roles.
> 
> I would suggest figuring out where `tdsutils_ProcessUtility` exits for DROP DATBASE command from postgres endpoint.

Thanks again for the guidance. After adding several log lines and debugging deeper, I was able to partially trace what’s happening under the hood.

I can now say with 100% confidence that handle_dropdb() is never reached when executing DROP DATABASE via psql. Instead, the PostgreSQL native drop database handler is invoked.

You're absolutely right that tdsutils_ProcessUtility() is being called — I’ve verified that. However, within that function, the flow always enters this if block due to the user being a superuser, and as a result, it never reaches the switch statement where handle_dropdb() is called:

image

To validate this, I initialized Babelfish and created a test database (testdb1). Then I issued:
DROP DATABASE testdb1;

I logged each critical branch and collected the output. I'll now share the relevant log lines that confirm this behavior. Also the code will miss drop_bbf_db() function since the dialect is not TSQL. Because I use psql. 

image
image
image
image
image
image
image
image
image
image

shalinilohia50:
> I can now say with 100% confidence that handle_dropdb() is never reached when executing DROP DATABASE via psql. Instead, the PostgreSQL native drop database handler is invoked.
>
> You're absolutely right that tdsutils_ProcessUtility() is being called — I’ve verified that. However, within that function, the flow always enters this if block due to the user being a superuser, and as a result, it never reaches the switch statement where handle_dropdb() is called:

Looks like you are trying with superuser via PSQL. If yes, then all the security checks for superuser are bypassed. Hence, it's not going to hit `handle_dropdb`. We should move `handle_dropdb` before that superuser check to restrict Babelfish db (physical database) to be dropped.



> To validate this, I initialized Babelfish and created a test database (testdb1). Then I issued:
> DROP DATABASE testdb1;
> 
> I logged each critical branch and collected the output. I'll now share the relevant log lines that confirm this behavior. Also the code will miss drop_bbf_db() function since the dialect is not TSQL. Because I use psql.

Is this a physical(database where babelfish is initialised) or logical(database which is created via TSQL) database?

kelepcera:
Thanks for the clarification — this makes the flow much clearer now.

> Is this a physical(database where babelfish is initialised) or logical(database which is created via TSQL) database?

Yes, it should be physical database which was created via psql not with SQLCMD and also TSQL if I understood the difference between physical and logical database.

> Looks like you are trying with superuser via PSQL. If yes, then all the security checks for superuser are bypassed. Hence, it's not going to hit `handle_dropdb`. We should move `handle_dropdb` before that superuser check to restrict Babelfish db (physical database) to be dropped.

You're absolutely right: I was testing via `psql` using a superuser account, and the flow inside `tdsutils_ProcessUtility()` enters the superuser short-circuit block early — which explains why `handle_dropdb()` is never called in that scenario.

I now understand that to enforce protection even for superusers, the call to `handle_dropdb()` must be moved *before* the `if (superuser())` check.

Thanks again for helping clarify this — I’ll rework the patch accordingly.

kelepcera:
Hello again, I have tried your suggested solution and it worked. I have just simply moved only the statemet that run handle_dropdb function to upper side of super user check in tdsutils_ProcessUtility function. Whenever I try to delete Babelfish exclusive database, I encountered this error:
```

jdbc_testdb=# DROP DATABASE jdbc_testdb;
ERROR:  cannot drop active babelfish database
```

So it worked and thank you. I guess I should close this merge request and open new one in Babelfish Extension side. If this is good enough for you then I can open new merge request.

kelepcera:
Hi @tanscorpio7 and @shalinilohia50,

I’ve just created a new pull request for the Babelfish Extension, implementing the solution you suggested. I hope I applied the changes correctly.

New PR: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3888

@tanscorpio7 — Apologies for the earlier confusion. I was focused on different code paths and superuser-related checks, which led me in the wrong direction. Thank you for your guidance and support.

Feel free to close this PR.

tanscorpio7:
Closing in favour of https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3888

PR #606, opened 2025-07-17
Added logs and asserts in temp table code path

ayushdsh:
### Description

Extension changes: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/3971
 
### Issues Resolved

### Check List

### DIFF
--- a/src/backend/catalog/storage.c
+++ b/src/backend/catalog/storage.c
@@ -227,6 +227,14 @@ RelationDropStorage(Relation rel)
 	 */
 	if (IsTsqlTableVariable(rel))
 	{
+		if (unlikely(rel->rd_backend == INVALID_PROC_NUMBER))
+		{
+			/* logging the relation */
+			elog(WARNING, "rd_id = %d, rd_backend = %d, oid = %d, relfilenode = %d, relname = %s, relnamespace = %d", 
+				rel->rd_id, rel->rd_backend, rel->rd_rel->oid, rel->rd_rel->relfilenode, rel->rd_rel->relname.data, rel->rd_rel->relnamespace);
+			/* proceeding to kill the backend as continuing from this point might lead to deletion of unintended rels */
+			elog(FATAL, "Table variable has an invalid backend id which might lead the engine to think this is a permanent relation.");
+		}
 		pending = (PendingRelDelete *)
 		MemoryContextAlloc(TopMemoryContext, sizeof(PendingRelDelete));
 		pending->rlocator = rel->rd_locator;
--- a/src/backend/catalog/storage.c
+++ b/src/backend/catalog/storage.c
@@ -227,7 +227,14 @@ RelationDropStorage(Relation rel)
 	 */
 	if (IsTsqlTableVariable(rel))
 	{
-		Assert(rel->rd_backend != INVALID_PROC_NUMBER);
+		if (rel->rd_backend == INVALID_PROC_NUMBER)
+		{
+			/* logging the relation */
+			elog(WARNING, "rd_id = %d, rd_backend = %d, oid = %d, relfilenode = %d, relname = %s, relnamespace = %d", 
+				rel->rd_id, rel->rd_backend, rel->rd_rel->oid, rel->rd_rel->relfilenode, rel->rd_rel->relname.data, rel->rd_rel->relnamespace);
+			/* proceeding to kill the backend as continuing from this point might lead to deletion of unintended rels */
+			elog(FATAL, "Invalid relation found while dropping a tsql table variable.");
+		}
 		pending = (PendingRelDelete *)
 		MemoryContextAlloc(TopMemoryContext, sizeof(PendingRelDelete));
 		pending->rlocator = rel->rd_locator;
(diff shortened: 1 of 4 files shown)


src/backend/utils/cache/inval.c
```diff
@@ -461,6 +461,12 @@ AddRelcacheInvalidationMessage(InvalidationMsgsGroup *group,
 	msg.rc.dbId = dbId;
 	msg.rc.relId = relId;
 	msg.rc.local_only = (pltsql_get_tsql_enr_from_oid_hook && (*pltsql_get_tsql_enr_from_oid_hook)(relId));
+
+	if (!msg.rc.local_only && GetENRTempTableWithOid(relId))
+	{
+		elog(WARNING, "Relcache invalidation message sent to all backends even though the relation is ENR; relId: %u", relId);
+	}
+
```
tanscorpio7: Could you explain the reasoning behind this logging ?
`pltsql_get_tsql_enr_from_oid_hook` has the same check for `GetENRTempTableWithOid` in it.

src/backend/catalog/storage.c
```diff
@@ -227,6 +227,14 @@ RelationDropStorage(Relation rel)
...
 	if (IsTsqlTableVariable(rel))
 	{
+		if (rel->rd_backend == INVALID_PROC_NUMBER)
+		{
+			/* logging the relation */
+			elog(WARNING, "rd_id = %d, rd_backend = %d, oid = %d, relfilenode = %d, relname = %s, relnamespace = %d", 
+				rel->rd_id, rel->rd_backend, rel->rd_rel->oid, rel->rd_rel->relfilenode, rel->rd_rel->relname.data, rel->rd_rel->relnamespace);
+			/* proceeding to kill the backend as continuing from this point might lead to deletion of unintended rels */
+			elog(FATAL, "Invalid relation found while dropping a tsql table variable.");
+		}
```
tanscorpio7: can this check be inside `IsTsqlTableVariable` instead ?

tanscorpio7 (changes requested):

src/backend/utils/cache/inval.c
ayushdsh: This would be true when temp_oid_buffer_size = 0; in which case we will send an inval msg across backends. So essentially we will log if its an ENR temp table and yet we are sending inval msg across backends

tanscorpio7: Which is the expected behaviour I think, currently we do have SOPs where we set `temp_oid_buffer_size = 0' in which case this log would be very frequent.

src/backend/catalog/storage.c
ayushdsh: Yes, it can be. You're right! I'll rather have it in the function itself

src/backend/utils/cache/inval.c
ayushdsh: I see. In that case, I'll remove it. makes sense.

src/backend/utils/cache/inval.c
```diff
@@ -461,6 +461,7 @@ AddRelcacheInvalidationMessage(InvalidationMsgsGroup *group,
 	msg.rc.dbId = dbId;
 	msg.rc.relId = relId;
 	msg.rc.local_only = (pltsql_get_tsql_enr_from_oid_hook && (*pltsql_get_tsql_enr_from_oid_hook)(relId));
+
```
time-change: nit: no unnecessary changes

src/backend/utils/misc/queryenvironment.c
```diff
@@ -1781,11 +1781,22 @@ bool SIMessageIsForTempTable(const SharedInvalidationMessage *msg)
 /* Simple wrapper for Table Variable checks */
```
time-change: Can we also add logging where the OID hooks are called? ie, `heap_create`, `index_create` etc.

src/backend/utils/misc/queryenvironment.c
```diff
@@ -1781,11 +1781,22 @@ bool SIMessageIsForTempTable(const SharedInvalidationMessage *msg)
...
 bool IsTsqlTableVariable(Relation relation)
 {
-	return sql_dialect == SQL_DIALECT_TSQL 
+	bool isTableVariable = sql_dialect == SQL_DIALECT_TSQL 
 		&& relation
 		&& relation->rd_rel->relpersistence == RELPERSISTENCE_TEMP 
 		&& strlen(relation->rd_rel->relname.data) >= 1 
 		&& relation->rd_rel->relname.data[0] == '@';
+	
+	if (isTableVariable && relation->rd_backend == INVALID_PROC_NUMBER)
```
time-change: why only in the `IsTsqlTableVariable` codepath? why are we checking this every time this is called?

ayushdsh: Isn't it better to check it everytime .. I can wrap it in an `unlikely` block, but wouldn't it be better to have this check at such a bottom layer

time-change (approved):

jsudrik (approved):
Approved.

PR #611, opened 2025-08-07
Ownership Chaining: Support relation references inside views

HarshLunagariya:
### Description

This commit implements the ownership chaining support for table/view object 
references inside view. The basic idea of Ownership chaining is that when 
one object references another, and both have the same owner, then permissions 
are only checked when the top-level object is accessed.

The scope of the current commit is to only support scenarios when parent 
object is view and child objects are table/view. "[parent] view -> [child] 
function" scenarios is out-of-scope for this effort.

Implementation:
- Add infrastructure changes to support ownership tracking:
  * insideView flag in RTEPermissionInfo for view context tracking
  * parentOwnerId and insideView fields in FuncExpr (for future use)
  * walk_view_rule_hook for extension-specific view rule processing and 
     apply view context marking in ApplyRetrieveRule

- Add walker to traverse view definition parse trees:
  * Mark relations and functions inside views for permission checking
  * Set checkAsUser to view_owner for owned relations to enable chaining
  * Track view context using insideView flag
  * Only applies to user-defined objects created in non-shared schemas in TSQL mode

This enables proper ownership chaining behavior where tables and 
views owned by the view owner can be accessed through the view.

Note: While infrastructure for function ownership chaining is included, 
the actual implementation of function ownership chaining is out of scope 
for this commit.
 
### Issues Resolved

BABEL-6026
 
### Check List

### DIFF
--- a/src/backend/rewrite/rewriteHandler.c
+++ b/src/backend/rewrite/rewriteHandler.c
@@ -48,6 +48,7 @@
 
 bbfViewHasInsteadofTrigger_hook_type bbfViewHasInsteadofTrigger_hook = NULL; /** BBF Hook to check Instead Of trigger on View */
 pre_QueryRewrite_hook_type pre_QueryRewrite_hook = NULL;
+walk_view_rule_hook_type walk_view_rule_hook = NULL;
 
 /* We use a list of these to detect recursion in RewriteQuery */
 typedef struct rewrite_event
@@ -1812,6 +1813,9 @@ ApplyRetrieveRule(Query *parsetree,
 	rule_action = copyObject(linitial(rule->actions));
 
 	AcquireRewriteLocks(rule_action, true, (rc != NULL));
+
+	if (walk_view_rule_hook)
+		walk_view_rule_hook(rule_action, relation->rd_rel->relowner);
 
 	/*
 	 * If FOR [KEY] UPDATE/SHARE of view, mark all the contained tables as
--- a/src/backend/rewrite/rewriteHandler.c
+++ b/src/backend/rewrite/rewriteHandler.c
@@ -49,6 +49,7 @@
 bbfViewHasInsteadofTrigger_hook_type bbfViewHasInsteadofTrigger_hook = NULL; /** BBF Hook to check Instead Of trigger on View */
 pre_QueryRewrite_hook_type pre_QueryRewrite_hook = NULL;
 walk_view_rule_hook_type walk_view_rule_hook = NULL;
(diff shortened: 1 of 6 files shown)


src/include/nodes/parsenodes.h
```diff
@@ -1297,6 +1297,8 @@ typedef struct RTEPermissionInfo
 	Bitmapset  *selectedCols;	/* columns needing SELECT permission */
 	Bitmapset  *insertedCols;	/* columns needing INSERT permission */
 	Bitmapset  *updatedCols;	/* columns needing UPDATE permission */
+	/* For Babelfish Ownership Chaining support */
+	int			insideView pg_node_attr(equal_ignore, query_jumble_ignore, read_write_ignore, read_as(0));
```
tanscorpio7, line 1301: I am curious why we cannot just set the check as user to table owner when ownership chaininig is formed.

src/include/nodes/primnodes.h
```diff
@@ -768,6 +768,9 @@ typedef struct FuncExpr
 	List	   *args;
 	/* token location, or -1 if unknown */
 	ParseLoc	location;
+	/* For Babelfish Ownership Chaining support */
+	Oid			parentOwnerId pg_node_attr(equal_ignore, query_jumble_ignore, read_write_ignore, read_as(0));
+	int			insideView pg_node_attr(equal_ignore, query_jumble_ignore, read_write_ignore, read_as(0));
```
tanscorpio7, line 773: Any specific reason for including these fields in this commit ? I think bringing it along with function/procedures ownership chaining support might be preferrable for git history.

tanscorpio7 (changes requested):

src/include/nodes/parsenodes.h
HarshLunagariya: This is for subsequent work where function/procedure ownership chaining will be implemented.

src/backend/rewrite/rewriteHandler.c
```diff
@@ -3557,6 +3562,8 @@ rewriteTargetView(Query *parsetree, Relation view)
 	new_rte->securityQuals = view_rte->securityQuals;
 	view_rte->securityQuals = NIL;
 
+	if (handle_target_view_hook)
+		handle_target_view_hook(new_perminfo, view_rte);
```
tanscorpio7: We can pass both relation and view's owner as argument here.

tanscorpio7 (changes requested):

tanscorpio7 (approved):

Deepesh125 (approved):

PR #615, opened 2025-08-25
added new grammer, parse node and hook for top N Percent

herambh-shah:
### Description
This PR implements comprehensive support for SELECT TOP N PERCENT syntax in Babelfish, addressing JIRA-BABEL-1358. The implementation adds percentage-based row limiting functionality to align with SQL Server behavior.

### Problem
Babelfish previously only supported SELECT TOP N (fixed count) but lacked support for SELECT TOP N PERCENT (percentage-based limiting), which is a commonly used SQL Server feature for selecting a percentage of rows from result sets.

### Related PR -
Extension Link - https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4041

### Solution

**Core Changes**

- [ ] **1. Grammar Extensions ( gram-tsql-rule.y)**
- Updated Grammer to support Percent Operator
- Also changed return type of tsql_top_clause node to selectLimit to match its behavior with postgres.

- [ ] **2. Parser Updates ( gram-tsql-epilogue.y.c)**
- Added a new field in LimitOption enum **"LIMIT_OPTION_PERCENT"** to represent Percent Op in TSQL.

- [ ] **3. Hook Implementation ( hooks.c)**
- Added error handling for unsupported UPDATE/DELETE/INSERT TOP N PERCENT
- Handled top N Percent by rewriting N -> select count(*) from parent_query
- handled group by by rewriting N -> select count(*) from (select groupby_cols from parent_query group by groupby_cols )

### Test Coverage
1. 43+ test cases covering all scenarios
2. Basic percentage operations (50%, 25%, 0%, 100%)
3. Edge cases (0.5%, 33.33% decimal precision)
4. Complex queries (JOINs, CTEs, window functions, subqueries)
5. FOR JSON integration 
6. Dependent Objects Test Case
7. UNION operations
8. Cross-schema functionality 
9. Error cases for UPDATE/DELETE TOP N PERCENT
10. Update/Delete/insert + Output clause
11.  Complex Update/Delete/insert top 100 percent


src/include/nodes/parsenodes.h
```diff
@@ -2175,8 +2178,23 @@ typedef struct SelectStmt
...
+ * This node is used in gram-tsql-rule.y to store limitCount and Percent op information.
+ * Furthuremore, through this node, we are updating limitCount and isPercent( Added for supporing tsql Percent Op) fields
+ * in SelectStmt, UpdateStmt, DeleteStmt
+ */
+typedef struct TopClause
+{
+    NodeTag   type;           /* type tag */
+    Node     *limitCount;     /* expression for the limit count/percent */
+    bool      isPercent;      /* true if PERCENT specified */
+} TopClause;
```
tanscorpio7: Instead of adding a new structure can we use the exisitng `limitCount` & `LimitOption` field in `SelectStmt` and `Query`. 
We can add this field to DeleteStmt and others if needed.

tanscorpio7 (changes requested):

src/include/nodes/parsenodes.h
```diff
@@ -2051,6 +2051,7 @@ typedef struct InsertStmt
 	OverridingKind override;	/* OVERRIDING clause */
 	Node	   *execStmt; 		/* for INSERT ... EXECUTE */
 	Node       *limitCount;		/* used by INSERT TOP in T-SQL*/
+	bool        isPercent pg_node_attr(equal_ignore); /* Used with UPDATE TOP N PERCENT in T-SQL*/
```
forestkeeper: I think we don't support DML in this release, right ?

src/include/nodes/parsenodes.h
herambh-shah: @tanscorpio7 using limitOption sounds like a good idea but it seems like we have to change a lot of engine and grammer code if we wanted to use limitOption instead of new Struct.
Also, it looks like LimitOption has been used at a lot of places. Hence changing it might have negative implication on simple top/limit queries. 
When I will get time, I will surely try this POC and if this seems feasible then will follow this approach for **update/delete/insert** and update it for **select** also.  Thanks @tanscorpio7 for this suggestion.

src/include/nodes/parsenodes.h
herambh-shah: Yes @forestkeeper, we are currently not supporting other DML operations like Update/Insert/Delete.

However, I had to add isPercent in all DML structs because when we have these DML operations (Update/Delete/Insert) with a TOP clause, they use the same grammar for the TOP clause. Since the PERCENT option is defined inside the TOP clause grammar, I had to add isPercent in all DML structs to avoid errors when running Update/Delete/Insert operations with only the TOP clause.

src/include/nodes/parsenodes.h
tanscorpio7: Adding new fields and structure in engine should always be the last option. Even if we are taking that approach, changes should be as minimal as possible.
For example in current PR we will need to prove `SelectStmt` structure is not persisted anywhere in the database otherwise that is a problem because persisted structures can only be changed/added in new major version.

src/include/nodes/parsenodes.h
forestkeeper: Did you mean the DML share part of the bison parser together with select stmt ?

src/include/nodes/parsenodes.h
forestkeeper: @tanscorpio7 we can't avoid adding new field / structure for this approach. From my understand changing LimitOptions will cause more potential issue compared to adding this TopClause structure.
The minor version upgrade concern about the view, we should add the scheduler test cases into latest schedule file. Pls remind me if I miss something about this.

tanscorpio7: Yes, I understand adding new structure might be unavoidable. Just wondering what the issue could be with extending `LimitOptions`

> The minor version upgrade concern about the view, we should add the scheduler test cases into latest schedule file. Pls remind me if I miss something about this.

We will need to verify it for postgres as well.

src/include/nodes/nodes.h
```diff
@@ -429,6 +429,7 @@ typedef enum LimitOption
 {
 	LIMIT_OPTION_COUNT,			/* FETCH FIRST... ONLY */
 	LIMIT_OPTION_WITH_TIES,		/* FETCH FIRST... WITH TIES */
+	LIMIT_OPTION_PERCENT,		/* FETCH FIRST... PERCENT */
```
tanscorpio7: /* FETCH FIRST... PERCENT FOR BABELFISH */

Small indication that this is babelfish specific only value.

src/backend/parser/analyze.c
```diff
@@ -1441,6 +1444,12 @@ transformSelectStmt(ParseState *pstate, SelectStmt *stmt)
 
 	qry->commandType = CMD_SELECT;
 
+	/* Transform query if percent Operator is present with Top Clause */
+	if(transform_percent_clause_hook)
+	{
+		(*transform_percent_clause_hook)(pstate, stmt);
+	}
+
```
tanscorpio7: We can use the existing `transform_unpivot_clause_hook` instead of adding another hook. We can also rename it to something more generic for example pre_transform_select_stmt_hook.

tanscorpio7 (changes requested):

src/backend/parser/analyze.c
herambh-shah: Changed.

forestkeeper (approved):
LGTM

PR #626, opened 2025-09-26
Fix: Temp table indexes with name starting without '#' do not follow rollback semantics

ayushdsh:
We use the is_bbf_temp_table flag in ENR metadata for transaction behaviour. Currently, we only mark this flag as true if the relname starts with '#', however all indexes on ENR temp tables should be have this flag set.
We are refactoring the register_ENR function and marking this flag early rather than in this method

Extension PR: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4126

### Issues Resolved

Task: BABEL-6096, BABEL-5982
 
### Check List

### DIFF
--- a/src/backend/catalog/index.c
+++ b/src/backend/catalog/index.c
@@ -1013,6 +1013,7 @@ index_create(Relation heapRelation,
 		enr->md.name = palloc0(strlen(indexRelationName) + 1);
 		strncpy(enr->md.name, indexRelationName, strlen(indexRelationName) + 1);
 		enr->md.reliddesc = indexRelationId;
+		enr->md.enrtype = ENR_TSQL_TEMP;
 		enr->md.parent_oid = heapRelationId;
 		register_ENR(currentQueryEnv, enr);
 		MemoryContextSwitchTo(oldcontext);
--- a/src/backend/catalog/index.c
+++ b/src/backend/catalog/index.c
@@ -1013,16 +1013,7 @@ index_create(Relation heapRelation,
 		enr->md.name = palloc0(strlen(indexRelationName) + 1);
 		strncpy(enr->md.name, indexRelationName, strlen(indexRelationName) + 1);
 		enr->md.reliddesc = indexRelationId;
-		enr->md.enrtype = ENR_TSQL_TEMP;
-		/*
-		 * table variables don't have any transaction semantics
-		 * due to their scope. Hence, we would mark the flag only
-		 * if the index is not on table variable
-		 */
-		if (indexRelationName[0] != '@')
-			enr->md.is_bbf_temp_table = true;
-		else
-			enr->md.is_bbf_temp_table = false;
+		enr->md.parent_oid = heapRelationId;
 		register_ENR(currentQueryEnv, enr);
 		MemoryContextSwitchTo(oldcontext);
 	}
--- a/src/backend/catalog/index.c
+++ b/src/backend/catalog/index.c
@@ -1014,6 +1014,15 @@ index_create(Relation heapRelation,
 		strncpy(enr->md.name, indexRelationName, strlen(indexRelationName) + 1);
 		enr->md.reliddesc = indexRelationId;
 		enr->md.enrtype = ENR_TSQL_TEMP;
+		/*
+		 * table variables don't have any transaction semantics
+		 * due to their scope. Hence, we would mark the flag only
+		 * if the index is not on table variable
+		 */
+		if (indexRelationName[0] != '@')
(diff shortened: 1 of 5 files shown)


src/backend/executor/spi.c
```diff
@@ -3365,7 +3365,15 @@ SPI_register_relation(EphemeralNamedRelation enr)
 			else
 				_SPI_current->queryEnv = create_queryEnv();
 		}
-
+		/*
+		 * Aurora Babelfish specific - 
+		 * We've introduced a flag in ENR metadata structure called is_bbf_temp_table
+		 * This flag helps Babelfish temp tables implement transaction semantics
+		 */
```
ayushdsh: This is added defensively. I don't think tsql temp tables will ever be created from this function.

tanscorpio7: I would prefer to set this to false always if it is not expected.

src/backend/catalog/index.c
```diff
@@ -1014,6 +1014,15 @@ index_create(Relation heapRelation,
...
 		enr->md.enrtype = ENR_TSQL_TEMP;
+		/*
+		 * table variables don't have any transaction semantics
+		 * due to their scope. Hence, we would mark the flag only
+		 * if the index is not on table variable
+		 */
+		if (indexRelationName[0] != '@')
+			enr->md.is_bbf_temp_table = true;
+		else
+			enr->md.is_bbf_temp_table = false;
```
tanscorpio7: I think we should keep this check inside `register_ENR` like it is today. Indexes can be handled specially inside `register_ENR`. We can just set the field `enr->md.is_bbf_temp_table` based on the parent table's value of `enr->md.is_bbf_temp_table`.

tanscorpio7 (changes requested):

src/backend/catalog/index.c
lejaokri: +1

ayushdsh: If we want to keep this check in register_ENR there are two ways - 
1. I don't think within register_ENR we have the knowledge about this being an index or relation. We would need to iterate over all the ENR's and within that the index catalog to identify (if this is an index) of a relation.
2. we change the signature of register_ENR, which might be a little risky considering this is a PG API

tanscorpio7: Third option can be to add a field in `EphemeralNamedRelationMetadataData`.

ayushdsh: Yes. We'll have to keep the type of object (enum) in the metadata structure, but even then we cannot avoid this.
> We would need to iterate over all the ENR's and within that the index catalog to identify (if this is an index) of a relation.

which we don't require before calling register_ENR because at that point of time, we do have index's relation with us.

src/backend/utils/misc/queryenvironment.c
```diff
@@ -212,10 +212,30 @@ register_ENR(QueryEnvironment *queryEnv, EphemeralNamedRelation enr)
...
+	if (enr->md.parent_oid != InvalidOid)
+	{
+		/*
+		 * This is an index on a ENR temp object. We mark the flag by checking 
+		 * the parent relations flag.
+		 * If it is an implicit index over tsql table variable, it gets marked 
+		 * as false, same as parent since table variable don't follow transactional
+		 * semantics.
+		 */
+		EphemeralNamedRelation parent_enr = GetENRTempTableWithOid(enr->md.parent_oid);
```
tanscorpio7, line 224: Does `GetENRTempTableWithOid` iterate over out query env ? If not then this fix will actually become a regression for cases like CREATE TABLE #t ... EXEC proc_which_creats_index_on_#t.

Also Postgres indenation for if requires curly braces & else in new line.

tanscorpio7 (changes requested):

src/backend/utils/misc/queryenvironment.c
ayushdsh: Not currently, but we will also be fixing that. Do you think we should have that fix along with this? 

Today, if a proc creates an index, it is created outside of ENR (register_ENR is never called in that case), so technically, if won't come in this codepath in that case. 

>Today, if a proc creates an index, it is created outside of ENR

This is another problem caused by not iterating over parent queryEnvs.

tanscorpio7: If this scenario already fails, I am fine with handling it seperately.

tanscorpio7 (approved):

lejaokri (approved):

jsudrik (approved):
Approved.

PR #632, opened 2025-10-12
Fix: Rectified dialect check in ENRGetSystableScan function to also check for the conn_type

ayushdsh:
### Description
In ENRGetSystableScan, if the dialect is not TSQL_DIALECT and there is a scan initiated for a catalog (other than Sequence), we return false i.e. the tuple being scanned for does not exist in ENR. But this is not reliable in case this scan is initiated from a PG function since the dialect then would be PG_DIALECT and the scan in ENR would return false, even though the actual tuple being searched for is in ENR (in case it's a T-SQL temp table). This results in a "could not open a relation" error since the tuple also does not exist in pg_catalog.

Extension PR (tests) - https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4174

Other dialect checks and why they don't need to change?
```
static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOperationType op, bool skip_cache_inval, bool in_enr_rollback)
{
        ...
	if (sql_dialect != SQL_DIALECT_TSQL)
		return false;

	catalog_oid = RelationGetRelid(catalog_rel);
```
> This is done during adding dependent objects and tuples into ENR temp tables. Since creation of T-SQL temp table should only happen from the tsql dialect, addition of tuples into ENR md should also happen only in tsql dialect.
 
```
void ENRDropEntry(Oid id)
{
	EphemeralNamedRelation	enr;
	MemoryContext oldcxt;

	if (sql_dialect != SQL_DIALECT_TSQL || !currentQueryEnv)
		return;

```
> Again, operations on ENR should happen in the sql_dialect; hence we don't need a conn check here

And other helpers -
```
bool IsTsqlTableVariable(Relation relation)
{
	return sql_dialect == SQL_DIALECT_TSQL 


bool IsTsqlTempTable(char relpersistence)
{
	return sql_dialect == SQL_DIALECT_TSQL 


bool UseTempOidBuffer()
{
	return sql_dialect == SQL_DIALECT_TSQL 

```

> These functions are also used during creation of temp table and dependent objects.

### Issues Resolved
Task: BABEL-6099
 
### Check List

### DIFF
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -52,8 +52,6 @@
 #include "utils/syscache.h"
 #include "utils/queryenvironment.h"
 #include "utils/rel.h"
-#include "libpq/libpq.h"
-#include "miscadmin.h"
 
 #define NUM_ENR_CATALOGS 11
 
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -51,8 +51,11 @@
 #include "utils/syscache.h"
 #include "utils/queryenvironment.h"
 #include "utils/rel.h"
+#include "libpq/libpq.h"
+#include "miscadmin.h"
 
 #define NUM_ENR_CATALOGS 11
+#define IS_TDS_CONN() (MyProcPort && MyProcPort->is_tds_conn)
 
 pltsql_get_tsql_enr_from_oid_hook_type pltsql_get_tsql_enr_from_oid_hook = NULL;
 
@@ -421,7 +424,7 @@ bool ENRGetSystableScan(Relation rel, Oid indexId, int nkeys, ScanKey key, List 
 
 	Oid reloid = RelationGetRelid(rel);
 
-	if (sql_dialect != SQL_DIALECT_TSQL)
+	if (sql_dialect != SQL_DIALECT_TSQL && !IS_TDS_CONN())
 	{
 		/*
 		* We cannot return false right away when sql_dialect is not TSQL.
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -32,6 +32,7 @@
 #include "catalog/catalog.h"
 #include "catalog/dependency.h"
 #include "catalog/indexing.h"
+#include "catalog/namespace.h"
 #include "catalog/pg_authid.h"
 #include "catalog/pg_collation.h"
 #include "catalog/pg_opclass.h"
@@ -55,7 +56,6 @@
 #include "miscadmin.h"
 
 #define NUM_ENR_CATALOGS 11
-#define IS_TDS_CONN() (MyProcPort && MyProcPort->is_tds_conn)
 
 pltsql_get_tsql_enr_from_oid_hook_type pltsql_get_tsql_enr_from_oid_hook = NULL;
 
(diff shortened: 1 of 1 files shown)


src/backend/utils/misc/queryenvironment.c
```diff
@@ -51,8 +51,11 @@
 #include "utils/syscache.h"
 #include "utils/queryenvironment.h"
 #include "utils/rel.h"
+#include "libpq/libpq.h"
+#include "miscadmin.h"
 
 #define NUM_ENR_CATALOGS 11
+#define IS_TDS_CONN() (MyProcPort && MyProcPort->is_tds_conn)
```
tanscorpio7: We can use the `is_bbf_tds_connection_hook` instead.

tanscorpio7 (changes requested):

src/backend/utils/misc/queryenvironment.c
tanscorpio7: Lets also update this comment
```
/*
		* We cannot return false right away when sql_dialect is not TSQL.
		* ...
```

Also a TODO mentioning that this dialect issue can still happen for PG Endpoint using T-SQL temp tables

src/backend/utils/misc/queryenvironment.c
```diff
@@ -421,7 +424,7 @@ bool ENRGetSystableScan(Relation rel, Oid indexId, int nkeys, ScanKey key, List
 
 	Oid reloid = RelationGetRelid(rel);
 
-	if (sql_dialect != SQL_DIALECT_TSQL)
+	if (sql_dialect != SQL_DIALECT_TSQL && !IS_TDS_CONN())
 	{
 		/*
 		* We cannot return false right away when sql_dialect is not TSQL.
```
surendravishnoi: is this special code required now?

ayushdsh: Yes. We are now allowing pg dialect to also look through the ENRs list. 
When PG functions try to rebuild relcache, currently with just a dialect check they are not allowed to look at the ENR list. With this we are opening up that scan for such usecases.

tanscorpio7 (approved):

src/backend/utils/misc/queryenvironment.c
surendravishnoi: I meant code from line 441 onwards for sequence etc, that is taken care of with tds conn check?

ayushdsh: Oh i see. Sorry for the misunderstanding. You're right. I don't think we require this separate handling now. Will rectify

tanscorpio7: We still need it for ENR tables from PostgreSQL endpoint, right ?

ayushdsh: Where do we do such scans? I dont think such a situation will arise ... if a backend from pg endpoint tries to scan for tuple in  🤔

lejaokri (commented):
LGTM - do we have any test for this?

ayushdsh:
> LGTM - do we have any test for this?

Yes. Extension PR - https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4174

src/backend/utils/misc/queryenvironment.c
tanscorpio7: We can execute t-sql procedure from PostgreSQL endpoint and I do not see any TDS connection checks in code whcih decides wheter temp table should go into ENR. We can validate this scenaro.

surendravishnoi: Temp tables from PG endpoint is not well defined behavior. We do not know what works and what does not, I will not worry about it

tanscorpio7: Yes, but wouldn't it be preferrable to keep the PG endpoint code as it is today. Changing it could create some issues.

ayushdsh: Maybe since this condition does not lead to unexpected behaviour we can keep it as it is 🤔

ayushdsh: > We can execute t-sql procedure from PostgreSQL endpoint and I do not see any TDS connection checks in code whcih decides wheter temp table should go into ENR. We can validate this scenaro.

Yes, and it will lead to the same behaviour even if this condition stays in place or not. But yes, you're right. If we are able to create temp tables with sequence today from psql, we won't if we remove this condition

```
1> create proc p as begin 
2> select a into #t from master_dbo.t;
3> select replace(a, '5', '4') from #t;
4> end
5> go
1> quit
```
-- psql


Screenshot 2025-10-14 at 11 28 00 AM

```                                                                                                                                                                     
babelfish_db=# call master_dbo.p();
ERROR:  block 0 of base/16385/t16_20098 is still referenced (local 2)
```

tanscorpio7: Could you remove the SELECT replace ... line and try the procedure call with and without this block

ayushdsh: It did not give me a difference. However, once that condition is removed, there's an infinite recursion because now each `systable_beginscan` calls `ENRGetSystableScan` and within that condition calls `get_func_language_oids_hook` which inturn calls systable_beginscan for the `pg_language` catalog. This is not a problem and can be simply rectified by moving the `IsCatalogENR` check to the top. 

Screenshot 2025-10-14 at 12 39 08 PM

However, I still think its better to keep this condition in because it does not hurt and there's no unexpected behaviour;

What do you think @surendravishnoi ?

surendravishnoi (approved):

PR #665, opened 2025-12-18
Fix: get_ENR_withoid should also search through the parent queryEnvs to find if the ENR with given OID exists.

ayushdsh:
### Description

* `get_ENR_withoid` previously only checked for the given OID within the provided queryEnv; however this function should be identical to `getENR` (as also suggested by the comment), which searches through parent envs as well. Hence, changing `get_ENR_withoid` to also search through parent queryEnvs
    * ~~`getENR` has a search flag which callers can use if they don't want to search through all the queryEnvs; `get_ENR_withoid` does not require this - Unlike getENR, getENR_withoid searches using OIDs and OIDs for every registered ENR will be different.~~
    * We have introduced a similar recurse flag which will be turned off for most cases, except the necessary ones. 

**NOTE: index and toast creation is still not correct as these relations get registered into the incorrect queryEnv. I've created a separate issue for it - BABEL-6272 and we will solve it next**

Extension PR: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4334
 
### Issues Resolved

BABEL-5605, BABEL-6192
 
### Check List

### DIFF
--- a/src/backend/catalog/aclchk.c
+++ b/src/backend/catalog/aclchk.c
@@ -1856,7 +1856,7 @@ ExecGrant_Relation(InternalGrant *istmt)
 		Oid			ownerId;
 		HeapTuple	tuple;
 		ListCell   *cell_colprivs;
-		bool		is_enr = (sql_dialect == SQL_DIALECT_TSQL && get_ENR_withoid(currentQueryEnv, relOid, ENR_TSQL_TEMP));
+		bool		is_enr = (sql_dialect == SQL_DIALECT_TSQL && get_ENR_withoid(currentQueryEnv, relOid, ENR_TSQL_TEMP, false));
 
 		if (is_enr)
 			tuple = SearchSysCache1(RELOID, ObjectIdGetDatum(relOid));
@@ -2206,7 +2206,7 @@ ExecGrant_common(InternalGrant *istmt, Oid classid, AclMode default_privs,
 		int			nnewmembers;
 		Oid		   *oldmembers;
 		Oid		   *newmembers;
-		bool		is_enr = (sql_dialect == SQL_DIALECT_TSQL && get_ENR_withoid(currentQueryEnv, objectid, ENR_TSQL_TEMP));
+		bool		is_enr = (sql_dialect == SQL_DIALECT_TSQL && get_ENR_withoid(currentQueryEnv, objectid, ENR_TSQL_TEMP, false));
 
 
 		if (is_enr)
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -319,16 +319,21 @@ EphemeralNamedRelation
 get_ENR_withoid(QueryEnvironment *queryEnv, Oid id, EphemeralNameRelationType type)
 {
 	ListCell   *lc;
+	QueryEnvironment *qe = queryEnv;
 
 	if (queryEnv == NULL || !OidIsValid(id))
 		return NULL;
 
-	foreach(lc, queryEnv->namedRelList)
+	while (qe)
 	{
-		EphemeralNamedRelation enr = (EphemeralNamedRelation) lfirst(lc);
+		foreach(lc, qe->namedRelList)
+		{
+			EphemeralNamedRelation enr = (EphemeralNamedRelation) lfirst(lc);
 
-		if (enr->md.reliddesc == id && enr->md.enrtype == type)
-			return enr;
+			if (enr->md.reliddesc == id && enr->md.enrtype == type)
+				return enr;
+		}
+		qe = qe->parentEnv;
 	}
 
 	return NULL;
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -222,7 +222,7 @@ register_ENR(QueryEnvironment *queryEnv, EphemeralNamedRelation enr)
 		 * as false, same as parent since table variable don't follow transactional
 		 * semantics.
 		 */
-		EphemeralNamedRelation parent_enr = GetENRTempTableWithOid(enr->md.parent_oid, false);
+		EphemeralNamedRelation parent_enr = GetENRTempTableWithOid(enr->md.parent_oid, true);
 		if(parent_enr)
 			enr->md.is_bbf_temp_table = parent_enr->md.is_bbf_temp_table;
 		else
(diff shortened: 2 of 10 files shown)


src/backend/utils/misc/queryenvironment.c
```diff
@@ -319,16 +319,21 @@ EphemeralNamedRelation
 get_ENR_withoid(QueryEnvironment *queryEnv, Oid id, EphemeralNameRelationType type)
```
tanscorpio7: Please validate that none of the get_ENR_withoid() calls try to search only the current query env. get_ENR() has a argument search for controlling this.

tanscorpio7 (changes requested):

src/backend/utils/misc/queryenvironment.c
ayushdsh: Correct. 
> getENR has a search flag which callers can use if they don't want to search through all the queryEnvs; get_ENR_withoid does not require this - Unlike getENR, getENR_withoid searches using OIDs and OIDs for every registered ENR will be different.

I will still check the callers of `get_ENR_withoid ` to make sure

ayushdsh: I think we should introduce this change slowly. It can't go wrong because OIDs don't overlap, but there are cases where we use get_ENR_withoid() while the caller iterates over the parentEnvs .. and if we do anything with the queryEnv in which we are trying to find the ENR, this will go horribly wrong. Fortunately as far as I see, there are no such operations on queryEnvs directly. However, I think we can start small and solve the problem at hand i.e. BABEL-5605, BABEL-6192. Hence, I am thinking I can do this - 

```
get_ENR_withoid(... bool recurse) 
```
and keep recurse false for most of the cases except for fixing the above 2 JIRAs.

@tanscorpio7 what are your thoughts on this?

tanscorpio7: I am always inclined towards limited and incremental changes : )
> and keep recurse false for most of the cases except for fixing the above 2 JIRAs

I would modify this statement to "and keep recurse false for most of the cases except the ones where recursion is absolutely needed"

For example BABEL-5605 is truncate repro but the exact same issue could also be reproduced with other statements (other DDLs that try to execute SearchSysCacheLocked on temp oid)

ayushdsh: Yes. That's correct!

src/backend/catalog/aclchk.c
```diff
@@ -1856,7 +1856,7 @@ ExecGrant_Relation(InternalGrant *istmt)
 		Oid			ownerId;
 		HeapTuple	tuple;
 		ListCell   *cell_colprivs;
-		bool		is_enr = (sql_dialect == SQL_DIALECT_TSQL && get_ENR_withoid(currentQueryEnv, relOid, ENR_TSQL_TEMP));
+		bool		is_enr = (sql_dialect == SQL_DIALECT_TSQL && get_ENR_withoid(currentQueryEnv, relOid, ENR_TSQL_TEMP, false));
```
tanscorpio7, line 1859: Can we test
```
CREATE TEMP TABLE (col1 INT); (do not create toast table for temporary table yet)
EXEC some_operation_on_temp_table
```

Where operation is
1. Grant on temporary table
2. Grant on column of temporary table
3. Drop temporary table
4. Create index on temporary table
5. SP_RENAME on temporary table name
6. ALTER TABLE ADD COLUMN <column_name_greater_than_64> VARCHAR(100); (too also test add toast)

tanscorpio7 (changes requested):

src/backend/catalog/aclchk.c
ayushdsh: I agree that index creation and toast creation won't work for this. But should we fix this within this same PR or should we have a separate PR for fixing these?

tanscorpio7: I'd prefer this PR because it has the exact same RCA as BABEL-5605

ayushdsh: Ack. That makes sense.

surendravishnoi, line 1859: Do we have cases where we do not want to search for get_ENR_withoid recursively? Similarly, what are cases for get_ENR?

ayushdsh: Correct, so the usage of `get_ENR` is at places where we want to find if an ENR with the same exists in the queryEnv. That's where we want to not go recursive. However, `get_ENR_withoid` usage is at places where we want to find an already existing ENRs. There are cases where we DO NOT want to traverse recursively for `get_ENR_withoid` for ex: in `ENR_tuple_operation` where we add a tuple to ENR catalog, this function itself traverses through the queryEnvs, so it does not make sense for `get_ENR_withoid` to also do this. 

So essentially, in cases where the caller itself traverses through the queryEnv is where we do not want `get_ENR_withoid` to also do it, because its possible that the caller depends on the queryEnv where the ENR is found.

ayushdsh: And for others - 
1. grants/revokes (permission model) do not work with local temp tables 
2. renaming local temp tables is not supported
Ref: https://tinyurl.com/5n7jnxkd

surendravishnoi: I am still not clear. get_ENR finds if an ENR with same name exists or not but still there are two use cases 1) ENR with same name at same level or 2) ENR with same at all levels. How do we decide which use case is which?
get_ENR_withoid explanation is more of a coding issue but in terms of use case is there case we want to search current level vs all levels?

ayushdsh: 1. About deciding whether to traverse through levels is very usecase specific. For example, during `heap_create_with_catalog`, we go through only the current level when trying to find ENR with a name. The usecase for this is to check duplicates; all related functions like `TypeCreate` or `RelationBuildDesc` also only go through current level vs. when calling `object_id` for a given name, we want to go through all the levels; similarly when we are trying to check if a stmt is OK on the read replica, we want to go through the existing ENRs at all levels to see if we are doing the operation on an existing ENR rel.
2. `get_ENR_withoid` usecases are a bit different - based on the usage of the function, I found that its mainly purpose is to find an existing ENR given an OID, and so there are limited or NO usecases where we DO NOT want to search through all the levels. Another way to think about this (which is very wrong) is that since we are searching for OIDs here which cannot overlap, it is also OK to go through all levels.

tanscorpio7 (approved):

PR #692, opened 2026-01-09
Fix: Rectified condition which led to constraint catalog corruption in ENR for TSQL temp tables

ayushdsh:
### Description

PG18 introduced a new constraint which is created for all non-null attributes including p_keys. So now, an attribute which is a p_key will create two constraints - p_key and not-null constraint.

When adding these constraints on the catalog in ENR, the condition tf2.oid > tf1.oid will keep the catalog list sorted, however, while removing or updating any tuple in this catalog list, we compare oids in a way which is interpreted as remove/update oid of tuple which has an oid greater than or equal to the requested oid. This leads to catalog corruption since we are removing/updating a different tuple than the requested one.

This change rectifies this condition and breaks it down based on the operation being performed.

More details can be found in the internal document which lists down the detail of the issue and the fix.

### Issues Resolved

BABEL-6211
 
### Check List

### DIFF
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -1012,10 +1012,18 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
 					tf1 = (Form_pg_constraint) GETSTRUCT(tup);
 					foreach(curlc, enr->md.cattups[ENR_CATTUP_CONSTRAINT]) {
 						tf2 = (Form_pg_constraint) GETSTRUCT((HeapTuple) lfirst(curlc));
-						if (tf2->oid >= tf1->oid) {
-							lc = curlc;
-							insert_at = foreach_current_index(curlc) + 1;
-							break;
+						switch (op)
+						{
+							case ENR_OP_ADD:
+								insert_at = foreach_current_index(curlc) + 1; // will simply be the end of list after iteration
+								break;
+							case ENR_OP_DROP:
+							case ENR_OP_UPDATE:
+								if (tf2->oid >= tf1->oid)
+									lc = curlc;
+								break;
+							default:
+								break;
 						}
 					}
 					ret = true;
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -1010,21 +1010,27 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
 
 					list_ptr = &enr->md.cattups[ENR_CATTUP_CONSTRAINT];
 					tf1 = (Form_pg_constraint) GETSTRUCT(tup);
-					foreach(curlc, enr->md.cattups[ENR_CATTUP_CONSTRAINT]) {
-						tf2 = (Form_pg_constraint) GETSTRUCT((HeapTuple) lfirst(curlc));
-						switch (op)
-						{
-							case ENR_OP_ADD:
-								insert_at = foreach_current_index(curlc) + 1; // will simply be the end of list after iteration
-								break;
-							case ENR_OP_DROP:
-							case ENR_OP_UPDATE:
+					
+					switch (op)
+					{
+						case ENR_OP_ADD:
+							/* For ADD, simply append to the end of the list */
+							insert_at = list_length(enr->md.cattups[ENR_CATTUP_CONSTRAINT]);
+							break;
+						case ENR_OP_DROP:
+						case ENR_OP_UPDATE:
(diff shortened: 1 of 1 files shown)


src/backend/utils/misc/queryenvironment.c
```diff
@@ -1012,10 +1012,18 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
...
 					foreach(curlc, enr->md.cattups[ENR_CATTUP_CONSTRAINT]) {
 						tf2 = (Form_pg_constraint) GETSTRUCT((HeapTuple) lfirst(curlc));
-						if (tf2->oid >= tf1->oid) {
-							lc = curlc;
-							insert_at = foreach_current_index(curlc) + 1;
-							break;
+						switch (op)
+						{
+							case ENR_OP_ADD:
+								insert_at = foreach_current_index(curlc) + 1; // will simply be the end of list after iteration
```
tanscorpio7: Please validate the behaviour of `index_beginscan` and `systable_beginscan_ordered`. Do they gurantee result set to be ordered ? If yes, then it is risky to remove sorting from here even if index_beginscan and systable_beginscan_ordered is never reached for ENR tuples.

tanscorpio7 (changes requested):

src/backend/utils/misc/queryenvironment.c
tanscorpio7: Actually this is a problem anyway. Different indexes will return different order of tuples where as ENR will always returns the result ordered based on OID. I do not even see an index on pg_constraint which has oid as indexed column.

ayushdsh: > Please validate the behaviour of index_beginscan and systable_beginscan_ordered. Do they gurantee result set to be ordered ?

I need to check `systable_beginscan_ordered`, but index_beginscan will work on a different heap altogether right ..  catalog heaps can be unordered, but index heaps need to be kept in a sorted order.

> I do not even see an index on pg_constraint which has oid as indexed column.

Yes. So the order shouldn't matter here right?

tanscorpio7: Yes, the order for oid column in pg_constraint would not matter. Will review the PR later today.

tanscorpio7 (dismissed):
Change makes sense to me since there is no index defined on pg_constratint.oid column. Please also get @surendravishnoi approval for this.

tanscorpio7 (dismissed):

tanscorpio7:
https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/e2a8b58f39e7d8cf9f209eb20e93cc2860c969dc/src/backend/utils/misc/queryenvironment.c#L653 related code for index scan of constraints, wonder how this works since there is not sorting.

Please file a jira to verify index scan for more than one tuple results in ENR

src/backend/utils/misc/queryenvironment.c
```diff
@@ -1010,13 +1010,27 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
...
+							foreach(curlc, enr->md.cattups[ENR_CATTUP_CONSTRAINT]) {
+								tf2 = (Form_pg_constraint) GETSTRUCT((HeapTuple) lfirst(curlc));
+								if (tf2->oid == tf1->oid)
+								{
+									lc = curlc;
+									break;
+								}
+							}
+							break;
+						default:
```
surendravishnoi, line 1032: what does default mean here?

ayushdsh: unreachable pretty much. if at all op is none of the above, the next execution also does nothing. ideally we should have an assert at the very start of the function .. i'll add the in.

ayushdsh:
> https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/e2a8b58f39e7d8cf9f209eb20e93cc2860c969dc/src/backend/utils/misc/queryenvironment.c#L653 related code for index scan of constraints, wonder how this works since there is not sorting.
> 
> 
> 
> Please file a jira to verify index scan for more than one tuple results in ENR

Thanks for pointing this out Tanzeel. let me check what this would mean. just by the looks of it. it shouldn't really matter because we still do oid checks by iterating completely; but let me check this out more.

surendravishnoi (approved):

ayushdsh:
> https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/blob/e2a8b58f39e7d8cf9f209eb20e93cc2860c969dc/src/backend/utils/misc/queryenvironment.c#L653
> 
> related code for index scan of constraints, wonder how this works since there is not sorting.
> Please file a jira to verify index scan for more than one tuple results in ENR

JIRA for tracking analysis - BABEL-6295

PR #698, opened 2026-01-20
[BABEL] Fix circular memory context in nested transactions with SPI cleanup

anju15bharti:
### Description
Fix circular memory context in nested transactions with SPI cleanup

Upstream commit 1afe31f03c changed transaction commit to restore the original memory context instead of switching to TopMemoryContext, fixing session-lifespan memory leaks. However, this exposed an issue in Babelfish
where SPI_finish() destroys temporary "SPI Proc" contexts that may still be referenced as priorContext in active transactions.

The problem manifests when executing bulk insert within a transaction with savepoints. During rollback, the subtransaction's priorContext points to the deleted SPI context. When this deleted memory is reused, it can create a circular reference where curTransactionContext->priorContext->parent points back to curTransactionContext, violating PostgreSQL's assertion that "context != CurrentMemoryContext".

This commit uses parent's transaction context instead of priorContext during subtransaction cleanup for TDS connections to avoid circular dependencies.

Extension PR https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4447 

### Issues Resolved

BABEL-6374, BABEL-6286
 
### Check List

### DIFF
--- a/src/backend/access/transam/xact.c
+++ b/src/backend/access/transam/xact.c
@@ -41,6 +41,7 @@
 #include "common/pg_prng.h"
 #include "executor/spi.h"
 #include "libpq/be-fsstubs.h"
+#include "libpq/libpq.h"
 #include "libpq/pqsignal.h"
 #include "miscadmin.h"
 #include "pg_trace.h"
@@ -2032,13 +2033,26 @@ AtSubCleanup_Memory(void)
 
 	Assert(s->parent != NULL);
 
-	/*
-	 * Return to the memory context that was current before we started the
-	 * subtransaction.  (In principle, this could not be any of the contexts
-	 * we are about to delete.  If it somehow is, assertions in mcxt.c will
-	 * complain.)
-	 */
-	MemoryContextSwitchTo(s->priorContext);
+	if (MyProcPort && MyProcPort->is_tds_conn && sql_dialect == SQL_DIALECT_TSQL)
+	{
+		/*
+		* For TDS/T-SQL connections, use parent's transaction context instead of
+		* priorContext. This avoids using stale priorContext pointers that may
+		* reference deleted SPI memory contexts, which can cause assertion failures
+		* when the memory is reused and creates circular parent-child relationships.
+		*/
+		MemoryContextSwitchTo(s->parent->curTransactionContext);
+	}
+	else
+	{
+		/*
+		* Return to the memory context that was current before we started the
+		* subtransaction.  (In principle, this could not be any of the contexts
+		* we are about to delete.  If it somehow is, assertions in mcxt.c will
+		* complain.)
+		*/
+		MemoryContextSwitchTo(s->priorContext);
+	}
 
 	/* Update CurTransactionContext (might not be same as priorContext) */
 	CurTransactionContext = s->parent->curTransactionContext;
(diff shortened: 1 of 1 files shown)


src/backend/access/transam/xact.c
```diff
@@ -2032,13 +2033,26 @@ AtSubCleanup_Memory(void)
...
 	Assert(s->parent != NULL);
 
-	/*
-	 * Return to the memory context that was current before we started the
-	 * subtransaction.  (In principle, this could not be any of the contexts
-	 * we are about to delete.  If it somehow is, assertions in mcxt.c will
-	 * complain.)
-	 */
-	MemoryContextSwitchTo(s->priorContext);
+	if (MyProcPort && MyProcPort->is_tds_conn && sql_dialect == SQL_DIALECT_TSQL)
```
tanscorpio7: 1. Can we use this hook for checking TDS connection. `if (is_bbf_tds_connection_hook && is_bbf_tds_connection_hook())`
2. Dialect check is not needed. TDS connection check is sufficient.
3. Please update the PR title and description.

tanscorpio7 (changes requested):

src/backend/access/transam/xact.c
tanscorpio7: There are some other paces in xact.c that switch memory context to `priorContext`. Do we also need the changes there ?

tanscorpio7 (changes requested):

src/backend/access/transam/xact.c
anju15bharti: No, There are two other places where we switch to `priorContext`: `AtCleanup_Memory` and `AtCommit_Memory`. However, both functions are only called for top-level transactions where `s->parent` is always NULL. In these cases, priorContext is always valid because it was set before entering any transaction typically in babelfish to the SPI context within which the transaction is running.

tanscorpio7: Isn't that still a problem ? SPI context goes away because its is tied to batch not transaction.

Also the meeting conclusion was (which I think is PG's previos behavior)
1. Switch to parent txn memory context for abort/commit subtransactions.
2. Switch to topmemorycontext for abort/commit transaction.

tanscorpio7 (changes requested):

src/backend/access/transam/xact.c
anju15bharti: We can switch to `TopMemoryContext` for the cleanup/commit transaction paths, but no changes are needed for the abort functions, as they already use the correct context rather than relying on SPI for both cases (transaction and subtransaction).
That said, I intentionally skipped switching to `TopMemoryContext` for the commit/cleanup path for the following reason:
For a top-level transaction, the current memory context remains the SPI context, which is valid in both cases  with and without memory pooling in Babelfish. The issue only occurred when an internal savepoint was created outside of SPI, causing that (non-SPI) context to be used as the CurTransactionContext. Since this doesn't apply to top-level transactions, the SPI context remains valid there.

Switching to TopMemoryContext is arguably safer as a defensive measure, but it isn't strictly required, I don't see a scenario where the current context would be invalid for a top-level transaction.

anju15bharti: As discussed online, we are moving with switching to `TopMemoryContext` even for top-level transaction as it is safe wrt future reference.

tanscorpio7 (approved):

surendravishnoi (approved):

PR #699, opened 2026-01-20
Fix: Rectified removal and updates of incorrect tuples from ENR catalogs

Ayush-061:
### Description

1)Fixed the issue of dropping an intermediate index leads to "could not open relation with oid" error.

--When dropping an index on a temp table that has multiple indexes, the wrong pg_index tuple was being removed from ENR's cattups[ENR_CATTUP_INDEX] because list_head() was used, which always returns the first element regardless of 
which index is being dropped.

2)Fixed the issue -ALTER TABLE on temp table with multiple column defaults corrupts ENR, causing INSERT and DROP TABLE to fail.

-- When altering a column default on a temp table that has multiple columns with DEFAULT values, the wrong pg_attrdef tuple was being removed from ENR's cattups[ENR_CATTUP_ATTR_DEF_REL] because list_head() was used, which always returns the first element regardless of which column's default is being modified.
 
### Issues Resolved

Jiras Fixed - BABEL-6168 and  BABEL-6305

Authored-By- Ayush Pandey  ayushhx@amazon.com
Signed off-By -Ayush Pandey  ayushhx@amazon.com
 
### Check List

### DIFF
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -1081,7 +1081,30 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
 				rel_oid = ((Form_pg_attrdef) GETSTRUCT(tup))->adrelid;
 				if ((enr = get_ENR_withoid(queryEnv, rel_oid, ENR_TSQL_TEMP, false))) {
 					list_ptr = &enr->md.cattups[ENR_CATTUP_ATTR_DEF_REL];
-					lc = list_head(enr->md.cattups[ENR_CATTUP_ATTR_DEF_REL]);
+					if (op ==ENR_OP_DROP || op ==ENR_OP_UPDATE)
+					{ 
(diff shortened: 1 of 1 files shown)


src/backend/utils/misc/queryenvironment.c
```diff
@@ -914,7 +914,27 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
 				rel_oid = ((Form_pg_index) GETSTRUCT(tup))->indrelid;
 				if ((enr = get_ENR_withoid(queryEnv, rel_oid, ENR_TSQL_TEMP, false))) {
 					list_ptr = &enr->md.cattups[ENR_CATTUP_INDEX];
+					if (op ==ENR_OP_DROP || op==ENR_OP_UPDATE){
```
ayushdsh: lets use switch instead of ifs. switch is optimized better in several cases by the compiler (in this particular case, not so much).

ayushdsh: also lets a small description comment on why we are splitting this w.r.t operation.

PR #709, opened 2026-02-15
Fix: Allow creation of dependency from ENR objects to system objects for Temp on RO

ayushdsh:
### Description

Currently, when we create a dependency from ENR object to a non-ENR, we add it to pg_catalog. On RO, we should prevent any ENR -> non-ENR dependency creation, however system objects are safe to create such dependencies because those cannot be deleted. Hence, in case the non-ENR is a system object, we can just skip adding this dependency at all.

Extension PR - https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4577
 
### Issues Resolved

Task: BABEL-6139
 
### Check List

### DIFF
--- a/src/backend/catalog/pg_depend.c
+++ b/src/backend/catalog/pg_depend.c
@@ -108,6 +108,10 @@ recordMultipleDependencies(const ObjectAddress *depender,
 		if (isObjectPinned(referenced))
 			continue;
 		
+		/*
+		 * If the depender object is an ENR and the referenced object is a system object
+		 * we do not need to record this dependency since system objects cannot be altered.
+		 */
 		if (is_enr_to_sys_object_dependency_hook && (*is_enr_to_sys_object_dependency_hook) (depender, referenced))
 			continue;
 
--- a/src/backend/catalog/pg_depend.c
+++ b/src/backend/catalog/pg_depend.c
@@ -107,6 +107,9 @@ recordMultipleDependencies(const ObjectAddress *depender,
 		 */
 		if (isObjectPinned(referenced))
 			continue;
+		
+		if (is_enr_to_sys_object_dependency_hook && (*is_enr_to_sys_object_dependency_hook) (depender, referenced))
+			continue;
 
 		if (slot_init_count < max_slots)
 		{
--- a/src/backend/catalog/pg_depend.c
+++ b/src/backend/catalog/pg_depend.c
@@ -108,10 +108,6 @@ recordMultipleDependencies(const ObjectAddress *depender,
 		if (isObjectPinned(referenced))
 			continue;
 		
-		/*
-		 * If the depender object is an ENR and the referenced object is a system object
-		 * we do not need to record this dependency since system objects cannot be altered.
-		 */
 		if (is_enr_to_sys_object_dependency_hook && (*is_enr_to_sys_object_dependency_hook) (depender, referenced))
 			continue;
 
--- a/src/backend/utils/misc/guc.c
+++ b/src/backend/utils/misc/guc.c
@@ -104,9 +104,6 @@ int			temp_oid_buffer_size;
  */
 bool		temp_table_xact_support;
 
-/* Control if we want to allow ENR <-> non-ENR dependency creation */
-bool pltsql_allow_enr_to_non_enr_dependency;
-
 /*
  * Unit conversion tables.
  *
--- a/src/backend/utils/misc/guc.c
+++ b/src/backend/utils/misc/guc.c
@@ -104,6 +104,9 @@ int			temp_oid_buffer_size;
  */
 bool		temp_table_xact_support;
 
+/* Control if we want to allow ENR <-> non-ENR dependency creation */
+bool pltsql_allow_enr_to_non_enr_dependency;
+
 /*
  * Unit conversion tables.
  *
--- a/src/backend/utils/misc/queryenvironment.c
+++ b/src/backend/utils/misc/queryenvironment.c
@@ -57,6 +57,7 @@
 
 pltsql_get_tsql_enr_from_oid_hook_type pltsql_get_tsql_enr_from_oid_hook = NULL;
 find_object_in_enr_hook_type find_object_in_enr_hook = NULL;
+is_enr_to_sys_object_dependency_hook_type is_enr_to_sys_object_dependency_hook = NULL;
 
 /*
  * This list must match ENRCatalogTupleType in queryenvironment.h.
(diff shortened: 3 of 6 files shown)


src/backend/utils/misc/queryenvironment.c
```diff
@@ -1206,86 +1208,72 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
...
-					}
-				}
+	/*
+	 * pg_depend entry shows relation/type/constraint depends on a given object.
+	 * Find the associated ENR for this pg_depend tuple. If found, make sure
+	 * to register the dependency of the ENR relation to this object.
+	 */
+	switch (entry->classid) {
+		case RelationRelationId:
+			return get_ENR_withoid(queryEnv, entry->objid, ENR_TSQL_TEMP, false);
```
tanscorpio7: 1. Can we use find_object_in_enr_hook for RelationRelationId ?
2. Can we move the `if (find_object_in_enr_hook)` check to the top of the function.
3. Ideally all ENR pg_depend entries should be ENR on both sides. If there are some cases where we are adding a ENR with non ENR pg_depend entry, then the fix should be to verify that its safe and skip it at higher level. So the final version of function should have the same checks as `ConstraintRelationId` for all cases.

src/backend/utils/misc/queryenvironment.c
```diff
@@ -1206,86 +1208,72 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
...
+				return ref_obj_enr;
+			}
+			else if (obj_enr)
+			{
+				
+				/*
+				 * If the dependent object is an ENR and dependency is a non-ENR which is also not a system object,
+				 * we will not allow this creation.
+				 */
+				elog(ERROR, "Cannot create dependency on a non-ENR object");
```
tanscorpio7: Lets use unsupported error code. Also error message is not complete. How about
`temporary objects cannot reference non-system permanent objects`

src/backend/utils/misc/queryenvironment.c
```diff
@@ -1206,86 +1208,72 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
...
+		default:
+			break;
 	}
 	return NULL;
 }
 
-static 
-EphemeralNamedRelation find_pg_depend_tuple(Form_pg_depend tf)
+static EphemeralNamedRelation
+find_pg_depend_tuple(QueryEnvironment *qe, HeapTuple tf, ListCell** lc)
```
tanscorpio7: Could you try flattening this into `_ENR_tuple_operation` ? This function does not do much now but I am not sure which way will be better.

src/backend/utils/misc/queryenvironment.c
tanscorpio7: Lets make it more explicit. `else if (op == ENR_OP_DELETE)`

src/backend/catalog/pg_depend.c
```diff
@@ -107,6 +107,13 @@ recordMultipleDependencies(const ObjectAddress *depender,
 		 */
 		if (isObjectPinned(referenced))
 			continue;
+		
+		/*
+		 * If the depender object is an ENR and the referenced object is a system object
+		 * we do not need to record this dependency since system objects cannot be altered.
+		 */
```
tanscorpio7: Lets move the comments to extension code. Even minor changes can cause merge conflicts with upstream so engine changes should be as minimal as possible.

tanscorpio7 (changes requested):

tanscorpio7:
Thanks for working on this revision. I agree with the high level idea and will review the test cases next week.

src/backend/utils/misc/queryenvironment.c
ayushdsh: agreed .. i think i can try to make it a macro instead .. would be better for readability

src/backend/utils/misc/queryenvironment.c
ayushdsh: agreed with first 2 .. for the 3rd case - are you saying the fn should assume ops from enr->enr and other cases should be dealt at the higher level?

tanscorpio7: Correct, we should only have three cases
1. ENR ↔ ENR: Add dependency in ENR.
2. ENR ↔ safe non ENR object: Skip (pg_depend wouldn't help since other session wouldn't see it for non ENR objects).
3. ENR ↔ non ENR user object: Reject (should never happen)

src/backend/utils/misc/queryenvironment.c
ayushdsh: I was checking on how to do this .. but I am finding it difficult to do it in a way we require the least explanation, flattening it makes it a little harder to understand. I'll add comments above this, and let me think if we can do anything else.

src/backend/utils/misc/guc.c
```diff
@@ -104,6 +104,9 @@ int			temp_oid_buffer_size;
  */
 bool		temp_table_xact_support;
 
+/* Control if we want to allow ENR <-> non-ENR dependency creation */
+bool pltsql_allow_enr_to_non_enr_dependency;
+
```
tanscorpio7: Can we declare this in queryenv.c ?

tanscorpio7 (changes requested):

tanscorpio7 (dismissed):

tanscorpio7:
LGTM! Please get Surendra's approval as well.

src/backend/utils/misc/queryenvironment.c
```diff
@@ -1206,86 +1208,86 @@ static bool _ENR_tuple_operation(Relation catalog_rel, HeapTuple tup, ENRTupleOp
...
-					}
-				}
-				break;
+	if (!depender_object_enr)
+		return NULL;
 
-			case ConstraintRelationId:
-			case AttrDefaultRelationId:
-				return get_ENR_withoid(queryEnv, entry->refobjid, ENR_TSQL_TEMP, false);
+	switch (entry->classid) {
```
surendravishnoi, line 1232: how did we come up with this classification that 
- relation id and type id we only need to check depender
- constrainst and att we check both
- Others we put in non-ENR

ayushdsh: This is still incomplete. For every pg_depend entry, we should be checking both depender and referenced objects. 
The reason why we did not do it yet for rel and type is - we currently have a problem with our transaction handling code (especially rollback). The problem is the incorrect order of adding back objects leading to crashes. More details in the JIRA - BABEL-6368.

I think this is a completely separate problem requiring some careful analysis, hence we did not do this change yet. 

But you're right. there shouldn't be this classification at all. pg_depend tuples should be evaluated on both the sides.

tanscorpio7 (approved):

src/backend/utils/misc/queryenvironment.c
ayushdsh: discussed offline

PR #716, opened 2026-02-20
[OSS-ONLY] Update workflow file to have write access to push tags

shardgupta:
### Description

Update manual_tags workflow file with write access. 
 
signed-off-by: Shard Gupta <shardga@amazon.com>


jsudrik (approved):
Approved.

PR #717, opened 2026-02-23
[BABEL] Fix: Replace pltsql_get_tsql_enr_from_oid_hook check with find_object_in_enr_hook instead

ayushdsh:
### Description

pltsql_get_tsql_enr_from_oid_hook checks if temp_oid_buffer is enabled and only then search for ENRs. However, temp_oid_buffer does not dictate if the relation goes into ENR or not. Intention with CheckRelationLockedByMe is to skip checking if the rel is ENR, hence we will simply replace this check with a simpler one - find_object_in_enr_hook.

ATRewriteTable now assumes that the new relation will be locked by the caller, however, if the rel is an ENR we skip locking it.

Extension PR: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4594

Community commit which makes this assumption - https://github.com/babelfish-for-postgresql/postgresql_modified_for_babelfish/commit/0eb763c691771fcb92af0143cb22219b08be798f
 
### Issues Resolved

BABEL-6284
 
### Check List

### DIFF
--- a/src/backend/storage/lmgr/lmgr.c
+++ b/src/backend/storage/lmgr/lmgr.c
@@ -336,7 +336,7 @@ CheckRelationLockedByMe(Relation relation, LOCKMODE lockmode, bool orstronger)
 {
 	LOCKTAG		tag;
 
-	if (pltsql_get_tsql_enr_from_oid_hook && (*pltsql_get_tsql_enr_from_oid_hook)(RelationGetRelid(relation)))
+	if (find_object_in_enr_hook && (*find_object_in_enr_hook)(RelationRelationId, RelationGetRelid(relation), NULL))
 		return true;
 
 	SET_LOCKTAG_RELATION(tag,
--- a/src/backend/storage/lmgr/lock.c
+++ b/src/backend/storage/lmgr/lock.c
@@ -907,8 +907,7 @@ LockAcquireExtended(const LOCKTAG *locktag,
 	bool		found_conflict;
 	ProcWaitStatus waitResult;
 	bool		log_lock = false;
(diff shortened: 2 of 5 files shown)


tanscorpio7 (changes requested):
1. Please add [BABEL] prefix to PR title.
2. Do you plan to cherry pick this to previous majors ?  I think we should since this is an improvment over the previous logic.
3. Please also add the community commit that caused the original failure.
4. Is it a good time to fix this logic in other places ? i.e. all the remaining places using prev_pltsql_get_tsql_enr_from_oid_hook

ayushdsh:
> 1. Please add [BABEL] prefix to PR title.

just wondering why do we do this btw?

> 2. Do you plan to cherry pick this to previous majors ?  I think we should since this is an improvment over the previous logic.

Yes. We will be cherry-picking this to 5_X as well. Recent temp table fixes are happening >= 5. 

> 4. Is it a good time to fix this logic in other places ? i.e. all the remaining places using prev_pltsql_get_tsql_enr_from_oid_hook


Yes actually. There were a couple of callers for this more. Out of which, I think the inval msg decision is actually logical and should have the buffer_size check. Locking is independent of buffer and should only depend on ENRs or not.

tanscorpio7:
Even shared invalidation is applicable for ENR tables. If other backends cannot see the ENR entry, then they cannot send SI messages for it.

tanscorpio7 (approved):

tanscorpio7:
> just wondering why do we do this btw?

A little bit of commit history cleanliness on gitfarm side.

PR #718, opened 2026-02-24
[BABEL] Fix for Index and Toast Creation Not getting registered in correct QueryEnvironment

Ayush-061:
### Description

Index and Toast creation in Case of ENR temp tables were getting registered in CurrentQueryEnv, while the table for which this toast is being created were in the parent queryEnv.
 
Extension PR- https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4596
### Issues Resolved

BABEL -6272

Authored By- ayushhhx(ayushhx@amazon.com)
Signed off by - ayushhhx(ayushhx@amazon.com)
 
### Check List

### DIFF
--- a/src/backend/catalog/heap.c
+++ b/src/backend/catalog/heap.c
@@ -1402,37 +1402,27 @@ heap_create_with_catalog(const char *relname,
 		enr->md.parent_oid = InvalidOid;
 
 		/*
-		 * For toast tables, register the ENR in the same queryEnv as the parent table
+		 * For toast tables, register the ENR in the same queryEnv as the parent table.
+		 * Toast table name format: #pg_toast_<parent_oid>
 		 */
-		if (relkind == RELKIND_TOASTVALUE)
+		if (relkind == RELKIND_TOASTVALUE && strncmp(relname, "#pg_toast_", 10) == 0)
 		{
-			Oid parent_oid = InvalidOid;
-			const char *oid_start = NULL;
+			Oid parent_oid = (Oid) strtoul(relname + 10, NULL, 10);
 
-			/* Extract parent OID from name pattern */
-			if (strncmp(relname, "#pg_toast_", 10) == 0)
-				oid_start = relname + 10;
-			else if (strncmp(relname, "@pg_toast_", 10) == 0)
-				oid_start = relname + 10;
-
-			if (oid_start)
+			if (OidIsValid(parent_oid))
 			{
-				parent_oid = (Oid) strtoul(oid_start, NULL, 10);
-				if (OidIsValid(parent_oid))
-				{
-					QueryEnvironment *qe = currentQueryEnv;
-					enr->md.parent_oid = parent_oid;
+				QueryEnvironment *qe = currentQueryEnv;
+				enr->md.parent_oid = parent_oid;
 
-					/* Find queryEnv where parent table is registered */
-					while (qe)
+				/* Find queryEnv where parent table is registered */
+				while (qe)
+				{
+					if (get_ENR_withoid(qe, parent_oid, ENR_TSQL_TEMP, false))
 					{
-						if (get_ENR_withoid(qe, parent_oid, ENR_TSQL_TEMP, false))
-						{
-							target_queryEnv = qe;
-							break;
-						}
-						qe = qe->parentEnv;
+						target_queryEnv = qe;
+						break;
 					}
+					qe = qe->parentEnv;
 				}
 			}
 		}
--- a/src/backend/catalog/heap.c
+++ b/src/backend/catalog/heap.c
@@ -1407,15 +1407,14 @@ heap_create_with_catalog(const char *relname,
 		if (relkind == RELKIND_TOASTVALUE && strncmp(relname, "#pg_toast_", 10) == 0)
 		{
 			Oid parent_oid = (Oid) strtoul(relname + 10, NULL, 10);
+			QueryEnvironment *parent_queryEnv;
 
-			if (OidIsValid(parent_oid))
-			{
-				QueryEnvironment *parent_queryEnv = find_ENR_queryEnv(parent_oid);
+			Assert(OidIsValid(parent_oid));
 
-				enr->md.parent_oid = parent_oid;
-				if (parent_queryEnv != NULL)
-					target_queryEnv = parent_queryEnv;
-			}
+			enr->md.parent_oid = parent_oid;
+			parent_queryEnv = find_ENR_queryEnv(parent_oid);
+			if (parent_queryEnv != NULL)
+				target_queryEnv = parent_queryEnv;
 		}
 
 		register_ENR(target_queryEnv, enr);
--- a/src/backend/catalog/heap.c
+++ b/src/backend/catalog/heap.c
@@ -1402,16 +1402,16 @@ heap_create_with_catalog(const char *relname,
 		enr->md.parent_oid = InvalidOid;
 
 		/*
-		 * For toast tables, register the ENR in the same queryEnv as the parent table.
+		 * For toast tables, register the relation as ENR in the same queryEnv as the parent relation table.
 		 */
 		if (relkind == RELKIND_TOASTVALUE)
 		{
-			Oid parent_oid = get_toast_parent_oid(relname);
+			Oid parent_rel_oid = get_toast_parent_oid(relname);
 
-			Assert(OidIsValid(parent_oid));
+			Assert(OidIsValid(parent_rel_oid));
 
-			enr->md.parent_oid = parent_oid;
-			target_queryEnv = find_ENR_queryEnv(parent_oid);
+			enr->md.parent_oid = parent_rel_oid;
+			target_queryEnv = find_ENR_queryEnv(parent_rel_oid);
 			if (!target_queryEnv)
 				target_queryEnv = currentQueryEnv;
 		}
--- a/src/backend/catalog/heap.c
+++ b/src/backend/catalog/heap.c
@@ -1412,8 +1412,7 @@ heap_create_with_catalog(const char *relname,
 
 			enr->md.parent_oid = parent_rel_oid;
 			target_queryEnv = find_ENR_queryEnv(parent_rel_oid);
-			if (!target_queryEnv)
-				target_queryEnv = currentQueryEnv;
+			Assert(target_queryEnv != NULL);
 		}
 
 		register_ENR(target_queryEnv, enr);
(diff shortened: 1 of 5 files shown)


src/backend/catalog/heap.c
```diff
@@ -1392,12 +1393,40 @@ heap_create_with_catalog(const char *relname,
...
+				/* Find queryEnv where parent table is registered */
+				while (qe)
+				{
+					if (get_ENR_withoid(qe, parent_oid, ENR_TSQL_TEMP, false))
+					{
+						target_queryEnv = qe;
+						break;
+					}
+					qe = qe->parentEnv;
+				}
```
ayushdsh: lets simplify this and reuse it everywhere. I remember we have a similar usecase in ENRDropEntry as well.

src/backend/catalog/heap.c
```diff
@@ -1392,12 +1393,40 @@ heap_create_with_catalog(const char *relname,
...
 		strncpy(enr->md.name, relname, strlen(relname) + 1);
 		enr->md.reliddesc = relid;
 		enr->md.enrtype = ENR_TSQL_TEMP;
 		enr->md.parent_oid = InvalidOid;
-		register_ENR(currentQueryEnv, enr);
+
+		/*
+		 * For toast tables, register the ENR in the same queryEnv as the parent table.
+		 */
+		if (relkind == RELKIND_TOASTVALUE && strncmp(relname, "#pg_toast_", 10) == 0)
```
ayushdsh: can you please check if there's already a utility function to do this check for us

src/backend/catalog/index.c
```diff
@@ -1010,12 +1011,32 @@ index_create(Relation heapRelation,
...
 		strncpy(enr->md.name, indexRelationName, strlen(indexRelationName) + 1);
 		enr->md.reliddesc = indexRelationId;
 		enr->md.enrtype = ENR_TSQL_TEMP;
 		enr->md.parent_oid = heapRelationId;
-		register_ENR(currentQueryEnv, enr);
+
+		/*
+		 * Find the query environment where the parent temp table's ENR lives.
+		 */
+		if (currentQueryEnv)
```
ayushdsh: seems like a redundant check right? why would currentQueryEnv not exist?

Ayush-061: yup fixed it

src/backend/catalog/heap.c
Ayush-061: I didnt find any utility function regarding this check

src/backend/catalog/heap.c
```diff
@@ -1392,12 +1393,32 @@ heap_create_with_catalog(const char *relname,
...
+		{
+			Oid parent_oid = (Oid) strtoul(relname + 10, NULL, 10);
```
ayushdsh: this is not the right way to get the parent oid .. we should get the parent rel and from there get the oid .. again, lets see if we have a utility fn to do this

src/backend/catalog/heap.c
```diff
@@ -1392,12 +1393,32 @@ heap_create_with_catalog(const char *relname,
...
+
+			if (OidIsValid(parent_oid))
```
ayushdsh: lets add an assert instead .. a toast table will always have a parent table

src/backend/catalog/index.c
```diff
@@ -1010,12 +1011,25 @@ index_create(Relation heapRelation,
...
 		enr->md.reliddesc = indexRelationId;
 		enr->md.enrtype = ENR_TSQL_TEMP;
 		enr->md.parent_oid = heapRelationId;
-		register_ENR(currentQueryEnv, enr);
+
+		/*
+		 * Register the index ENR in the same queryEnv as its parent table.
+		 * This is important for SP_EXECUTESQL scenarios where the temp table
+		 * is created in an outer scope but the index is created in a nested
+		 * query environment.
```
ayushdsh: > * This is important for SP_EXECUTESQL scenarios where the temp table
 > * is created in an outer scope but the index is created in a nested
 >  * query environment.

No need to add this. Just the first line is enough .. it is obvious why index should also be in the same queryEnv as rel

src/backend/utils/misc/queryenvironment.c
```diff
@@ -1970,3 +1970,25 @@ bool has_existing_enr_relations()
...
+
+/*
+ * find_ENR_queryEnv
+ *
+ * Find the query environment where an ENR with the given OID is registered.
+ * Searches from currentQueryEnv up through parent environments.
+ * Returns the queryEnv where the ENR is found, or NULL if not found.
+ */
+QueryEnvironment *
+find_ENR_queryEnv(Oid relid)
```
ayushdsh: can you also similarly use this in ENRDropEntry as well

src/backend/catalog/index.c
Ayush-061: Rectified the comment

src/backend/catalog/heap.c
ayushdsh: checking relkind is enough here. toast tables cannot be created explicitly.

src/backend/catalog/heap.c
ayushdsh: I don't think there can be a way to get the parentreloid. this would be the only way. can we create a simple inline function for this here .. we should get rid of magic numbers like 10 and #define them as well

src/backend/catalog/heap.c
```diff
@@ -1392,12 +1393,31 @@ heap_create_with_catalog(const char *relname,
...
+		{
+			Oid parent_oid = (Oid) strtoul(relname + 10, NULL, 10);
+			QueryEnvironment *parent_queryEnv;
+
+			Assert(OidIsValid(parent_oid));
+
+			enr->md.parent_oid = parent_oid;
+			parent_queryEnv = find_ENR_queryEnv(parent_oid);
+			if (parent_queryEnv != NULL)
+				target_queryEnv = parent_queryEnv;
```
ayushdsh: lets simplify this to directly - `target_queryEnv = find_ENR_queryEnv()`

src/backend/catalog/index.c
```diff
@@ -1010,12 +1011,22 @@ index_create(Relation heapRelation,
...
 		enr->md.enrtype = ENR_TSQL_TEMP;
 		enr->md.parent_oid = heapRelationId;
-		register_ENR(currentQueryEnv, enr);
+
+		/*
+		 * Register the index ENR in the same queryEnv as its parent table.
+		 */
+		target_queryEnv = find_ENR_queryEnv(heapRelationId);
+		if (target_queryEnv == NULL)
+			target_queryEnv = currentQueryEnv;
```
ayushdsh: lets do this while declaring the var itself .. makes it cleaner

src/backend/utils/misc/queryenvironment.c
```diff
@@ -1297,17 +1297,14 @@ void ENRDropEntry(Oid id)
...
-	queryEnv = currentQueryEnv;
-	while (queryEnv)
-	{
-		if ((enr = get_ENR_withoid(queryEnv, id, ENR_TSQL_TEMP, false)) != NULL)
-			break;
-		queryEnv = queryEnv->parentEnv;
-	}
+	/* Find the query environment containing the ENR */
+	queryEnv = find_ENR_queryEnv(id);
+	if (queryEnv == NULL)
```
ayushdsh: simplified to `if (!queryEnv)` .. similarly below

src/backend/catalog/heap.c
Ayush-061: > checking relkind is enough here. toast tables cannot be created explicitly.

done, removed the check for length

src/backend/utils/misc/queryenvironment.c
Ayush-061: Simplified

ayushdsh:
Please add [BABEL] in the title

src/backend/catalog/heap.c
```diff
@@ -1392,12 +1393,30 @@ heap_create_with_catalog(const char *relname,
...
+
 		enr->md.name = palloc0(strlen(relname) + 1);
 		strncpy(enr->md.name, relname, strlen(relname) + 1);
 		enr->md.reliddesc = relid;
 		enr->md.enrtype = ENR_TSQL_TEMP;
 		enr->md.parent_oid = InvalidOid;
-		register_ENR(currentQueryEnv, enr);
+
+		/*
+		 * For toast tables, register the ENR in the same queryEnv as the parent table.
```
ayushdsh: Nit: `register the relation as ENR in the same queryEnv as the parent relation table`

src/backend/catalog/heap.c
```diff
@@ -1392,12 +1393,30 @@ heap_create_with_catalog(const char *relname,
...
+		 */
+		if (relkind == RELKIND_TOASTVALUE)
+		{
+			Oid parent_oid = get_toast_parent_oid(relname);
```
ayushdsh: Nit: `parent_rel_oid`

src/backend/catalog/index.c
```diff
@@ -1010,12 +1011,19 @@ index_create(Relation heapRelation,
 	{
 		MemoryContext oldcontext = MemoryContextSwitchTo(CacheMemoryContext);
 		EphemeralNamedRelation enr = palloc0(sizeof(EphemeralNamedRelationData));
+
+		/* Register the index ENR in the same queryEnv as its parent table */
+		QueryEnvironment *target_queryEnv = find_ENR_queryEnv(heapRelationId);
+		if (!target_queryEnv)
+			target_queryEnv = currentQueryEnv;
```
ayushdsh: when can it happen that a targetEnv for the heapRel will not be present? If this cannot ever happen, lets instead add an `Assert` instead.

src/include/utils/queryenvironment.h
```diff
@@ -190,4 +191,24 @@ extern PGDLLIMPORT find_object_in_enr_hook_type find_object_in_enr_hook;
...
+ * Toast table names follow the pattern "#pg_toast_<parent_oid>".
+ * Returns the parent OID, or InvalidOid if the name doesn't match the pattern.
+ */
+static inline Oid
+get_toast_parent_oid(const char *relname)
+{
+	if (strncmp(relname, BBF_TEMP_TOAST_PREFIX, BBF_TEMP_TOAST_PREFIX_LEN) == 0)
+		return (Oid) strtoul(relname + BBF_TEMP_TOAST_PREFIX_LEN, NULL, 10);
+	return InvalidOid;
+}
```
ayushdsh, line 214: lets not write definitions in header files .. was there an example that you saw where we've done something like this?

ayushdsh (approved):

ayushdsh (approved):

shardgupta (approved):

PR #725, opened 2026-03-03
[BABEL] Fix to prevent SPI_finish crash on partially-initialized SPI connection after OOM

RuchaSK1:
### Description

SPI_connect_ext() increments _SPI_connected before allocating memory contexts (procCxt, execCxt). If AllocSetContextCreate fails with OOM between the counter increment and context allocation, the SPI stack is left in an inconsistent state as the counter says a connection exists but the memory contexts are NULL.

In vanilla PostgreSQL this is harmless because SPI_finish is never called on error paths — AtEOXact_SPI safely resets the counter during transaction abort. However, Babelfish calls SPI_finish explicitly in error handling and cleanup paths to emulate SQL Server's error handling semantics. This exposes the inconsistent state and causes a crash in MemoryContextDelete(NULL).

SPI_finish crashes on such entries because it unconditionally calls MemoryContextDelete on NULL pointers.

SPI_finish_safe checks whether the connection is fully initialized. If so, it uses SPI_finish. If not, it safely cleans up any partial allocation and pops the stack. 

Extension PR: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4629 

### Issues Resolved

Task: BABEL-6349

Authored-by: Rucha Kulkarni [ruchask@amazon.com](mailto:ruchask@amazon.com)
Signed-off-by: Rucha Kulkarni [ruchask@amazon.com](mailto:ruchask@amazon.com)
 
### Check List

### DIFF
--- a/src/backend/executor/spi.c
+++ b/src/backend/executor/spi.c
@@ -216,6 +216,55 @@ SPI_finish(void)
 }
 
 /*
+ * Modified version of SPI_finish for Babelfish, which safely handles
+ * the case where the current SPI connection was not fully initialized.
+ * For fully initialized connections, delegates to SPI_finish.
+ */
+int
+SPI_finish_safe(void)
+{
+	int			res;
+
+	/* Check if there is an active SPI connection */
+	res = _SPI_begin_call(false);
+	if (res < 0)
+		return res;
+
+	/* Fully initialized */
+	if (_SPI_current->procCxt != NULL &&
+		_SPI_current->execCxt != NULL)
+	{
+		return SPI_finish();
+	}
+
+	/* Clean up any partially allocated context */
+	if (_SPI_current->execCxt)
+	{
+		MemoryContextDelete(_SPI_current->execCxt);
+		_SPI_current->execCxt = NULL;
+	}
+	if (_SPI_current->procCxt)
+	{
+		MemoryContextDelete(_SPI_current->procCxt);
+		_SPI_current->procCxt = NULL;
+	}
+
+	/* Restore outer API variables */
+	SPI_processed = _SPI_current->outer_processed;
+	SPI_tuptable = _SPI_current->outer_tuptable;
+	SPI_result = _SPI_current->outer_result;
+
+	/* Pop the stack */
+	_SPI_connected--;
+	if (_SPI_connected < 0)
+		_SPI_current = NULL;
+	else
+		_SPI_current = &(_SPI_stack[_SPI_connected]);
+
+	return SPI_OK_FINISH;
+}
+
+/*
  * SPI_start_transaction is a no-op, kept for backwards compatibility.
  * SPI callers are *always* inside a transaction.
  */
--- a/src/backend/executor/spi.c
+++ b/src/backend/executor/spi.c
@@ -216,54 +216,6 @@ SPI_finish(void)
 }
 
 /*
- * Modified version of SPI_finish for Babelfish, which safely handles
- * the case where the current SPI connection was not fully initialized.
- * For fully initialized connections, delegates to SPI_finish.
- */
-int
-SPI_finish_safe(void)
-{
-	int			res;
-
-	/* Check if there is an active SPI connection */
-	if (_SPI_current == NULL)
-		return SPI_ERROR_UNCONNECTED;
-
-	/* Fully initialized */
-	if (_SPI_current->procCxt != NULL &&
-		_SPI_current->execCxt != NULL)
-	{
-		return SPI_finish();
-	}
-
-	/* Clean up any partially allocated context */
-	if (_SPI_current->execCxt)
-	{
-		MemoryContextDelete(_SPI_current->execCxt);
-		_SPI_current->execCxt = NULL;
-	}
-	if (_SPI_current->procCxt)
-	{
-		MemoryContextDelete(_SPI_current->procCxt);
-		_SPI_current->procCxt = NULL;
-	}
-
-	/* Restore outer API variables */
-	SPI_processed = _SPI_current->outer_processed;
-	SPI_tuptable = _SPI_current->outer_tuptable;
-	SPI_result = _SPI_current->outer_result;
-
-	/* Pop the stack */
-	_SPI_connected--;
-	if (_SPI_connected < 0)
-		_SPI_current = NULL;
-	else
-		_SPI_current = &(_SPI_stack[_SPI_connected]);
-
-	return SPI_OK_FINISH;
-}
-
-/*
  * SPI_start_transaction is a no-op, kept for backwards compatibility.
  * SPI callers are *always* inside a transaction.
  */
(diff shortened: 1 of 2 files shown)


src/backend/executor/spi.c
```diff
@@ -215,6 +215,55 @@ SPI_finish(void)
...
+ */
+int
+SPI_finish_safe(void)
+{
+	int			res;
+
+	/* Check if there is an active SPI connection */
+	res = _SPI_begin_call(false);
+	if (res < 0)
+		return res;
```
tanscorpio7: 1. Is this safe if the SPI's memory context is NULL ?
2. SPI_finish will also do this, so first line in this function should be calling SPI_finish for safe case.

src/backend/executor/spi.c
```diff
@@ -215,6 +215,55 @@ SPI_finish(void)
...
+	if (_SPI_current->procCxt)
+	{
+		MemoryContextDelete(_SPI_current->procCxt);
+		_SPI_current->procCxt = NULL;
+	}
+
+	/* Restore outer API variables */
+	SPI_processed = _SPI_current->outer_processed;
+	SPI_tuptable = _SPI_current->outer_tuptable;
+	SPI_result = _SPI_current->outer_result;
```
tanscorpio7: Is it possible that error occured before SPI set this variable during SPI_connect ?

src/backend/executor/spi.c
```diff
@@ -215,6 +215,55 @@ SPI_finish(void)
...
+	if (res < 0)
+		return res;
+
+	/* Fully initialized */
+	if (_SPI_current->procCxt != NULL &&
+		_SPI_current->execCxt != NULL)
+	{
+		return SPI_finish();
+	}
+
```
tanscorpio7: If we reach here for internalxact = true, we should throw a FATAL error and terminate the session.
It should be unreachable.

tanscorpio7 (changes requested):

src/backend/executor/spi.c
tanscorpio7: Also please move SPI_finish_safe to the bottom of the file similar to other babelfish specific SPI functions.

tanscorpio7 (changes requested):

src/backend/executor/spi.c
RuchaSK1: 1. Yes. _SPI_begin_call(false) only checks if (_SPI_current == NULL). It never accesses any memory context. So it is safe when memory contexts are NULL.

2. Agreed that SPI_finish should be the first call for the safe case. But to verify it's a safe case, we need to check procCxt and execCxt. And to access those, we need to confirm _SPI_current is not NULL first. The _SPI_begin_call(false) in the current code does exactly this — just if (_SPI_current == NULL). If we want to avoid that call, we still need the same check manually:

```
if (_SPI_current == NULL)
    return SPI_ERROR_UNCONNECTED;
```
I can replace _SPI_begin_call with this inline check to make it clearer. But it would remain functionally identical.

src/backend/executor/spi.c
RuchaSK1: No, this cannot happen. These are set before AllocSetContextCreate which is the only point where OOM can occur. So they are always saved correctly.

Order in SPI_connect_ext:

```
    _SPI_current->outer_processed = SPI_processed;
	_SPI_current->outer_tuptable = SPI_tuptable;
	_SPI_current->outer_result = SPI_result;

	_SPI_current->procCxt = AllocSetContextCreate(_SPI_current->atomic ? TopTransactionContext : PortalContext,
												  "SPI Proc",
												  ALLOCSET_DEFAULT_SIZES);
```

src/backend/executor/spi.c
RuchaSK1: Every Babelfish query passes through PL/tsql which sets internal_xact = true during SPI execution. This includes internal queries during login/session setup as well. So this state is reachable and is the normal path.The existing SPI_finish() never checks this flag either.

I verified by adding a FATAL check it triggers during login itself, before any user query, and breaks the entire functionality:

```
dev-dsk-ruchask-1c-a99d43f3 % run_sqlcmd
Sqlcmd: Error: Microsoft ODBC Driver 17 for SQL Server : TCP Provider: Error code 0x2746.
Sqlcmd: Error: Microsoft ODBC Driver 17 for SQL Server : Communication link failure.
```

Logs:
`2026-03-06 06:19:44.138 UTC [37404] CONTEXT:  TDS Protocol: Message Type: TDS Pre-Login, Phase: Login. 
2026-03-06 06:19:44.138 UTC [37404] LOG:  message_type: TDS7 Login
2026-03-06 06:19:44.138 UTC [37404] CONTEXT:  TDS Protocol: Message Type: TDS Login7, Phase: Login. 
2026-03-06 06:19:44.174 UTC [37404] FATAL:  SPI_finish_safe called during internal transaction
2026-03-06 06:19:44.174 UTC [37404] CONTEXT:  TDS Protocol: Message Type: SQL BATCH, Phase: TDS_REQUEST_PHASE_PROCESS. Writing Done Token`

src/backend/executor/spi.c
```diff
@@ -3471,3 +3471,49 @@ SPI_get_depth(void)
...
+	/* Check if there is an active SPI connection */
+	if (_SPI_current == NULL)
+		return SPI_ERROR_UNCONNECTED;
+
+	/* Fully initialized */
+	if (_SPI_current->procCxt != NULL &&
+		_SPI_current->execCxt != NULL)
+	{
+		return SPI_finish();
+	}
```
tanscorpio7, line 3479: SPI_finish switches to savedcxt as well. `MemoryContextSwitchTo(_SPI_current->savedcxt);`

src/backend/executor/spi.c
```diff
@@ -3471,3 +3471,49 @@ SPI_get_depth(void)
 {
 	return _SPI_connected;
 }
+
```
tanscorpio7, line 3461: Unrelated to this change but I see some redundant babelfsih code in spi.c
```
	if (!isCommit)
	{
		while (_SPI_current && !_SPI_current->internal_xact)
		{
			_SPI_connected--;
			if (_SPI_connected < 0)
				_SPI_current = NULL;
			else
				_SPI_current = &(_SPI_stack[_SPI_connected]);
		}
	}
```

Community has fixed SPI handling in error case and this code is no longer needed. Lets clean it up in this PR.

tanscorpio7 (changes requested):

src/backend/executor/spi.c
tanscorpio7: Sorry, I wasn't clear earlier. I meant something like this.

```
if (_SPI_current->procCxt != NULL &&
	_SPI_current->execCxt != NULL)
{
	return SPI_finish();
}

if (_SPI_current->internal_xact)
	FATAL error

/* Clean up any partially allocated context */
if (_SPI_current->execCxt)
{
	MemoryContextDelete(_SPI_current->execCxt);
	_SPI_current->execCxt = NULL;
}
```

src/backend/executor/spi.c
RuchaSK1: savedcxt is not initialized to NULL before the allocations that can OOM in SPI_connect_ext. So in the partially-initialized path, it could be garbage or a stale pointer and a NULL check won't help.

Since MemoryContextSwitchTo was never called during the failed SPI_connect_ext, CurrentMemoryContext is unchanged. I think there is nothing to restore.

Should we still restore it, or leave it as is?

tanscorpio7 (approved):

tanscorpio7:
Please add [BABEL] prefix to PR title and add link to engine & extension PR to description.

surendravishnoi (approved):

PR #769, opened 2026-06-02
Fix: ENR temp table creation fails for long table names

anju15bharti:
### Description

Temp tables with names longer than 63 characters were erroring out because the name was clipped before storing in ENR. Pass the full name to ENR during creation, and use the full ENR name in ALTER TABLE when rebuilding the RangeVar.
 
### Issues Resolved
BABEL-6433

Extension PR : https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4846
 
### Check List

### DIFF
--- a/src/backend/commands/tablecmds.c
+++ b/src/backend/commands/tablecmds.c
@@ -797,25 +797,12 @@ DefineRelation(CreateStmt *stmt, char relkind, Oid ownerId,
 	LOCKMODE	parentLockmode;
 	Oid			accessMethodId = InvalidOid;
 	ObjectAddress tsql_tabletype_address;
-	const char *enr_relname;
 
 	/*
 	 * Truncate relname to appropriate length (probably a waste of time, as
 	 * parser should have done this already).
 	 */
 	strlcpy(relname, stmt->relation->relname, NAMEDATALEN);
-
-	/*
-	 * For T-SQL #temp tables, ENR stores names in heap-allocated memory
-	 * without NAMEDATALEN limit.  Use the full original name so that
-	 * subsequent lookups via the scanner match the registered ENR entry.
-	 */
-	if (sql_dialect == SQL_DIALECT_TSQL &&
-		stmt->relation->relpersistence == RELPERSISTENCE_TEMP &&
-		stmt->relation->relname[0] == '#')
-		enr_relname = stmt->relation->relname;
-	else
-		enr_relname = relname;
 
 	/*
 	 * Check consistency of arguments
@@ -1091,7 +1078,7 @@ DefineRelation(CreateStmt *stmt, char relkind, Oid ownerId,
 	 * passed in for immediate handling --- since they don't need parsing,
 	 * they can be stored immediately.
 	 */
-	relationId = heap_create_with_catalog(enr_relname,
+	relationId = heap_create_with_catalog(relname,
 										  namespaceId,
 										  tablespaceId,
 										  InvalidOid,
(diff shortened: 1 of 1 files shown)


src/backend/commands/tablecmds.c
```diff
@@ -797,13 +797,26 @@ DefineRelation(CreateStmt *stmt, char relkind, Oid ownerId,
...
 	strlcpy(relname, stmt->relation->relname, NAMEDATALEN);
 
+	/*
+	 * For T-SQL #temp tables, ENR stores names in heap-allocated memory
+	 * without NAMEDATALEN limit.  Use the full original name so that
+	 * subsequent lookups via the scanner match the registered ENR entry.
+	 */
+	if (sql_dialect == SQL_DIALECT_TSQL &&
+		stmt->relation->relpersistence == RELPERSISTENCE_TEMP &&
+		stmt->relation->relname[0] == '#')
```
ayushdsh: a simple `relpersistence` check should be fine enough right? why a `relname` check?

ayushdsh: also lets simply keep effective_relname and eliminate relname for better clarity

src/backend/commands/tablecmds.c
```diff
@@ -5821,6 +5834,23 @@ ATParseTransformCmd(List **wqueue, AlteredTableInfo *tab, Relation rel,
...
+	 */
+	if (sql_dialect == SQL_DIALECT_TSQL &&
+		RelationGetRelationName(rel)[0] == '#')
+	{
+		EphemeralNamedRelation enr = get_ENR_withoid(currentQueryEnv,
+													RelationGetRelid(rel),
+													ENR_TSQL_TEMP, true);
+		if (enr)
+			atstmt->relation->relname = pstrdup(enr->md.name);
+	}
```
ayushdsh: can be simplified to - 

```
EphemeralNamedRelation enr = sql_dialect == SQL_DIALECT_TSQL && get_ENR_withoid(currentQueryEnv,
													RelationGetRelid(rel),
													ENR_TSQL_TEMP, true);

if (enr)
    atstmt->relation->relname = pstrdup(enr->md.name);
```

anju15bharti: I think while with this simplification enr will be NULL when dialect is not TSQL (due to short-circuit), Babelfish coding convention requires all engine-side Babelfish logic to be explicitly fenced under a `sql_dialect == SQL_DIALECT_TSQL` check for readability and to minimize merge conflicts with upstream. Additionally, this drops the relname[0] == '#' guard needed to avoid ENR lookups for table variables and table types.

src/backend/commands/tablecmds.c
anju15bharti: 1. The relname[0] == '#' check is needed to differentiate #temp tables from other RELPERSISTENCE_TEMP objects like table variables(@-prefixed) and table types, which are also temp but don't use ENR.
  2. We can't eliminate relname — it's a char[NAMEDATALEN] stack buffer that strlcpy truncates into, and the truncated name is what gets stored in pg_catalog per core PG code. `effective_relname` is a const char * that points to the full untruncated name for ENR storage. They serve different purposes and can't be unified into a single variable.

ayushdsh: 1. TV do use ENRs, but you're right - I forgot about them.
2. at the end relname and effective_relname both are char* right? I see relname used in 2 places in this function - in the call to heap_create_with_catalog and during partition bound check in `check_new_partition_bound`. why can't we simply use effective_relname in `check_new_partition_bound` as well?

anju15bharti: 2. relname is a char[NAMEDATALEN] array (not a char *), needed as the destination buffer for strlcpy — so it can't be removed. We could use effective_relname in check_new_partition_bound as well, but it adds no value since #temp tables are never partitioned (so both always have the same value there), and it would unnecessarily increase our engine diff against upstream PG which we try to minimize. However I will rename the variable which shows it's only related to temp table.

PR #779, opened 2026-07-30
BABEL-5975: Engine support for long view column names and alias resolution

anju15bharti:
Part 1 of 5 for BABEL-5975 (Long Identifier Support). This PR adds engine-level support for T-SQL views and SELECT statements with column names/aliases that exceed PostgreSQL's NAMEDATALEN (63-byte) limit.
  
Long column names are MD5-truncated in pg_attribute.attname (so they fit the catalog), but the full original name must survive to the client and to dump/restore. To carry it, this PR adds a new resorigname field to TargetEntry that holds the original (untruncated) column name, populated from the query source at parse time and propagated through query transformation, set operations, and planning. Babelfish's TDS layer and metadata views read this field (and the persisted bbf_original_name attoption) to display the original name.
  
  ### Changes
  
  New TargetEntry.resorigname field (primnodes.h)
  
  /* original (untruncated) column name for T-SQL long identifiers (could be NULL) */
  char *resorigname pg_node_attr(equal_ignore, query_jumble_ignore, read_write_ignore, read_as(NULL));
  
  Marked equal_ignore / query_jumble_ignore / read_write_ignore / read_as(NULL) so it is purely informational —
  it does not affect node equality, query jumbling, or serialization, and defaults to NULL for non-Babelfish
  paths.
  
  Populate it for SELECT target entries (parse_target.c, parse_target.h)
  
  - Adds a post_transform_target_entry_hook invoked right after each SELECT target entry is transformed. The
  Babelfish extension installs this hook to extract the original identifier from the query source text and set
  te->resorigname when the name was truncated. Engine-side default is a no-op (NULL hook).
  
  Propagate resorigname through rewrites (tlist.c, analyze.c, prepunion.c)
  
  - tlist.c: copy resorigname when building a matching target entry.
  - analyze.c (bbf_copy_target_entry_resorigname) and prepunion.c (bbf_copy_setop_tle_resorigname): small helpers
  that carry resorigname across set-operation and recursive-CTE target lists, so UNION/INTERSECT/EXCEPT and CTE
  outputs retain the original name. Extracted into helpers to keep the propagation points minimal and reduce
  future merge-conflict surface.
  
  Capture the view column-list location (gram.y)
  
  - For T-SQL CREATE VIEW v (collist) AS ..., the grammar appends an internal bbf_view_collist_loc option
  carrying the source-text offset of the column list (@5), so the extension can recover the original alias names
  for an explicit view column list.
  
  Allow Babelfish attoptions/reloptions in the relevant paths (tablecmds.c)
  
  - Adds allow_attoptions_for_babelfish() and gates the option-validation paths on it, so the bbf_original_name
  attoption / bbf_original_rel_name reloption used by the long-identifier feature are accepted where needed (e.g.
  SELECT INTO / CREATE TABLE AS and column option paths).
  
  Dump/restore of view column attoptions (pg_dump: dump_babel_utils.c/.h, pg_dump.c)
  
  - Adds babelDumpViewColumnAttoptions() and wires it into dumpTableSchema. PostgreSQL's existing per-column
  attoptions dump only runs for tables (the table branch), so view columns need explicit handling. This emits
  ALTER TABLE ONLY <view> ALTER COLUMN <col> SET (...) for the bbf_original_name attoption, ensuring dump/restore preserves the full original view-column name.

extension: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4919
 
### Issues Resolved

BABEL-5975, BABEL-2650, BABEL-4557, BABEL-5052, BABEL-5053,
BABEL-5207, BABEL-5321, BABEL-5416, BABEL-5650
 
### Check List

### DIFF
--- a/src/backend/commands/tablecmds.c
+++ b/src/backend/commands/tablecmds.c
@@ -5139,12 +5139,23 @@ ATPrepCmd(List **wqueue, Relation rel, AlterTableCmd *cmd,
 			break;
 		case AT_SetOptions:		/* ALTER COLUMN SET ( options ) */
 		case AT_ResetOptions:	/* ALTER COLUMN RESET ( options ) */
-			ATSimplePermissions(cmd->subtype, rel,
-								ATT_TABLE | ATT_PARTITIONED_TABLE |
-								ATT_MATVIEW | ATT_FOREIGN_TABLE);
+		{
+			const char *dump_restore = GetConfigOption("babelfishpg_tsql.dump_restore", true, false);
+
+			/* Allow attoptions on view columns in babelfish */
+			if (sql_dialect == SQL_DIALECT_TSQL ||
+				(dump_restore && strcmp(dump_restore, "on") == 0))
+				ATSimplePermissions(cmd->subtype, rel,
+									ATT_TABLE | ATT_PARTITIONED_TABLE |
+									ATT_VIEW | ATT_MATVIEW | ATT_FOREIGN_TABLE);
+			else
+				ATSimplePermissions(cmd->subtype, rel,
+									ATT_TABLE | ATT_PARTITIONED_TABLE |
+									ATT_MATVIEW | ATT_FOREIGN_TABLE);
 			/* This command never recurses */
 			pass = AT_PASS_MISC;
 			break;
+		}
 		case AT_SetStorage:		/* ALTER COLUMN SET STORAGE */
 			ATSimplePermissions(cmd->subtype, rel,
 								ATT_TABLE | ATT_PARTITIONED_TABLE |
--- a/src/backend/commands/tablecmds.c
+++ b/src/backend/commands/tablecmds.c
@@ -58,6 +58,7 @@
 #include "commands/cluster.h"
 #include "commands/comment.h"
 #include "commands/defrem.h"
+#include "commands/extension.h"
 #include "commands/event_trigger.h"
 #include "commands/sequence.h"
 #include "commands/tablecmds.h"
@@ -71,6 +72,7 @@
 #include "foreign/fdwapi.h"
 #include "foreign/foreign.h"
 #include "miscadmin.h"
+#include "libpq/libpq-be.h"
 #include "nodes/makefuncs.h"
 #include "nodes/nodeFuncs.h"
 #include "nodes/parsenodes.h"
@@ -4984,6 +4986,35 @@ ATController(AlterTableStmt *parsetree,
 }
 
 /*
+ * allow_attoptions_for_babelfish
+ *
+ * Check if Babelfish-specific attoptions/reloptions should be allowed.
+ * Returns true only when:
+ *   1. The babelfishpg_tds extension is installed (confirming this is a
+ *      Babelfish database), AND
+ *   2. Either the connection is a TDS connection in T-SQL dialect, OR
+ *      we are in a babelfish dump/restore operation.
+ *
+ * The extension check prevents non-Babelfish users from accidentally
+ * enabling this path by setting babelfishpg_tsql.dump_restore.
+ */
+static bool
+allow_attoptions_for_babelfish(void)
+{
+	const char *dump_restore;
+
+	/* Quick check: babelfishpg_tds must be installed */
+	if (get_extension_oid("babelfishpg_tds", true) == InvalidOid)
+		return false;
+
+	if (sql_dialect == SQL_DIALECT_TSQL &&
+		MyProcPort && MyProcPort->is_tds_conn)
+		return true;
+	dump_restore = GetConfigOption("babelfishpg_tsql.dump_restore", true, false);
+	return (dump_restore && strcmp(dump_restore, "on") == 0);
+}
+
+/*
  * ATPrepCmd
  *
  * Traffic cop for ALTER TABLE Phase 1 operations, including simple
@@ -5139,12 +5170,7 @@ ATPrepCmd(List **wqueue, Relation rel, AlterTableCmd *cmd,
 			break;
 		case AT_SetOptions:		/* ALTER COLUMN SET ( options ) */
 		case AT_ResetOptions:	/* ALTER COLUMN RESET ( options ) */
-		{
-			const char *dump_restore = GetConfigOption("babelfishpg_tsql.dump_restore", true, false);
-
-			/* Allow attoptions on view columns in babelfish */
-			if (sql_dialect == SQL_DIALECT_TSQL ||
-				(dump_restore && strcmp(dump_restore, "on") == 0))
+			if (allow_attoptions_for_babelfish())
 				ATSimplePermissions(cmd->subtype, rel,
 									ATT_TABLE | ATT_PARTITIONED_TABLE |
 									ATT_VIEW | ATT_MATVIEW | ATT_FOREIGN_TABLE);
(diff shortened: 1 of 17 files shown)


src/backend/commands/view.c
```diff
@@ -70,6 +72,19 @@ DefineVirtualRelation(RangeVar *relation, List *tlist, bool replace,
...
+			 * long column names here using the MD5 strategy (handled by
+			 * truncate_identifier) so they fit in the catalog. The original
+			 * full-length name is preserved in pg_attribute.attoptions as
+			 * 'bbf_original_name' and resolved at query time via
+			 * bbf_get_view_column_name().
+			 */
+			if (def->colname && strlen(def->colname) >= NAMEDATALEN)
+				truncate_identifier(def->colname,
+								strlen(def->colname), false);
+
```
tanscorpio7: Why is the parser not truncating it. In PG truncation is parser's responsibility.

Currently there is very rare places where identifier greater than namedatalen flows in PG code in the identifier fields after parsing. We shouldn't break that.

tanscorpio7 (changes requested):

src/backend/commands/view.c
anju15bharti: The parser truncates all identifiers. tle->resname is the only field carrying the full-length name — set post-parse by the extension hook. Without this, the original view column name would be permanently lost: pg_attribute.attname is clipped to NAMEDATALEN, attoptions is not supported for view columns, and the query string is gone after CREATE VIEW completes. The rewrite rule's tle->resname is the only persistent storage for the original name (please ignore *full-length name is preserved in pg_attribute.attoptions* in the comment as it is true for tables/select into and not for views, will remove it).

It is safe because:
  
  1. It's a char * (no NAMEDATALEN constraint), stored in pg_rewrite as serialized nodes
  2. Every path into Name-type catalog fields applies truncate_identifier first:
    - Views: this engine PR (in DefineVirtualRelation)
    - SELECT INTO: extension truncates before ExecCreateTableAs
  
  3. All other usages (TDS output, ORDER BY, ruleutils) only read it — no fixed-buffer copy
  
  No > NAMEDATALEN value reaches catalog Name fields without hash-based truncation.

src/backend/commands/tablecmds.c
```diff
@@ -5139,12 +5139,23 @@ ATPrepCmd(List **wqueue, Relation rel, AlterTableCmd *cmd,
...
 		case AT_ResetOptions:	/* ALTER COLUMN RESET ( options ) */
-			ATSimplePermissions(cmd->subtype, rel,
-								ATT_TABLE | ATT_PARTITIONED_TABLE |
-								ATT_MATVIEW | ATT_FOREIGN_TABLE);
+		{
+			const char *dump_restore = GetConfigOption("babelfishpg_tsql.dump_restore", true, false);
+
+			/* Allow attoptions on view columns in babelfish */
+			if (sql_dialect == SQL_DIALECT_TSQL ||
+				(dump_restore && strcmp(dump_restore, "on") == 0))
```
ayushdsh: just a question for my own clarification: when do we use `babelfish_dump_restore` flag vs. GetConfigOption() - @tanscorpio7 ?

ayushdsh (commented):
otherwise lgtm

src/backend/commands/tablecmds.c
tanscorpio7: Always preferrable to use the GUC variable directly instead of GetConfigOption. In some case direct GUC variable is not accessible for example cross extension usage.

src/bin/pg_dump/pg_dump.c
```diff
@@ -17091,6 +17091,24 @@ dumpTableSchema(Archive *fout, const TableInfo *tbinfo)
...
+				if (tbinfo->attoptions[j][0] != '\0')
+				{
+					fixAttoptionsBbfOriginalName(fout, tbinfo->dobj.catId.oid, tbinfo, j);
+					appendPQExpBuffer(q, "ALTER TABLE ONLY %s ALTER COLUMN %s SET (%s);\n",
+									 qualrelname,
+									 fmtId(tbinfo->attnames[j]),
+									 tbinfo->attoptions[j]);
+				}
+			}
+		}
```
tanscorpio7: 1. Lets use the existing pattern for making to changed to dump. A new function babel_* in dump_babel_utils.c and a quick exit with isBabelfishDatabase
2. Wouldn't the existing logic for dumping column level attribute automatically pick this up ?

src/backend/commands/tablecmds.c
tanscorpio7: @anju15bharti Please consider converting this to a static function and use both dialect + is_tds_conn check for safety.
Also no babelfish specific code should executed for non babelfish database, in this case GetConfigOption. Harmless but lets follow the standard practice.

tanscorpio7 (changes requested):

src/backend/commands/tablecmds.c
```diff
@@ -5139,9 +5156,19 @@ ATPrepCmd(List **wqueue, Relation rel, AlterTableCmd *cmd,
...
 		case AT_ResetOptions:	/* ALTER COLUMN RESET ( options ) */
-			ATSimplePermissions(cmd->subtype, rel,
-								ATT_TABLE | ATT_PARTITIONED_TABLE |
-								ATT_MATVIEW | ATT_FOREIGN_TABLE);
+			/*
+			 * In Babelfish, allow attoptions on view columns (needed for
+			 * storing original long column names). During dump/restore the
+			 * dialect is PG but we still need to allow it.
+			 */
+			if (is_babelfish_view_attoption_allowed())
```
tanscorpio7: Duplicate comment, commentary in is_babelfish_view_attoption_allowed() should be sufficent.

src/backend/commands/tablecmds.c
```diff
@@ -4992,7 +4993,23 @@ ATController(AlterTableStmt *parsetree,
...
+{
+	const char *dump_restore;
+
+	if (sql_dialect == SQL_DIALECT_TSQL &&
+		MyProcPort && MyProcPort->is_tds_conn)
+		return true;
+	dump_restore = GetConfigOption("babelfishpg_tsql.dump_restore", true, false);
+	return (dump_restore && strcmp(dump_restore, "on") == 0);
+}
+
```
tanscorpio7, line 5016: 1. Lets use a more generic name for `is_babelfish_view_attoption_allowed` ? allow_attoptions_for_babelfish() ? since it is also used for partitioned index below.
2. Lets have a `babelfish_tds` extension installed check as well to avoid cases where a non babelfish customers sets `babelfishpg_tsql.dump_restore` and changes regular PG behaviour.

src/backend/parser/parse_target.c
```diff
@@ -188,13 +189,19 @@ transformTargetList(ParseState *pstate, List *targetlist,
...
 		 * variable, so transform as a single expression
 		 */
-		p_target = lappend(p_target,
-						   transformTargetEntry(pstate,
-												res->val,
-												NULL,
-												exprKind,
-												res->name,
-												false));
+		{
```
tanscorpio7: Braces without if,for, etc is not in PostgreSQL indentation style.

tanscorpio7 (changes requested):

src/bin/pg_dump/pg_dump.c
anju15bharti: 1. done
2. The existing column attoption dump logic only fires for tables, not views. Views are dumped via CREATE VIEW which has no syntax for column-level attoptions. So we need explicit ALTER VIEW ... ALTER COLUMN ... SET (bbf_original_name=...) statements after the CREATE VIEW to restore them during dump/restore.

tanscorpio7 (approved):

PR #783, opened 2026-08-13
[BABEL] Fix statement-level trigger firing N times during FK CASCADE

rohit01010:
### Description

Statement-level AFTER DELETE/UPDATE triggers with transition tables fire once per cascaded row instead of once per statement when the trigger language is pltsql. Each firing sees the full set of cascaded rows in the shared transition table, producing N² writes instead of N.

The root cause is that `cancel_prior_stmt_triggers()` only scans the query_stack event list to deduplicate statement-level trigger events. Composite (pltsql) trigger events are routed to a separate list (`compositeTriggers.triggers->data.events`) and were never cancelled, allowing duplicate events to accumulate during FK CASCADE. Fixed by extending `cancel_prior_stmt_triggers()` to also scan and cancel prior events in the composite trigger event list.

Extension PR: https://github.com/babelfish-for-postgresql/babelfish_extensions/pull/4933

Authored-by: Rohit Bhagat <rohitbgt@amazon.com>
Signed-off-by: Rohit Bhagat <rohitbgt@amazon.com>

### Issues Resolved

BABEL-7022
 
### Check List

### DIFF
--- a/src/backend/commands/trigger.c
+++ b/src/backend/commands/trigger.c
@@ -125,6 +125,7 @@ static void set_iot_state(Oid relid, CmdType cmdType, IOTState newState);
 static bool IsCompositeTriggerActive(void);
 static void AddCompositeTriggerLevelData(void);
 static bool IsCompositeTrigger(Oid fn_oid, List *fn_name);
+static void cancel_prior_composite_stmt_triggers(Oid relid, int tgevent);
 
 static bool FreeTriggerTable(int query_depth);
 
@@ -7413,39 +7414,11 @@ cancel_prior_stmt_triggers(Oid relid, CmdType cmdType, int tgevent)
 done:
 
 	/*
-	 * BABELFISH: Also cancel matching statement-level events in the
-	 * composite trigger event list. Composite (pltsql) trigger events
-	 * are stored in a separate list from the standard query_stack, so
-	 * the scan above does not reach them. Without this, duplicate
-	 * statement-level events can accumulate when multiple row-level
-	 * actions (e.g. FK CASCADE) each queue a statement-level event
-	 * for the same relation and operation.
+	 * BABELFISH: Also cancel matching statement-level events in the composite
+	 * trigger event list, which the query_stack scan above does not reach.
 	 */
-	if (table->after_trig_done && IsCompositeTriggerActive())
-	{
-		AfterTriggerEvent event;
-		AfterTriggerEventChunk *chunk;
-
-		for_each_event_chunk(event, chunk, compositeTriggers.triggers->data.events)
-		{
-			AfterTriggerShared evtshared = GetTriggerSharedData(event);
-
-			if (evtshared->ats_relid != relid)
-				continue;
-			if ((evtshared->ats_event & TRIGGER_EVENT_OPMASK) != tgevent)
-				continue;
-			if (!TRIGGER_FIRED_FOR_STATEMENT(evtshared->ats_event))
-				continue;
-			if (!TRIGGER_FIRED_AFTER(evtshared->ats_event))
-				continue;
-			if (event->ate_flags & AFTER_TRIGGER_DONE)
-				continue;
-
-			/* Mark it DONE so it won't fire again */
-			event->ate_flags &= ~AFTER_TRIGGER_IN_PROGRESS;
-			event->ate_flags |= AFTER_TRIGGER_DONE;
-		}
-	}
+	if (table->after_trig_done)
+		cancel_prior_composite_stmt_triggers(relid, tgevent);
 
 	/* In any case, save current insertion point for next time */
 	table->after_trig_done = true;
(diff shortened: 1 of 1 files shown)


src/backend/commands/trigger.c
```diff
@@ -7412,6 +7412,41 @@ cancel_prior_stmt_triggers(Oid relid, CmdType cmdType, int tgevent)
...
+	/*
+	 * BABELFISH: Also cancel matching statement-level events in the
+	 * composite trigger event list. Composite (pltsql) trigger events
+	 * are stored in a separate list from the standard query_stack, so
+	 * the scan above does not reach them. Without this, duplicate
+	 * statement-level events can accumulate when multiple row-level
+	 * actions (e.g. FK CASCADE) each queue a statement-level event
+	 * for the same relation and operation.
+	 */
+	if (table->after_trig_done && IsCompositeTriggerActive())
```
KushaalShroff: Why is this a direct change in engine with no TSQL dialect checks at all?

KushaalShroff (changes requested):

src/backend/commands/trigger.c
KushaalShroff, line 7565: Is this only valid for TSQL? Why is it in engine and not a hook? Where are we safeguarding with the dialect checks?

KushaalShroff (dismissed):

src/backend/commands/trigger.c
ayushdsh: +1 but I think we need to think if it should be a dialect or a tdsconn check?

rohit01010: > Why is this a direct change in engine with no hook?

The composite trigger data structures (compositeTriggers, CompositeTriggerData, AfterTriggerEvent, ate_flags) and functions (IsCompositeTriggerActive(), for_each_event_chunk) are all defined statically in trigger.c on the engine side. Moving this logic to the extension via a hook would require exposing these engine-internal structures to the extension. We can consider that as a separate refactoring task if needed as it is not very trivial. And current implementation follows the same pattern as existing code.

> No TSQL dialect checks at all?

IsCompositeTriggerActive() is sufficient here, it returns true only when compositeTriggers.triggers != NULL and the trigger level matches, which can only happen when BeginCompositeTriggers() was called from the pltsql executor during pltsql trigger execution. The underlying struct is allocated in AddCompositeTriggerLevelData(), which is itself guarded by check_pltsql_support_tsql_transactions_hook (blocks non-TDS connections). So there is no code path where IsCompositeTriggerActive() returns true outside of a pltsql trigger context.

This follows the same pattern used across other composite trigger functions in trigger.c (e.g., FreeTriggerTable, AfterTriggerSaveEvent, GetAfterTriggersTableData) all use compositeTriggers.triggerLevel > 0 or IsCompositeTriggerActive() as the guard without an additional dialect check.

That said, we can still add sql_dialect == SQL_DIALECT_TSQL as a defensive outer check.

src/backend/commands/trigger.c
rohit01010: The composite trigger mechanism was added as direct engine code in the initial Babelfish commits and hasn't been refactored since. Our fix extends the existing engine-side logic following the same pattern. We can file a separate JIRA to explore migrating the composite trigger logic to the extension via hooks.

> Where are we safeguarding with the dialect checks?

We have this safeguard in `AddCompositeTriggerLevelData()`:
```
	/* Check if this trigger is being added as part of any of postgres' function, procedure or trigger in the stack */
	if (check_pltsql_support_tsql_transactions_hook && !(*check_pltsql_support_tsql_transactions_hook)())
	{
		ereport(ERROR,
					(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
					 errmsg("T-SQL trigger can not be executed from PostgreSQL function, procedure or trigger.")));
	}
```
This blocks non-TDS connections from allocating the `CompositeTriggerData` struct. Since `IsCompositeTriggerActive()` requires `compositeTriggers.triggers != NULL` (which is only set when this struct is allocated), it can never return true outside of a pltsql trigger execution context.

src/backend/commands/trigger.c
```diff
@@ -7520,14 +7555,21 @@ void EndCompositeTriggers(bool error)
...
-			/* Mark all triggers as IN PROGRESS */
+			/* Mark all triggers as IN PROGRESS, skipping already-cancelled ones */
 			for_each_event_chunk(event, chunk, triggers->data.events)
 			{
 				AfterTriggerShared evtshared = GetTriggerSharedData(event);
 
 				evtshared->ats_firing_id = 0;
-				Assert(!(event->ate_flags &
-						 (AFTER_TRIGGER_DONE | AFTER_TRIGGER_IN_PROGRESS)));
+				/*
```
surendravishnoi, line 7585: can we double check this against PG equivalent function for end trigger?
Also, make sure that logic works with
explicit transaction and commit
explicit transaction and rollback
explicit transaction and release savepoint
explicit transaction and rollback to savepoint

src/backend/commands/trigger.c
```diff
@@ -7412,6 +7412,41 @@ cancel_prior_stmt_triggers(Oid relid, CmdType cmdType, int tgevent)
 	}
 done:
 
+	/*
```
surendravishnoi, line 7416: create a static function and move with other functions related to composite trigger

rohit01010: Done

src/backend/commands/trigger.c
rohit01010: > can we double check this against PG equivalent function for end trigger?

Checked against PG's equivalent (afterTriggerMarkEvents, used by AfterTriggerEndQuery). PG skips events that are already DONE or IN_PROGRESS before marking them to fire:
```
    if (!(event->ate_flags & (AFTER_TRIGGER_DONE | AFTER_TRIGGER_IN_PROGRESS)))
        ... mark IN_PROGRESS ...
```
Our change does the equivalent DONE skip. 

> Also, make sure that logic works with
explicit transaction and commit
explicit transaction and rollback
explicit transaction and release savepoint
explicit transaction and rollback to savepoint

Verified and added tests for the above scenario's.

KushaalShroff (commented):
/code-review

surendravishnoi (approved):

src/backend/commands/trigger.c
```diff
@@ -7450,6 +7458,54 @@ bool IsCompositeTriggerActive(void)
...
+ * that AFTER STATEMENT triggers fire only once even when several FK enforcement
+ * triggers sequentially queue triggers for the same table into the same trigger
+ * query level.
+ *
+ * Unlike the query_stack scan, we cannot stop at the first non-matching event:
+ * the composite list may interleave events for different relations and
+ * operations, so we scan the whole list and skip non-matching entries.
+ */
+static void
+cancel_prior_composite_stmt_triggers(Oid relid, int tgevent)
```
tanscorpio7, line 7479: 1. Please check if we also need to handle something similar to https://github.com/rohit01010/postgresql_modified_for_babelfish/commit/fd31f9f033213e2ebf00b57ef837e1828c338fc4

2. PG has a different logic for looping over the events. Could you please explain the difference ?

ayushdsh (approved):

tanscorpio7 (changes requested):

src/backend/commands/trigger.c
rohit01010: > 1. Please check if we also need to handle something similar to https://github.com/rohit01010/postgresql_modified_for_babelfish/commit/fd31f9f033213e2ebf00b57ef837e1828c338fc4

No additional fix is needed here. BEFORE STATEMENT triggers use a structurally different mechanism that isn't susceptible to this bug.

AFTER STATEMENT triggers are *deferred*: they get queued as events via AfterTriggerSaveEvent() and fire later, which is exactly why cancel_prior_stmt_triggers() needs to scan/cancel previously-queued events when a new set gets queued (multiple FK actions can each add a queued event for the same table+op).

BEFORE STATEMENT triggers are *not deferred*: they fire immediately, inline, via ExecBSInsertTriggers()/ExecBSDeleteTriggers() calling ExecCallTriggerFunc() directly. They never go through AfterTriggerSaveEvent() or get added to any event list (composite or query_stack). Their "only fire once" guarantee comes from a simple sticky flag, before_stmt_triggers_fired(), checked against AfterTriggersTableData.before_trig_done, not a list scan. That flag is looked up via GetAfterTriggersTableData(relid, cmdType), which is keyed only on relid+cmdType and is shared correctly regardless of composite/pltsql state, so it already dedupes correctly for T-SQL triggers with no special-casing needed.

I also confirmed SQL Server does not support native BEFORE triggers. So a composite/pltsql trigger can never be a BEFORE STATEMENT trigger through the T-SQL surface this class of trigger doesn't exist in the T-SQL dialect at all.

> 2. PG has a different logic for looping over the events. Could you please explain the difference ?

The difference is intentional, and is already called out in the comment above cancel_prior_composite_stmt_triggers: PG's cancel_prior_stmt_triggers() scans query_stack's event list, which only ever contains AFTER-trigger events queued in insertion order for that query level, so it can exit at the first non-matching event via goto done, because once it stops matching (different relid/op/timing), everything after is guaranteed not to match either (the list is contiguous per relid+cmdType at that point in the scan).

The composite list (compositeTriggers.triggers->data.events) doesn't have that guarantee, it's a single shared list across *all* relations and operations touched by the pltsql trigger execution, so events for different tables/ops can be interleaved in it. We can't assume "first non-match means no more matches later in the list," so the helper scans the whole list and continues past non-matching entries instead of exiting early on the first mismatch.

tanscorpio7: So If we are deleting a n rows, will this loop run per parent row, i.e. n^2 complexiity ? 

Just for completeness please check if we not regressing in perofmrnace because of this, unlikley because same case would run n^2 triggers without this fix.

Also a jira for optimizing this later on.

tanscorpio7 (approved):

